PR #41, opened 2020-07-17
Revert "Improve capa checkout in sync workflow"

mr-tz:
Seems like something's off here

Reverts fireeye/capa-rules#40


Ana06:
it seems the token is needed. I'll send another one. It is the same problem as in https://github.com/fireeye/capa-testfiles.

PR #43, opened 2020-07-17
Adding QueryWorkingSet from psapi to be used in ReadProcessMemory rule and second NtQuerySystemInformation rule.

re-fox:
QueryWorkingSet from psapi will also work in a similar fashion to VirtualQueryEx.  

PoC code is on github for using QueryWorkingSet or VirtualQueryEx  -> https://github.com/SpiderLabs/malware-analysis/blob/master/C/queryWorkingSet.c 

A blog post with full instructions can be found here -> https://www.trustwave.com/en-us/resources/blogs/spiderlabs-blog/new-memory-scraping-technique-in-cherry-picker-pos-malware/

Both calls will ultimately land at NtQueryVirtualMemory.

I'm not sure if there is a process to modify authors of these rules as contributors commit up changes, so I'm leaving the author field as is.

### DIFF
--- a/nursery/enumerate-processes-via-ntquerysysteminformation.yml
+++ b/nursery/enumerate-processes-via-ntquerysysteminformation.yml
@@ -1,0 +1,14 @@
+rule:
+  meta:
+    name: query processes via NtQuerySystemInformation
+    namespace: host-interaction/process/list
+    author: "@_re_fox"
+    scope: basic block
+    att&ck:
+      - Discovery::Process Discovery [T1057]
+    examples:
+      - 31bd8dd48ac0de3d4da340bf29f4d280:0x00401be3
+  features:
+    - and:
+      - number: 0x5 = SYSTEM_PROCESS_INFORMATION
+      - api: ntdll.NtQuerySystemInformation
--- a/nursery/read-process-memory.yml
+++ b/nursery/read-process-memory.yml
@@ -11,3 +11,4 @@ rule:
         - or:
           - api: kernel32.OpenProcess
           - api: kernel32.VirtualQueryEx
+          - api: psapi.QueryWorkingSet


re-fox:
I've also added up a nursery rule enumerating process via NtQuerySystemInformation in `5d86cde`

A mapping of SystemInformationClass constants can be found here -> http://www.exploit-monday.com/2013/06/undocumented-ntquerysysteminformation.html, which SYSTEM_PROCESS_INFORMATION maps to 0x5.

This can be observed the sample 31bd8dd48ac0de3d4da340bf29f4d280 where NtQuerySystemInformation is used to obtain a list of processes and that is passed along to functions that will even ReadProcessMemory.

williballenthin:
> I'm not sure if there is a process to modify authors of these rules as contributors commit up changes, so I'm leaving the author field as is.

I'm not sure of the process either 🤔 but I think we really want to recognize everyone that contributes. So, let me leave this PR open so that you can add another commit updating the `author` field to be a comma separated list that includes your handle. We'll get your handle in the release notes for the next version, too!

Alternatively, given that you have documentation for the specific technique using `QueryWorkingSet`, we could pull this one into its own rule. The benefit is that its easier to associated the references with the logic (and also your handle).

Let me know when I should hit "merge" and I will :-)

re-fox:
> Let me know when I should hit "merge" and I will :-)

I think you're good to hit merge.  Especially if it's cleaner/easier to open an issue (for additional authors) and close a PR. 

The `QueryWorkingSet` rule will still need the other `OpenProcess`, `ReadProcessMemory`, etc.. calls to do anything that is useful from the malware perspective.  Its really just a swap out for `VirtualQueryEx`.  Leaving it in the ReadProcessMemory rule is fine with me.  As with most of these API calls, they are all just flavours of something that really happens down at the `NT*` level.

I plan to contrib up more rules, I've been working with this tool over the weekend and found it a very useful addition to my process.  Thank you.

mr-tz:
> I'm not sure of the process either 🤔 but I think we really want to recognize everyone that contributes. So, let me leave this PR open so that you can add another commit updating the author field to be a comma separated list that includes your handle.

You can also add a new entry like so:
```yml
author:
  - matthew.williams@fireeye.com
  - "@_re_fox"
```

williballenthin:
@moritz, I think that will break the renderer because we currently assume
`author` to be a string, not a list. We should support this schema though.
Should be a simple change.

mr-tz:
No, you already accounted for it :) https://github.com/fireeye/capa/blob/master/capa/render/vverbose.py#L188

williballenthin:
🤦

PR #44, opened 2020-07-18
Adding anti-vm rule to check for an unmoving mouse cursor

BitsOfBinary:
Simple anti-vm technique that makes use of the "GetCursorPos" API (called at least twice) to check if the mouse cursor ever moves.

Initially placed this in the nursery for testing.


BitsOfBinary:
In hindsight, it's probably more accurate to put this under a category of "anti-sandboxing" than "anti-vm".

williballenthin:
very cool, thank you!

PR #45, opened 2020-07-22
Create check-mutex-and-exit.yml

re-fox:
The point of sale malware `1d8fd13c890060464019c0f07b928b1a` will check for the existence of a mutex and bail (`ExitProcess`) if the mutex exists.  This rule will describe that feature.  

Attached is a screenshot of the function, with the error codes the malware will check for (`0xb7 = ERROR_ALREADY_EXISTS` and `5 = ERROR_ACCESS_DENIED`)

image

This does create a question though, is this too specific of a rule?  Capa already has a rule for creating a mutex and another rule for checking a mutex, both are similar to my rule.  This is just one extension more to check and see if the program will exit.  

I'm (personally) a proponent of having many specific rules, even if a lot of them overlap with similar features.  Let me know your feedback and I'm happy to make any required changes.

### DIFF
--- a/nursery/check-mutex-and-exit.yml
+++ b/nursery/check-mutex-and-exit.yml
@@ -8,9 +8,7 @@ rule:
       - 1d8fd13c890060464019c0f07b928b1a:0x00402eb0
   features:
     - and:
-        - or:
-          - api: CreateMutex
-          - api: CreateMutexEx
+        - match: create mutex
         - api: GetLastError
         - api: ExitProcess
         - or:
--- a/nursery/check-mutex-and-exit.yml
+++ b/nursery/check-mutex-and-exit.yml
@@ -1,0 +1,19 @@
+rule:
+  meta:
+    name: check mutex and exit
+    namespace: host-interaction/mutex
+    author: "@_re_fox"
+    scope: function
+    examples:
+      - 1d8fd13c890060464019c0f07b928b1a:0x00402eb0
+  features:
+    - and:
+        - or:
+          - api: CreateMutex
+          - api: CreateMutexEx
+        - api: GetLastError
+        - api: ExitProcess
+        - or:
+            - number: 2 = ERROR_FILE_NOT_FOUND
+            - number: 0xb7 = ERROR_ALREADY_EXISTS
+            - number: 5 = ERROR_ACCESS_DENIED


nursery/check-mutex-and-exit.yml
```diff
@@ -0,0 +1,19 @@
+rule:
+  meta:
+    name: check mutex and exit
+    namespace: host-interaction/mutex
+    author: "@_re_fox"
+    scope: function
+    examples:
+      - 1d8fd13c890060464019c0f07b928b1a:0x00402eb0
```
williballenthin, line 8: would you add this file to the [capa-testfiles](https://github.com/fireeye/capa-testfiles/) repository? this way our linter can ensure the rule matches as expected.

nursery/check-mutex-and-exit.yml
```diff
@@ -0,0 +1,19 @@
...
+  features:
+    - and:
+        - or:
```
williballenthin, line 11: this block could be `match: create mutex`. your choice.

nursery/check-mutex-and-exit.yml
re-fox: Created PR https://github.com/fireeye/capa-testfiles/pull/3

mr-tz:
> This does create a question though, is this too specific of a rule? Capa already has a rule for creating a mutex and another rule for checking a mutex, both are similar to my rule. This is just one extension more to check and see if the program will exit.

I think combining rules into new (more specific) capabilities is a key idea with capa. In that regard, I would also recommend to rely on existing rules, i.e. `create mutex` here.

williballenthin:
outstanding! i think this is a perfect example of a capa rule - describes a specific behavior and is self-documenting.

i do agree there is overlap with the `check mutex` rule; however, this rule goes a bit further in its description, requiring the `ExitProcess` call.

williballenthin:
with the latest changes, this looks good to me. thanks @re-fox !

re-fox:
Thank you for the comments and feedback @mr-tz @williballenthin  .  I changed the rule to lean on existing logic `match: create mutex`.

williballenthin:
now that this is merged, i can ask the hard questions...

in your screenshot above, that looks a lot like ghidra. are you using capa with ghidra? how did you get that highlighting of the relevant instructions?

PR #47, opened 2020-07-23
Create credit-card-feature-parsing.yml

re-fox:
I've been working through sample `1d8fd13c890060464019c0f07b928b1a` (which was pushed to the testfiles repo) and making sure that rules are triggering on all significant behaviour.

@williballenthin I noticed that you had a credit card rule, but it wasn't triggering on this sample, due to the `read process memory` rule hitting on a different function.  I added in the additional sentinel checks that were performed by this sample and added it as another nursery rule.

Running against the sample will now show 
```
credit card feature parsing
namespace  collection/credit-card
scope      function
matches    0x402860
```

### DIFF
--- a/nursery/credit-card-feature-parsing.yml
+++ b/nursery/credit-card-feature-parsing.yml
@@ -1,0 +1,38 @@
+rule:
+  meta:
+    name: credit card feature parsing
+    namespace: collection/credit-card
+    author: "@_re_fox"
+    scope: function
+  examples:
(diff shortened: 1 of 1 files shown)


williballenthin:
good rule!

if this passes the linter with no warnings, then we can put it directly in the appropriate directory. the nursery is our way of saying to the linter "i know, i know, i'll get to it".

williballenthin:
i'll merge this and move the file around myself, so that i can address https://github.com/fireeye/capa/issues/182

thanks @re-fox !

PR #48, opened 2020-07-23
Create read-initialization-file.yml

re-fox:
Iterating through the features of `1d8fd13c890060464019c0f07b928b1a` already in the `capa-testfiles` repo.  I added in the feature to determine if the malware is reading from an `.ini` file.   In the mentioned hash, the malware reads in `1.ini`.

Here's a screenshot of the basic block responsible for the action:
image

I have the rule scoped for `function` right now, but if this causes too many FP's or is error prone, scoping it to `basic block` might make sense.

Per your last comments, I posted directly to ` host-interaction/file-system/read/`, if this rule needs to go to the nursery, I'd be happy to move it there.

Thanks for the help and feedback.

### DIFF
--- a/host-interaction/file-system/read/read-initialization-file.yml
+++ b/host-interaction/file-system/read/read-initialization-file.yml
@@ -1,0 +1,18 @@
+rule:
+  meta:
+    name: read initialization file
+    namespace: host-interaction/file-system/read
+    author: "@_re_fox"
+    scope: function
+  examples:
+    - 1d8fd13c890060464019c0f07b928b1a:0x401070
+  features:
+    - and:
+      - or:
+        - api: GetPrivateProfileInt
+        - api: GetPrivateProfileString
+        - api: GetPrivateProfileSection
+        - api: GetPrivateProfileSectionNames
+      - string: /\.ini/i
+      - optional:
+        - api: GetFullPathName
--- a/host-interaction/file-system/read/read-ini-file.yml
+++ b/host-interaction/file-system/read/read-ini-file.yml
@@ -1,6 +1,6 @@
 rule:
   meta:
-    name: read ini file
+    name: read .ini file
     namespace: host-interaction/file-system/read
     author: "@_re_fox"
     scope: function
--- a/host-interaction/file-system/read/read-ini-file.yml
+++ b/host-interaction/file-system/read/read-ini-file.yml
@@ -1,6 +1,6 @@
 rule:
   meta:
-    name: read .ini file
+    name: read ini file
     namespace: host-interaction/file-system/read
     author: "@_re_fox"
     scope: function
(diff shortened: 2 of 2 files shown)


re-fox:
Also forgot to post, when running the rule, it should now show.

```
read initialization file
namespace  host-interaction/file-system/read
author     @_re_fox
scope      function
function @ 0x401070
  and:
    or:
      api: GetPrivateProfileString @ 0x4010E0
    string: 1.ini @ 0x401079, 0x4010B6
    optional:
      api: GetFullPathName @ 0x4010BB
```

host-interaction/file-system/read/read-initialization-file.yml
```diff
@@ -0,0 +1,18 @@
+rule:
+  meta:
+    name: read initialization file
+    namespace: host-interaction/file-system/read
+    author: "@_re_fox"
+    scope: function
+  examples:
```
williballenthin, line 7: examples needs to be under the `rule.meta` block

if you have the source of capa checked out, then you can run `python scripts/linter.py rules` to find quick fixes like this

host-interaction/file-system/read/read-initialization-file.yml
```diff
@@ -0,0 +1,18 @@
+rule:
+  meta:
+    name: read initialization file
```
williballenthin, line 3: can we make this name `read .ini file`? i forgot that ".ini" probably stands for "initialization". not a blocker though, if you'd prefer not to change this.

williballenthin:
> read initialization file
> namespace  host-interaction/file-system/read
> author     @_re_fox
> scope      function
> function @ 0x401070
>   and:
>     or:
>       api: GetPrivateProfileString @ 0x4010E0
>     string: 1.ini @ 0x401079, 0x4010B6    <<<<<<<<<<<<<<<<<<<<<<<
>     optional:
>       api: GetFullPathName @ 0x4010BB

@mr-tz its neat how the file path is actually pulled out here due to how the regex matching works

host-interaction/file-system/read/read-initialization-file.yml
re-fox: Absolutely, I prefer `.ini`

host-interaction/file-system/read/read-initialization-file.yml
re-fox: Ah, silly mistake.  Fixed.

williballenthin:
need to fix the linter to ignore `.` in rule names - I'll work on that. Can we keep the `.ini` in the rule name?

rule looks good. i'll merge it with that name fix.

re-fox:
Reverted the name back, let's hope the linter passes.

williballenthin:
it doesn't, but not your fault :-)

thanks @re-fox !

PR #52, opened 2020-07-24
Update validate-credit-card-number-with-luhn-algorithm.yml

re-fox:
This is a rule tightening PR.  

Added loop modifier now that https://github.com/fireeye/capa/pull/192 is pulled.  

Added case where a sample (`af13e7583ed1b27c4ae219e344a37e2b`) was using a `lea eax,[eax + -0x30]` rather than a `sub`.

I've tested against multiple POS malware samples and this seems to catch a fair bit of them, I've got some slight modifications that will be coming in the next couple days to support the corner cases.

### DIFF
--- a/data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
+++ b/data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
@@ -9,6 +9,8 @@ rule:
       - 60abaef3fda131ffa20df480cb3f8029:0x4048e0
   features:
     - and:
+      - match: contain loop
+        description: Iterate over CC digits
       - basic block:
         - and:
           - number: 0x0
@@ -23,10 +25,15 @@ rule:
           - number: 0x9
         description: Digital root lookup table
       - basic block:
-        - and:
-          - number: 0x30
-          - mnemonic: sub
-          description: Conversion of chr to int 
+        - or:
+          - and:
+            - number: 0x30
+            - mnemonic: sub
+            description: Conversion of chr to int (SUB 0x30) 
+          - and:
+            - mnemonic: lea
+            - offset: -0x30
+            description: Conversion of chr to int (LEA REG,[REG+ -0x30])
       - basic block:
         - and:
           - mnemonic: idiv
--- a/data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
+++ b/data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
@@ -35,10 +35,20 @@ rule:
             - offset: -0x30
             description: Conversion of chr to int (LEA REG,[REG+ -0x30])
       - basic block:
-        - and:
-          - mnemonic: idiv
-          - mnemonic: cdq
-          - number: 0xa
-          - optional : 
-            - mnemonic: neg
-          description: Final section returning checkum % 10
+        - or:
+          - and:
+            - mnemonic: idiv
+            - mnemonic: cdq
+            - number: 0xa
+            - optional : 
+              - mnemonic: neg
+            description: Final section returning checkum % 10
+          - and:
+            - mnemonic: shr
+            - mnemonic: imul
+            - number: 0x66666667
+            - number: 0x1f
+            - number: 0x2
+            - optional : 
+              - mnemonic: neg
+            description: Compiler optimized returning checkum % 10 


data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
```diff
@@ -23,10 +25,15 @@ rule:
...
-          - mnemonic: sub
-          description: Conversion of chr to int 
+        - or:
+          - and:
+            - number: 0x30
+            - mnemonic: sub
+            description: Conversion of chr to int (SUB 0x30) 
+          - and:
+            - mnemonic: lea
+            - offset: -0x30
```
williballenthin, line 35: i don't think negative numbers are supported.

right @mike-hunhoff ? or will this work for offsets, just not numbers?

re-fox:
I noticed this error in the linter:
```
validate credit card number with luhn algorithm 
  FAIL: feature value is negative: capa treats all numbers as unsigned values; you may specify the number's two's complement representation; will not match on "-48"
```
However the rule matches on my test sample
```
validate credit card number with luhn algorithm
namespace  data-manipulation/checksum
author     @_re_fox
scope      function
examples   1d8fd13c890060464019c0f07b928b1a:0x401920, 60abaef3fda131ffa20df480cb3f8029:0x4048e0
function @ 0x402A41
  and:
    match: contain loop = Iterate over CC digits @ 0x402A41
      or:
        characteristic: loop @ 0x402A41
    basic block:
      and:
        number: 0x0 @ 0x402A47, 0x402A94
        number: 0x2 @ 0x402A4E
        number: 0x4 @ 0x402A55
        number: 0x6 @ 0x402A5C
        number: 0x8 @ 0x402A63
        number: 0x1 @ 0x402A6A, 0x402A8D
        number: 0x3 @ 0x402A71
        number: 0x5 @ 0x402A78
        number: 0x7 @ 0x402A7F
        number: 0x9 @ 0x402A86
    basic block:
      or:
        and:
          mnemonic: lea @ 0x402AAD
          offset: 0x-30 @ 0x402AAD
    basic block:
      and:
        mnemonic: idiv @ 0x402AF6
        mnemonic: cdq @ 0x402AF5
        number: 0xA @ 0x402AEE
```

The key part being
```
          mnemonic: lea @ 0x402AAD
          offset: 0x-30 @ 0x402AAD
```
Is this intended behaviour or a lucky hit?

data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
mike-hunhoff: **Short answer:** no we do not officially support negative numbers

**Because**: By default both the viv and IDA APIs default to two's complement for representing numbers so capa can't match `-1` because it only sees the two's complement form `0xFFFFFFFF`.

**However:** It appears that the viv API does not default to two's complement for representing offsets which allows `0x-30` to match. 

Unfortunately the IDA API does default to two's complement for representing offsets therefore, and I haven't tested this yet but I assume, this rule will match in viv but not in IDA.

I would probably consider this a bug as we need to establish a consistent representation of numbers, offsets, etc. across capa and the extractors.

Offsets in viv:
```
('Offset', -4)
('Offset', -20)
('Offset', -20)
('Offset', 0)
('Offset', -20)
('Offset', -4)
('Offset', 16)
('Offset', 0)
('Offset', 20)
('Offset', -4)
```

Offsets in IDA:
```
Python>insn = sark.code.instruction.Instruction(here())
Python>insn.operands[1].offset
0xfffffef8
```

williballenthin:
@re-fox i thought that it wouldn't work, but maybe it does, so i'm not 100% sure where we stand. i've opened an issue for us to discuss and solidify this behavior https://github.com/fireeye/capa/issues/197

re-fox:
Thanks for the feedback, I forget that you're using other analysis engines (like IDA).  So data types are important.  I read @mike-hunhoff comments, and it verified that the 2's compliment form did not trigger when running capa. 

I did add in another case where a compiler optimizes a `%10` checksum at the end of luhns (which doesn't affect the overall rule).

I suppose the best thing would be to wait before merging until data types are normalized across different analysis engine backends?

williballenthin:
yeah, if you have the patience for this, it would be best to finalize how we're normalizing offsets before merging this.

i'm probably going to experiment normalizing the IDA Pro backend. if that works, then we'll have an answer soon. id hope that sometime on monday we can move forward again.

data-manipulation/checksum/luhn/validate-credit-card-number-with-luhn-algorithm.yml
```diff
@@ -23,15 +25,30 @@ rule:
...
+            description: Final section returning checkum % 10
+          - and:
+            - mnemonic: shr
+            - mnemonic: imul
+            - number: 0x66666667
+            - number: 0x1f
+            - number: 0x2
+            - optional : 
+              - mnemonic: neg
+            description: Compiler optimized returning checkum % 10
```
williballenthin, line 54: i wonder if we should refactor this into its own `lib` rule "mod 10"?

re-fox:
Plenty of patience.  I've got more rule ideas and samples to dig through.

williballenthin:
in https://github.com/fireeye/capa/pull/208 we finalized that offsets can be negative, so your rules should work as written. mind double checking these match as expected? then we can merge.

re-fox:
I validated the rule worked.  Got back the expected result with negative offset.  Thanks!

```
    basic block:
      or:
        and:
          mnemonic: lea @ 0x402AAD
          offset: 0x-30 @ 0x402AAD
```

PR #54, opened 2020-07-27
add a couple new rules derived from Cl0p ransomware

williballenthin:
unfortunately, haven't found a public sample of the ransomware yet

### DIFF
--- a/nursery/acquire-debug-privileges.yml
+++ b/nursery/acquire-debug-privileges.yml
@@ -11,3 +11,5 @@ rule:
   features:
     - and:
       - string: "SeDebugPrivilege"
+      - optional:
+        - match: modify access privileges
--- a/nursery/acquire-debug-privileges.yml
+++ b/nursery/acquire-debug-privileges.yml
@@ -1,0 +1,13 @@
+rule:
+  meta:
+    name: acquire debug privileges
+    namespace: host-interaction/process/modify
+    author: william.ballenthin@fireeye.com
+    scope: basic block
+    att&ck:
+      - Privilege Escalation::Access Token Manipulation [T1134]
+    examples:
+      # - unpacked Cl0p ransomware
+  features:
+    - and:
+      - string: "SeDebugPrivilege"
--- a/nursery/import-public-key.yml
+++ b/nursery/import-public-key.yml
@@ -5,10 +5,12 @@ rule:
     author: william.ballenthin@fireeye.com
     scope: function
     examples:
-      # - unpacked Cl0p ransomware
+      - ffeae4a391a1d5203bd04b4161557227:0x4047A0
   features:
     - and:
-      - api: crypt32.CryptStringToBinary
-      - api: crypt32.CryptDecodeObjectEx
-      - api: crypt32.CryptAcquireContext
+      - api: advapi32.CryptAcquireContext
       - api: crypt32.CryptImportPublicKeyInfo
+      - optional:
+        - and:
+          - api: crypt32.CryptStringToBinary
+          - api: crypt32.CryptDecodeObjectEx
--- a/nursery/import-public-key.yml
+++ b/nursery/import-public-key.yml
@@ -1,0 +1,14 @@
+rule:
+  meta:
+    name: import public key
+    namespace: data-manipulation/encryption
+    author: william.ballenthin@fireeye.com
+    scope: function
+    examples:
+      # - unpacked Cl0p ransomware
+  features:
+    - and:
+      - api: crypt32.CryptStringToBinary
+      - api: crypt32.CryptDecodeObjectEx
+      - api: crypt32.CryptAcquireContext
+      - api: crypt32.CryptImportPublicKeyInfo
--- a/nursery/terminate-process-by-name.yml
+++ b/nursery/terminate-process-by-name.yml
@@ -1,0 +1,14 @@
+rule:
+  meta:
+    name: terminate process by name
+    namespace: host-interaction/process/terminate
+    author: william.ballenthin@fireeye.com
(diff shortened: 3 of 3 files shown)


williballenthin:
@re-fox can't add you as a reviewer yet, but please let me know if you have any thoughts.

re-fox:
Thank you for sharing, I can run these against my corpus and see if there is a shareable version of the malware.

nursery/terminate-process-by-name.yml
```diff
@@ -0,0 +1,14 @@
...
+    author: william.ballenthin@fireeye.com
+    scope: function
+    examples:
+      # - unpacked Cl0p ransomware
+  features:
+    - and:
+      - match: terminate process
+      - match: enumerate processes
+      - or:
+        - offset: 0x24 = pe.szExeFile (x32)
```
mr-tz, line 14: nice!

nursery/import-public-key.yml
```diff
@@ -0,0 +1,14 @@
...
+    author: william.ballenthin@fireeye.com
+    scope: function
+    examples:
+      # - unpacked Cl0p ransomware
+  features:
+    - and:
+      - api: crypt32.CryptStringToBinary
+      - api: crypt32.CryptDecodeObjectEx
+      - api: crypt32.CryptAcquireContext
+      - api: crypt32.CryptImportPublicKeyInfo
```
mr-tz, line 14: are all of these APIs necessary or can we loosen it up to catch similar samples?

nursery/acquire-debug-privileges.yml
```diff
@@ -0,0 +1,13 @@
...
+    namespace: host-interaction/process/modify
+    author: william.ballenthin@fireeye.com
+    scope: basic block
+    att&ck:
+      - Privilege Escalation::Access Token Manipulation [T1134]
+    examples:
+      # - unpacked Cl0p ransomware
+  features:
+    - and:
+      - string: "SeDebugPrivilege"
```
re-fox, line 13: This rule looks like it would sit alongside of `modify access privileges` nicely.  

I was thinking of ways to tighten the rule down, but (maybe by looking for `LookupPrivilegeValue`) but this is already covered in other rules.  The string `SeDebugPrivilege` shouldn't really appear errantly in many executables without the intent to adjust privs.  So this rule is good to go in my opinion.

nursery/import-public-key.yml
re-fox: I looked up a reference sample in an old copy of cryptolocker (`ffeae4a391a1d5203bd04b4161557227`).  The functionality of the rule was broken up into 2 functions.

The first function @ `00404c20` will call `CryptStringToBinary` and then `CryptDecodeObjectEx`.  
The second function @ `004047a0` will call `CryptAcquireContext` then ` CryptImportPublicKeyInfo`

Breaking them up to help trigger on this variant may lose the original intent of the rule.  

`CryptImportPublicKeyInfo` seems to be the primary functionality that the rule is triggering on. So we could probably reduce the rule more to be:

```
      - and:
         - api: crypt32.CryptAcquireContext
         - api: crypt32.CryptImportPublicKeyInfo
         - optional: 
             - and:
                - api: crypt32.CryptDecodeObjectEx
                - api: crypt32.CryptStringToBinary
```

The other functionality of `CryptStringToBinary` would be nice to have as an optional indicator though as it provides nice insight into the behavior.

williballenthin: i like the suggestions here a lot.

since i was working off a single example (the malware i was currently reversing) i wasn't sure how the rule generalized. the updates make a lot of sense. i'll fix this PR.

nursery/acquire-debug-privileges.yml
williballenthin: i would like to put it next to `match: modify access privileges`, though in my experience, there's often a helper function that does the lookup and modify given a string argument. 

unfortunately, right now, we can't match across functions in capa, so catching all this context would be difficult. still, i'll add an `optional: match: modify access privileges` to catch this when feasible.

nursery/terminate-process-by-name.yml
williballenthin: i wonder if we should have a feature/characteristic that allows us to switch based on the bitness of the file. something like:

```yml
- arch:
  - x32:
    - offset: 0x24 = pe.szExeFile
  - x64:
    - offset: 0x?? = pe.szExeFile
```

williballenthin: or maybe make the bitness part of the offset feature?

```yml
- or:
  - offset/32: 0x24 = pe.szExeFile
  - offset/64: 0x?? = pe.szExeFile
```

this would be an easy add, since we could just emit an additional feature for each offset that's tagged with the bitness.

williballenthin: https://github.com/fireeye/capa/issues/210

nursery/acquire-debug-privileges.yml
williballenthin: added in 95d6d53

nursery/import-public-key.yml
williballenthin: added in 36bbb25

DeerSpotter:
> unfortunately, haven't found a public sample of the ransomware yet

I used to work for a company that was affected by this. Right before cl0p was launched we found a .l0cky ransomware on our systems that escaped firewall detection from 2016. Which means however this ransomware works it bypasses the active directory firewall somehow.

DeerSpotter:
You can get a public sample from Netzsch they will gladly help you for this to not happen to anyone again.

PR #61, opened 2020-07-30
Create capture-network-configuration-via-ipconfig.yml

re-fox:
Working through the copy of Arkei that was uploaded to `capa-testfiles` (hash: `7204e3efc2434012e13ca939db0d0b02`)

The malware will make a `system` call via the C runtime `msvcr100` to run `ipconfig > ipconfig.txt`.   Seen in the following screenshot.

image

When running the rule against the sample it looks like the regex in the `string` returns the last occurrence.  This isn't a problem, it just led to an unexpected result.

Where rather than matching the actual command `ipconfig >ipconfig.txt`, we matched the `ipconfig.txt`.
```
    string: ipconfig.txt @ 0x403043
    api: msvcr100.system @ 0x403035
```
The full results are shown here.  I added in a small description as this rule that can call out the format strings and banners, in the case that anyone will want to use this rule for categorization.  

I can see this rule being built out more (other ways to shell out via ShellExecute, WinExec, etc...)

The full results:
```
capture network configuration via ipconfig
namespace  collection/network
author     @_re_fox
scope      basic block
examples   -7204e3efc2434012e13ca939db0d0b02:0x403028
basic block @ 0x403028
  and:
    string: ipconfig.txt @ 0x403043
    api: msvcr100.system @ 0x403035
    optional:
      and:
        string: ipconfig.txt @ 0x403043
        string: [Windows IP Configuration]
 = Arkei Stealer Filename and Banner @ 0x403028
```

### DIFF
--- a/collection/network/capture-network-configuration-via-ipconfig.yml
+++ b/collection/network/capture-network-configuration-via-ipconfig.yml
@@ -1,0 +1,17 @@
+rule:
+  meta:
+    name: capture network configuration via ipconfig
+    namespace: collection/network
+    author: "@_re_fox"
+    scope: basic block
+    examples:
+      -7204e3efc2434012e13ca939db0d0b02:0x403028
+  features:
+    - and:
+      - string: /ipconfig(\.exe)?/i
+      - api: msvcr100.system
+      - optional:
+        - and:
+          - string: ipconfig.txt
+          - string: /\[Windows IP Configuration\]/
+            description: Arkei Stealer Filename and Banner
--- a/collection/network/capture-network-configuration-via-ipconfig.yml
+++ b/collection/network/capture-network-configuration-via-ipconfig.yml
@@ -5,7 +5,7 @@ rule:
     author: "@_re_fox"
     scope: basic block
     examples:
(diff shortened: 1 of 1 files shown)


williballenthin:
i think the regex weirdness is due to https://github.com/fireeye/capa/issues/159

looks like capa identifies two strings that contain `ipconfig.txt`:
image

and due to the existing limitation, only one of them gets returned.

collection/network/capture-network-configuration-via-ipconfig.yml
```diff
@@ -0,0 +1,17 @@
...
+    name: capture network configuration via ipconfig
+    namespace: collection/network
+    author: "@_re_fox"
+    scope: basic block
+    examples:
+      - 7204e3efc2434012e13ca939db0d0b02:0x403028
+  features:
+    - and:
+      - string: /ipconfig(\.exe)?/i
+      - api: msvcr100.system
```
williballenthin, line 12: i agree that here is a place that we can expand.

one way of doing this might be to introduce a new rule "run shell command" (or something) and then reference it from here. lets do that when we have more examples.

williballenthin:
thank you!

PR #64, opened 2020-07-31
New rules from 03B236B23B1EC37C663527C1F53AF3FE

mike-hunhoff:
Adding new rules identified in sample with MD5 hash 03B236B23B1EC37C663527C1F53AF3FE.


data-manipulation/encryption/encrypt-or-decrypt-via-wincrypt.yml
```diff
@@ -10,10 +10,11 @@ rule:
...
-        - api: CryptGenKey
-        - api: CryptImportKey
       - or:
         - api: CryptEncrypt
         - api: CryptDecrypt
+      - optional:
+        - or:
+          - api: CryptAcquireContext
+          - api: CryptGenKey
+          - api: CryptImportKey
```
mike-hunhoff, line 20: Making this block optional so the rule can match samples that call these APIs in a different function.

host-interaction/session/get-session-integrity-level.yml
```diff
@@ -0,0 +1,13 @@
+rule:
+  meta:
+    name: get session integrity level
+    namespace: host-interaction/session
```
mike-hunhoff, line 4: This might not be the best name or correct location. Open to alternatives.

mike-hunhoff:
Sample binary for these rules in PR https://github.com/fireeye/capa-testfiles/pull/7.

data-manipulation/encryption/encrypt-or-decrypt-via-wincrypt.yml
williballenthin: nice, i encountered the same issue this week

PR #66, opened 2020-07-31
Create gather-firefox-profile-information.yml

re-fox:
This is a rule I've been toying with while looking at different infostealers.  

I personally don't care much for rules that are just a collection of strings, but many of the malware families (that I've been looking at while building this rule) have custom wrapper functions around system calls that do the work.  Which make tightening down a rule difficult.

Any additional feedback is very welcome, some of the regexes are redundant, but I left them for clarification and debugging.  Those can be easily removed if they aren't helping.


williballenthin:
This looks good to me.

I agree that a rule containing just a bunch of strings doesn't feel that elegant; however, I'd encourage you to think of capa as a system that lets you express intuitive patterns quickly. While we don't want to settle for anything inferior, collecting a thorough corpus and making tweaks should be as easy as possible. So, lets record what you have here, and then fix it up as we find more samples.

PR #68, opened 2020-08-03
Update validate-credit-card-number-with-luhn-algorithm.yml

re-fox:
This is the variant of the luhn checksum where the authors used negative numbers as part of the digital root lookup.  This version is used in the sample `ce0296e2d77ec3bb112e270fc260f274`.  I didn't upload this sample to `capa-testfiles` since it's a one off, but I'm able to if it is part of the standard process.

There are some slightly different variants that I have listed that I'll be pushing coverage for in the next couple weeks.

I can refactor the rule to make it a bit more compact, but right now I have each variant of the rule listed so the rule isn't a nightmare to debug or figure out what it is intended to trigger on.


williballenthin:
thanks!

PR #69, opened 2020-08-03
Create encrypt-data-using-camellia.yml

re-fox:
This rule covers the Camellia cipher covered here -> https://github.com/gpg/libgcrypt/blob/355f5b7f69075c010fe33aa5b10ac60c08fae0c7/cipher/camellia.c .  I'll upload the sample to capa-testfiles and I have a couple more planned rules off that sample.  Rather than uploading them in bulk, I'll post them up one at a time for review to make the PR's a bit more painless.

The tables are larger than what capa allows (large tables of DWORDs) so I converted them to bytes and truncated at a 256 cutoff.

### DIFF
--- a/data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
+++ b/data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
@@ -1,0 +1,35 @@
+rule:
+  meta:
+    name: encrypt-data-using-camellia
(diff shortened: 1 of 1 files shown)


williballenthin:
> so I converted them to bytes and truncated at a 256 cutoff.

this makes sense

williballenthin:
thank you!

PR #71, opened 2020-08-03
Create hash-data-using-tiger.yml

re-fox:
Adding support for the tiger hashing algo that is available in libgcrypt -> https://github.com/gpg/libgcrypt/blob/master/cipher/tiger.c

Tiger contains large sbox entries, so they have been truncated to best fit the capa rule.  

In addition to the sbox constants, I included a small basic block method to catch on the key schedule.


williballenthin:
thank you!

PR #72, opened 2020-08-03
Create encrypt-data-using-blowfish.yml

re-fox:
Adding in support for blowfish as seen in libgcrypt -> https://github.com/gpg/libgcrypt/blob/master/cipher/blowfish.c

SBoxes are truncated at 256 bytes.


mr-tz:
🙇  thank you!

PR #73, opened 2020-08-03
adding misc new rules and rule updates

mike-hunhoff:
Misc new rules and rule updates 🚀

### DIFF
--- a/nursery/set-graphical-window-text.yml
+++ b/nursery/set-graphical-window-text.yml
@@ -1,7 +1,7 @@
 rule:
   meta:
     name: set console window title
-    namespace: host-interaction/gui/console/set-text
+    namespace: host-interaction/gui/console
     author: michael.hunhoff@fireeye.com
     scope: function
   features:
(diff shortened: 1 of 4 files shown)


host-interaction/gui/taskbar/hide/hide-the-windows-taskbar.yml
```diff
@@ -9,5 +9,4 @@ rule:
   features:
     - and:
       - match: find taskbar
-      - api: user32.ShowWindow
-      - number: 0x0 = SW_HIDE
+      - match: hide graphical window
```
mike-hunhoff, line 12: move to separate rule so we can detect hiding other types of windows

nursery/set-graphical-window-text.yml
```diff
@@ -0,0 +1,9 @@
+rule:
+  meta:
+    name: set console window title
+    namespace: host-interaction/gui/console/set-text
```
williballenthin, line 4: since there aren't tons of sub-namespaces here, lets use `host-interaction/gui/console`

PR #74, opened 2020-08-03
add doc around arch flavors of offset/number

williballenthin:


doc/format.md
```diff
@@ -358,11 +361,18 @@ This should not be a stack offset.
 The parameter is a number; if prefixed with `0x` then in hex format, otherwise, decimal format. Negative offsets are supported.
 An offset can be followed by an optional description.
 
+If the number is only relevant for a particular architecture, then you can use one of the architecture flavors: `number/x32` or `number/x64`.
```
mr-tz, line 364: fix:
If the offset is only relevant for a particular architecture, then you can use one of the architecture flavors: `offset/x32` or `offset/x64`.

doc/format.md
```diff
@@ -358,11 +361,18 @@ This should not be a stack offset.
...
 An offset can be followed by an optional description.
 
+If the number is only relevant for a particular architecture, then you can use one of the architecture flavors: `number/x32` or `number/x64`.
+
 Examples:
 
-    offset: 0xC
-    offset: 0x14
-    offset: -0x4
+```yaml
```
mr-tz, line 368: neat, should we update all docs like this?

PR #76, opened 2020-08-05
Create bypass-uac-via-appinfo-alpc.yml

agithubuserlol:


agithubuserlol:
Uploaded wrong rule version. Opened https://github.com/fireeye/capa-rules/pull/76 with correct.

PR #77, opened 2020-08-05
Create Update bypass-uac-via-appinfo-alpc.yml

agithubuserlol:
Correct version of rule 👍


williballenthin:
sample added to capa-testfiles in https://github.com/fireeye/capa-testfiles/pull/10

PR #93, opened 2020-08-10
Coverage for alternate Luhn variant and moving rules to lib

re-fox:
A little bit of file movement in this PR.  

Moving the existing luhn rule to lib and creating a new luhn rule in lib.  The rule that exists in `data-manipulation/checksum/luhn` is now just a reference to matching either one of those implementations.

More about this was discussed under https://github.com/fireeye/capa-rules/issues/90.  Which, in my opinion, can be closed after this PR.  

Thanks for the input and suggestions.


mr-tz:
Thank you!

PR #96, opened 2020-08-13
Create check-for-sandbox-and-av-modules.yml

re-fox:
Simple rule looking for known sandbox and AV dlls.  

In most samples that I've looked at, the function will resemble the following 
image

Check if any exist and jump out if it finds one.  Calling GetModuleHandle at each basic block.

I also took a list from https://github.com/LordNoteworthy/al-khaser/blob/8c7a5f378617e01f28844a3ec7543eda7731abf4/al-khaser/AntiVM/Generic.cpp#L14 and combined it to cover most variations.

I know that not *all* of them are AV, but since most of them are, it felt like the best place to store this was under `anti-av`.  Let me know if this needs to be broken up into different sandboxes and placed into different directories, etc..

Thanks!

### DIFF
--- a/anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
+++ b/anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
@@ -1,0 +1,40 @@
+rule:
+  meta:
+    name: check for sandbox and av modules
+    namespace: anti-analysis/anti-vm/
+    author: "@_re_fox"
+    scope: basic block
+    examples:
+      - ccbf7cba35bab56563c0fbe4237fdc41:0x0040a4a0
+  features:
+    - and:
+      - api: GetModuleHandle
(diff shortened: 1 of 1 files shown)


anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
```diff
@@ -0,0 +1,40 @@
+rule:
+  meta:
+    name: check for sandbox and av modules
+    namespace: anti-analysis/anti-vm
```
williballenthin, line 4: would you please fix the namespace here?

re-fox: Fixed, thanks for catching the typo.

williballenthin:
awesome, thanks!

PR #97, opened 2020-08-13
Improve hash-data-using-murmur3 rule

Ana06:
Include more ways to detect MurmurHash3 algorithm.

For reference check the algorithm at:
https://en.wikipedia.org/wiki/MurmurHash

### DIFF
--- a/data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
+++ b/data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
@@ -28,3 +28,25 @@ rule:
         - and:
           - number: 0x87c37b91114253d5 = 64-bit c1 for 128-bit hash
           - number: 0x4cf5ad432745937f = 64-bit c2 for 128-bit hash
+        - basic block:
+          - and:
+            - count(mnemonic(shr)): 3
+            - number: 16
+            - number: 13
+            - optional:
+              - count(characteristic(nzxor)): 3 or more
+            description: 'hash >> 16; hash >> 13; hash >> 16'
+        - and:
+          # Group this two blocks under an `and` as on their own they are not
+          # unique enough and would cause false positives
+          - basic block:
+            - and:
+              - mnemonic: rol
+              - number: 15 = r1
+              description: k ROL r1
+          - basic block:
+            - and:
+              - mnemonic: rol
+              - number: 13 = r2
+              description: hash ROL r2
+
--- a/data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
+++ b/data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
@@ -9,44 +9,43 @@ rule:
     examples:
       - c66172b12971a329f8d5ff01665f204b:0x404A18
   features:
-    - and:
-      - or:
+    - or:
+      - and:
+        - number: 0x85ebca6b = 32-bit finalization mix constant 1
+        - number: 0xc2b2ae35 = 32-bit finalization mix constant 2
+      - and:
+        - number: 0xff51afd7ed558ccd = 64-bit finalization mix constant 1
+        - number: 0xc4ceb9fe1a85ec53 = 64-bit finalization mix constant 2
+      - and:
+        - number: 0xcc9e2d51 = c1 32-bit hash
+        - number: 0x1b873593 = c2 32-bit hash
+      - and:
+        - number: 0x239b961b = 32-bit c1 for 128-bit hash
+        - number: 0xab0e9789 = 32-bit c2 for 128-bit hash
+        - number: 0x38b34ae5 = 32-bit c3 for 128-bit hash
+        - number: 0xa1e38b93 = 32-bit c4 for 128-bit hash
+      - and:
+        - number: 0x87c37b91114253d5 = 64-bit c1 for 128-bit hash
+        - number: 0x4cf5ad432745937f = 64-bit c2 for 128-bit hash
+      - basic block:
         - and:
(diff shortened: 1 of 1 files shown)


data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
```diff
@@ -28,3 +28,21 @@ rule:
...
+            - and:
+              - mnemonic: rol
+              - number: 15 = r1
+              description: k ROL r1
+          - basic block:
+            - and:
+              - mnemonic: rol
+              - number: 13 = r2
+              description: hash ROL r2
+        - basic block:
```
mr-tz, line 42: is this indented correctly?

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
```diff
@@ -28,3 +28,21 @@ rule:
         - and:
           - number: 0x87c37b91114253d5 = 64-bit c1 for 128-bit hash
           - number: 0x4cf5ad432745937f = 64-bit c2 for 128-bit hash
+        - and:
```
mr-tz, line 31: thoughts on adding the XOR part?

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
Ana06: yes, this block could be in a different function than the other two. The reason why I put the other two together is that `ROL 13` and `ROL 15` are not that unique and will probably cause false positives. But together this should be unique enough.

Should I add a comment or clarify this somehow?

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
Ana06: do you mean adding to the last block `count(mnemonic(xor)): 3`? Should we add this as optional? I think that basic block is already unique enough to conclude this is MurmurHash3. But still a good idea to mark those 🤔

Ana06:
I am thinking now... Should I use `count(...): 3` or more instead of `count(...): 3`? 🤔

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
mr-tz: as is, this block will be a stand-alone detection though (its grouped under the top-level `or`)

I'm not sure about using `3 or more` for the count. I'd probably leave it for now.

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
mr-tz: I think including it as a (optional) new rule part makes sense, maybe even using the `match` on XOR encode

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
```diff
@@ -28,3 +28,21 @@ rule:
...
+              - mnemonic: rol
+              - number: 13 = r2
+              description: hash ROL r2
+        - basic block:
+          - and:
+            - count(mnemonic(shr)): 3
+            - number: 16
+            - number: 13
+            description: 'hash >> 16; hash >> 13; hash >> 16'
+
```
mr-tz, line 48: unrelated to this PR, can you please clean up the top-level `and`/`or`, while you're at it?

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
Ana06: @mr-tz 
> as is, this block will be a stand-alone detection though (its grouped under the top-level or)

yes, do you think it shouldn't? I think finding `hash >> 16; hash >> 13; hash >> 16` is enough to say this is most likely Murmur Hashing.

mr-tz: Ok, in that case we're good. Thanks for the clarifications.

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
Ana06: @mr-tz 

> maybe even using the match on XOR encode

Do you mean the `encode data using XOR`? That rule includes `characteristic: tight loop`, which doesn't apply here.

mr-tz: Ok, then just XOR (as optional per your discretion)

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
Ana06: I'll add a comment so that it is clear.

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
```diff
@@ -28,3 +28,25 @@ rule:
         - and:
           - number: 0x87c37b91114253d5 = 64-bit c1 for 128-bit hash
           - number: 0x4cf5ad432745937f = 64-bit c2 for 128-bit hash
+        - basic block:
+          - and:
+            - count(mnemonic(shr)): 3
+            - number: 16
+            - number: 13
+            - optional:
+              - count(characteristic(nzxor)): 3 or more
```
Ana06, line 37: Here we need `3 or more` because there are more `nzxor` in murmur which may be matched as well. I am looking to a sample whose murmur3 implementation looks like the Wikipedia one and capa matches 4.

PR #102, opened 2020-08-20
fixed description key, and added x32 and x64 flavours

0ssigeno:
As title


mr-tz:
Great, thank you, again!

williballenthin:
thank you @0ssigeno!

PR #104, opened 2020-08-27
add new rules

mr-tz:
New rules from recent samples, including a ransomware (added in https://github.com/fireeye/capa-testfiles/pull/16)

### DIFF
--- a/data-manipulation/encryption/rsa/reference-public-rsa-key.yml
+++ b/data-manipulation/encryption/rsa/reference-public-rsa-key.yml
@@ -1,0 +1,29 @@
+rule:
+  meta:
+    name: reference public RSA key
+    namespace: data-manipulation/encryption/rsa
+    author: moritz.raabe@fireeye.com
+    scope: function
+    examples:
+      - b7b5e1253710d8927cbe07d52d2d2e10:0x417DF0
+  features:
+    - or:
+        # typedef struct _PUBLICKEYSTRUC {
+        #   BYTE   bType;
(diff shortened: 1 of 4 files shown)


host-interaction/mutex/check-mutex-and-exit.yml
```diff
@@ -2,16 +2,21 @@ rule:
...
   features:
     - and:
-        - match: create mutex
-        - api: GetLastError
-        - api: ExitProcess
-        - or:
+      - match: create mutex
+      - api: ExitProcess
+      - or:
+        - api: WaitForSingleObject
```
mr-tz, line 16: added this check, so had to move around the logic a bit

data-manipulation/encryption/rsa/reference-public-rsa-key.yml
```diff
@@ -0,0 +1,29 @@
...
+        #   DWORD bitlen;
+        #   DWORD pubexp;
+        # } RSAPUBKEY;
+        #
+      - bytes: 06 02 00 00 00 A4 00 00 52 53 41 31
+        #      ^^ bType = PUBLICKEYBLOB
+        #         ^^ bVersion = CUR_BLOB_VERSION
+        #            ^^ ^^ reserved
+        #                  ^^ ^^ ^^ ^^ aiKeyAlg = CALG_RSA_KEYX
+        #                              ^^ ^^ ^^ ^^ magic = RSA1 for public keys
```
williballenthin, line 29: very cool

PR #106, opened 2020-08-28
Fix #105 - Update encrypt-data-using-camellia

### LINKED ISSUE #105: FN: encrypt-data-using-camellia - 112f9f0e8d349858a80dd8c14190e620
## Summary

`encrypt-data-using-camellia` doesn't match on 112f9f0e8d349858a80dd8c14190e620

## Examples

112f9f0e8d349858a80dd8c14190e620

Reference:
 - https://krabsonsecurity.com/2020/08/22/bitrat-the-latest-in-copy-pasted-malware-by-incompetent-developers/

From a quick glance at the binary, the s-box from https://github.com/calccrypto/Encryptions/blob/master/Encryptions/Camellia_Const.h#L35 can be found at address 0x008cb760 (offset 0x004ca160)

## Possible improvements

I'm not familiar with Camellia or `encrypt-data-using-camellia` enough to understand specifically what the existing logic is looking for, but hopefully it's as simple as adding the s-box bytes (and maybe some of the sigma byte sequences as well) from above into the rule.

recvfrom:
Fixes #105 

Updates encrypt-data-using-camellia so that it works with 112f9f0e8d349858a80dd8c14190e620

I ran into some issues when getting these ready to submit - let me know if you'd like me to open tickets for any these:

When I run `capafmt.py` on these, it changes the bottom section to the following, which breaks the rule:
```
        - and:
          - string: /A09E667F3BCC908B/i
          description: sigma1_str
          - string: /B67AE8584CAA73B
          description: sigma2_str
          - string: /C6EF372FE94F82BE/i
          description: sigma3_str
          - string: /54FF53A5F1D36F1C/i
          description: sigma4_str
          - string: /10E527FADE682D1D/i
          description: sigma5_str
          - string: /B05688C2B3E6C1FD/i
          description: sigma6_str
```

When testing this rule with capa's `-v` parameter, the reported offsets don't appear to be correct:
```
python2.7 ~/.local/bin/capa -r ../capa-rules/ ../capa-testfiles/112f9f0e8d349858a80dd8c14190e620.exe_ -v
...
encrypt-data-using-camellia (2 matches)
namespace  data-manipulation/encryption/camellia
scope      basic block
matches    0x4CE3B2
           0x4CE3DD
...
```

Using yara, the actual Sbox offset is reported to be at 0x4ca160, and I confirmed this to be correct.

### DIFF
--- a/data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
+++ b/data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
@@ -2,16 +2,18 @@ rule:
   meta:
     name: encrypt-data-using-camellia
     namespace: data-manipulation/encryption/camellia
-    author: "@_re_fox"
+    author: '@_re_fox'
     scope: basic block
-    examples: 
+    examples:
       - 0761142efbda6c4b1e801223de723578:0x6541CD50
+      - 112f9f0e8d349858a80dd8c14190e620:0x004ca160
   features:
     - or:
-      - bytes: 00 70 70 70 00 82 82 82 00 2c 2c 2c 00 ec ec ec 00 b3 b3 b3 00 27 27 27 00 c0 c0 c0 00 e5 e5 e5 00 e4 e4 e4 00 85 85 85 00 57 57 57 00 35 35 35 00 ea ea ea 00 0c 0c 0c 00 ae ae ae 00 41 41 41 00 23 23 23 00 ef ef ef 00 6b 6b 6b 00 93 93 93 00 45 45 45 00 19 19 19 00 a5 a5 a5 00 21 21 21 00 ed ed ed 00 0e 0e 0e 00 4f 4f 4f 00 4e 4e 4e 00 1d 1d 1d 00 65 65 65 00 92 92 92 00 bd bd bd 00 86 86 86 00 b8 b8 b8 00 af af af 00 8f 8f 8f 00 7c 7c 7c 00 eb eb eb 00 1f 1f 1f 00 ce ce ce 00 3e 3e 3e 00 30 30 30 00 dc dc = sp1110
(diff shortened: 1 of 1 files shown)


williballenthin:
> Using yara, the actual Sbox offset is reported to be at 0x4ca160, and I confirmed this to be correct.

0x4ca160 is the file offset of the embedded s-box bytes:

image

capa reports the address of the basic block (since that's the rule's scope) that references the s-box. this is expected behavior:

image

image

capa reports it this way so that other rules can combine the logic of "has camellia sbox" with other features found in the same basic block/function.

williballenthin:
the `capafmt` thing is definitely a bug, thanks for pointing this out!

data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
```diff
@@ -33,3 +35,16 @@ rule:
...
+          - string: /B67AE8584CAA73B
+            description: sigma2_str
+          - string: /C6EF372FE94F82BE/i
+            description: sigma3_str
+          - string: /54FF53A5F1D36F1C/i
+            description: sigma4_str
+          - string: /10E527FADE682D1D/i
+            description: sigma5_str
+          - string: /B05688C2B3E6C1FD/i
+            description: sigma6_str
```
williballenthin, line 50: i can't believe these are actually included as ASCII strings, but you're certainly right!

image

williballenthin:
thanks @recvfrom ! please keep up the good work!

recvfrom:
@williballenthin ah ok, that makes sense then - thanks for the explanation! Should I update the rule so that the example references the basic block offset, then?

williballenthin:
yes, definitely. didn't realize that the reference was off. thanks!

PR #107, opened 2020-08-28
Update example offset in encrypt-data-using-camellia

recvfrom:
Per https://github.com/fireeye/capa-rules/pull/106#issuecomment-683137923


williballenthin:
b00m

thanks!

PR #110, opened 2020-08-31
update mappings for mbc 2.0

dzbeck:
Updated mappings in the anti-analysis namespace for Malware Behavior Catalog (MBC) 2.0.


williballenthin:
this is awesome! thanks for the hard work in updating these references, including a couple fixes to the ATT&CK mappings.

is there anything we can do to:
  1. make updates like this easier for you, and/or
  2. push ideas from capa to MBC?

dzbeck:
thanks! updates seem straight forward, and as far as pushing ideas from capa to mbc - as we review the rules in each of the namespaces, we'll be sure to add behaviors/methods as needed (in what will eventually be MBC 2.1).

PR #111, opened 2020-08-31
Update query-environment-variable.yml

re-fox:
Adding in support for some of the C runtime calls.  In this case, `getenv` as described here -> https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=vs-2019

I used an existing sample in the `capa-testfiles` repo to make it a bit easier.


williballenthin:
easy merge, thank you!

PR #112, opened 2020-08-31
Create compiled-with-py2exe.yml

re-fox:
Creating a rule to trigger on py2exe samples.  The sample in the rule has been uploaded to `capa-testfiles`.

It's a start to covering  https://github.com/fireeye/capa-rules/issues/11


williballenthin:
looks great, thanks!

PR #114, opened 2020-09-02
Add QuickLZ decompression rule

edeca:
Tested on 32-bit and 64-bit samples from the QuickLZ website, and on a malware sample that uses QuickLZ (ShadowPad). Good detection in both.

Am not sure of the best way to run this at scale and ensure there are no false positives, ideas welcomed.


re-fox:
Thanks for the contribution @edeca!  Nice use of description to put contextual data in the rule.

I'm running a quick check against the `capa-testfiles` to check for false positives.  If there are any I will post them up here so the rule can be adjusted.

re-fox:
I got only one result back in the `capa-testfiles` repo.

```
+------------------------+------------------------------------------------------------------------------------+
| md5                    | 6a352c3e55e8ae5ed39dc1be7fb964b1                                                   |
| sha1                   | e2cfb70c2757270288d10008fb4c8d3c2afd8fc3                                           |
| sha256                 | c69121a994ea8ff188510f41890208625710870af9a06b005db817934b517bc1                   |
| path                   | capa-testfiles/6a352c3e55e8ae5ed39dc1be7fb964b1.dll_                               |
+------------------------+------------------------------------------------------------------------------------+

+------------------------------------------------------+------------------------------------------------------+
| CAPABILITY                                           | NAMESPACE                                            |
|------------------------------------------------------+------------------------------------------------------|
| decompress data using QuickLZ                        | data-manipulation/compression                        |
+------------------------------------------------------+------------------------------------------------------+
```

Would you be able to download https://github.com/fireeye/capa-testfiles/raw/master/6a352c3e55e8ae5ed39dc1be7fb964b1.dll_ and check offset `0x10010DE0` to ensure it matches the intended logic of this rule?

edeca:
Looks fine - references to the file were added in [this commit](https://github.com/fireeye/capa/commit/8d16274b1e33de0a3b8b28dda89f4807b0a28e70) which is expected.

williballenthin:
can we add the `6a352c3e55e8ae5ed39dc1be7fb964b1` sample to the list of examples? this way the linter shouldn't warn about a missing example.

re-fox:
> Looks fine - references to the file were added in [this commit](https://github.com/fireeye/capa/commit/8d16274b1e33de0a3b8b28dda89f4807b0a28e70) which is expected.

Great, thanks for checking!  Everything looks good to me. 

Like @williballenthin said, adding ` - 6a352c3e55e8ae5ed39dc1be7fb964b1:0x10010DE0` to the example list will keep the linter from warning.

edeca:
Added hash to the list of examples. Linter now returns:

```
(nursery) decompress data using QuickLZ
  WARN: no violations: Graduate the rule
```

williballenthin:
lets graduate it then!

this means moving the rule from the nursery directory into the appropriate directory structure.

williballenthin:
🥳 thanks @edeca !

PR #116, opened 2020-09-02
Fix #115

### LINKED ISSUE #115: host-interaction/network/dns/resolve
## Summary

host-interaction/network/dns/resolve is overly broad, firing on any function that uses inet_addr (https://github.com/fireeye/capa-rules/blob/master/host-interaction/network/dns/resolve/resolve-dns.yml#L12). This function really has nothing to do with DNS (it merely converts an IP address string into the internal 32 bit int representation).

## Examples

For an example, see sample 3467147283c76dd00e2f6c34e1118ada3577405dd11229c7ae1ebd4e6a37fef2 function at 0x405E60

## Possible improvements

Likely just remove inet_addr from the list. I don't think it adds anything over the other function matches.

toomanybananas:
See #115 for explanation of the issue

### DIFF
--- a/host-interaction/network/dns/resolve/resolve-dns.yml
+++ b/host-interaction/network/dns/resolve/resolve-dns.yml
@@ -9,7 +9,6 @@ rule:
       - Practical Malware Analysis Lab 01-01.dll_:0x10001010
   features:
     - or:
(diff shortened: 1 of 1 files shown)


williballenthin:
thank you!

PR #123, opened 2020-09-16
adding new rules to nursery

mike-hunhoff:


nursery/create-restart-manager-session.yml
```diff
@@ -0,0 +1,9 @@
+rule:
+  meta:
+    name: create Restart Manager session
```
mr-tz, line 3: It may be useful to add a description and/or reference to how this is used in malware. I think most recently a lot of ransomware has been using this, e.g. see
- https://techdator.net/revil-ransomware-2-2-version-uses-windows-restart-manager-to-encrypt-more-files/ 
- https://www.carbonblack.com/blog/tau-threat-discovery-conti-ransomware/
- https://www.binarydefense.com/threat_watch/sodinokibi-ransomware-now-encrypts-more-files/

nursery/create-restart-manager-session.yml
```diff
@@ -0,0 +1,9 @@
...
+    namespace: host-interaction/process
```
mr-tz, line 4: Not sure if this is the best namespace here. At least in the context of ransomware, this resolves more around file access?

mr-tz (approved):
good stuff, thank you!

mr-tz (approved):
good stuff, thank you!

nursery/create-restart-manager-session.yml
mike-hunhoff: You are right! I saw this used in a ransomware sample to free up files locked by a running application. I'll get some documentation added to the rule.

nursery/create-restart-manager-session.yml
mike-hunhoff: Yeah, I was unsure where to place this rule. The Restart Manager itself can be used to stop applications so I figured throw it under `host-interaction/process` to start. However, like you said, within the context of ransomware the goal is file access, not process manipulation. 🤔

re-fox (approved):

PR #124, opened 2020-09-17
Create references-logon-banner.yml

re-fox:
Apologies on the earlier direct commit.  Mistake.

I recently was looking at a rust exe that changes the Windows logon banner.  The strings referencing the registry keys are passed to a function that modifies the keys so I can't tighten it down too much more.

Since the sample is in rust, the strings are not nice null terminated string.  It has to match a substring in a larger blob.  I'll upload the sample to test-files as well.

### DIFF
--- a/host-interaction/gui/logon/references-logon-banner.yml
+++ b/host-interaction/gui/logon/references-logon-banner.yml
@@ -1,0 +1,14 @@
+rule:
+  meta:
+    name: references logon banner
+    namespace: host-interaction/gui/logon
+    author: "@_re_fox"
+    scope: basic block
+    examples:
+      - c3341b7dfbb9d43bca8c812e07b4299f:0x4066FC
+  features:
+    - and:
+      - string: /\\Microsoft\\Windows\\CurrentVersion\\Policies\\System/
+      - or:
+        - string: /LegalNoticeCaption/
+        - string: /LegalNoticeText/


re-fox:
Referencing the above, here's a better explanation of the strings.  There is a string "blob" with offsets and length specified.  

The blob is `"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\SystemLegalNoticeCaptionLegalNoticeTextcmdnet user "`

```
0040670a  6a12               push    0x12
0040670c  6845574800         push    0x485745  {"LegalNoticeCaptionLegalNoticeTex…"}
00406711  6a39               push    0x39 {var_644}
00406713  680c574800         push    0x48570c {var_648}  {"SOFTWARE\Microsoft\Windows\Curre…"}
00406718  e8530d0000         call    sub_407470
```

Starting from the bottom string at `0x48570c`, it has a length of `0x39`, so the string will be 
```
'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\System'
```
and the second string has len `0x12` at offset `0x485745`
```
`LegalNoticeCaption`
```

A collection of strings feels like a weak rule, but I haven't seen these particular keys referenced in many samples and restricting it to a basic block should cut down on any potential FPs.  I'm running the rules against the existing capa-testfiles repo now to ensure no FPs.

re-fox:
Rule has been run against capa-testfiles with no FPs

mr-tz (approved):
very cool, I think given the specificity here this is good

PR #126, opened 2020-09-17
new nursery rules, update create-thread.yml

mike-hunhoff:


nursery/encode-data-using-base64-via-winapi.yml
```diff
@@ -7,4 +7,4 @@ rule:
   features:
     - and:
       - number: 1 = CRYPT_STRING_BASE64
-      - api: CryptStringToBinary
+      - api: CryptBinaryToString
```
mike-hunhoff, line 10: `CryptBinaryToString` = binary -> string (Base64 in this case), which I believe is the original intention of rule.

mr-tz: good catch!

mr-tz: what about the other constants, especially `0x6`, identified above?

mr-tz (approved):

nursery/encode-data-using-base64-via-winapi.yml
mike-hunhoff: `CRYPT_STRING_BASE64_ANY` and `CRYPT_STRING_ANY` are only relevant when converting from string to binary.

There are two additional flags that we could account for in this rule, `CRYPT_STRING_BASE64HEADER`and `CRYPT_STRING_BASE64REQUESTHEADER` as identified in MS docs for `CryptBinaryToString`:

Screen Shot 2020-09-18 at 8 14 02 AM

However, the intent of both `decode-data-using-base64-via-winapi.yml` and `encode-data-using-base64-via-winapi.yml` is to identify pure Base64 encoding/decoding so I don't think we should account for these additional flags in these specific rules.

mr-tz: Thanks for the clarification!

PR #130, opened 2020-09-18
Create compiled-with-rust.yml

re-fox:
Re: https://github.com/fireeye/capa-rules/issues/125

I was hoping to tie this rule down a bit more, but after looking at several samples I haven't found many dependable features.

At some point committing an initial rule makes more sense than looking at a problem forever.


williballenthin (approved):
> At some point committing an initial rule makes more sense than looking at a problem forever.

completely agree

PR #133, opened 2020-09-21
Fix #132 - Add encrypt-data-using-hc-128

### LINKED ISSUE #132: encrypt-data-using-hc-128
## Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your rule idea isn't already filed: [search](https://github.com/fireeye/capa-rules/issues?q=is%3Aissue+is%3Aopen+)

## Summary

encrypt-data-using-hc-128

## Examples

e69a8eb94f65480980deaf1ff5a431a6 from https://download.bitdefender.com/resources/files/News/CaseStudies/study/318/Bitdefender-TRR-Whitepaper-Maze-creat4351-en-EN-GenericUse.pdf

## Features

There don't appear to be any unique constants used by this algorithm, but the following pieces seem like they will be unique enough when combined to work.

From https://github.com/rost1993/hc128/blob/master/hc128.c
```
#define ROTR32(v, n)	((v >> n) | (v << (32 - n)))
...
#define F1(x)		(ROTR32(x,  7) ^ ROTR32(x, 18) ^ (x >>  3))
#define F2(x)		(ROTR32(x, 17) ^ ROTR32(x, 19) ^ (x >> 10))
...
static void
hc128_initialization_process(struct hc128_context *ctx)
...
	for(i = 16; i < (256 + 16); i++)
		ctx->w[i] = F2(ctx->w[i-2]) + ctx->w[i-7] + F1(ctx->w[i-15]) + ctx->w[i-16] + i;
```

### Namespace

Nursery for now, per https://github.com/fireeye/capa-rules/issues/131#issuecomment-695590953

recvfrom:
Fix #132 - Add encrypt-data-using-hc-128

### DIFF
--- a/nursery/encrypt-data-using-hc-128.yml
+++ b/nursery/encrypt-data-using-hc-128.yml
@@ -1,0 +1,48 @@
+rule:
+  meta:
+    name: encrypt-data-using-hc-128
+    namespace: data-manipulation/encryption/hc-128
+    author: '@recvfrom'
+    description: Looks for instruction mnemonics associated with initialization of the HC-128 stream cipher
+    scope: basic block
+    examples:
+      - e69a8eb94f65480980deaf1ff5a431a6:0x405D0D
+    references:
+      - https://download.bitdefender.com/resources/files/News/CaseStudies/study/318/Bitdefender-TRR-Whitepaper-Maze-creat4351-en-EN-GenericUse.pdf
+      - https://github.com/rost1993/hc128/blob/master/hc128.c
+  features:
+    - and:
+      - and:
+        - number: 0x0F = (v << (32 - 17)) from ROTR32(x, 17) in F2(x)
+        - mnemonic: shl
+      - and:
+        - number: 0x11 = (v >> 17) from ROTR32(x, 17) in F2(x)
+        - mnemonic: shr
+      - and:
+        - number: 0xD = (v << (32 - 19)) from ROTR32(x, 19) in F2(x)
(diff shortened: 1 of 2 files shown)


nursery/encrypt-data-using-hc-128.yml
```diff
@@ -0,0 +1,48 @@
+rule:
+  meta:
+    name: encrypt-data-using-hc-128
+    namespace: data-manipulation/encryption/hc-128
+    author: '@recvfrom'
+    description: Looks for instruction mnemonics associated with initialization of the HC-128 stream cipher
+    scope: basic block
+    examples:
```
mr-tz, line 8: can you please add this ATT&CK info here?
```yaml
att&ck:
```

mr-tz (commented):
Very good, thanks a lot for the rule and extensive documentation! If you have a moment, can you please add the ATT&CK info before merging.

I'm not too worried about FPs as this is a tight rule given the BB scope and features.

nursery/encrypt-data-using-hc-128.yml
recvfrom: Will do! I too feel pretty good about the rule after adding in the mnemonic counts.  Should the rule be moved out of the nursery, then?

mr-tz: Yes, please!

mr-tz (approved):
awesome, thank you!

PR #134, opened 2020-09-23
Graduate hash-data-using-fnv from the nursery

recvfrom:
Corresponding test file: https://github.com/fireeye/capa-testfiles/pull/22


williballenthin:
still looking for a good example (this one takes a ton of resources to analyze).

PR #137, opened 2020-09-30
use new description syntax

mr-tz:
I've enabled the descriptions again and reordered in most instances. Would appreciate a quick check that I didn't mess up indentations.


re-fox (approved):
These all look good to me.

mr-tz:
Thank you! 🙇

PR #138, opened 2020-09-30
update mappings for mbc (part 2)

dzbeck:
Added Malware Behavior Catalog (MBC) mappings in the capa c2, collection, and communication namespaces.


mr-tz (approved):
Awesome, looks good to me! Thank you very much.

PR #143, opened 2020-10-05
Create terminate-process-via-fastfail.yml

re-fox:
Some coverage for `__fastfail` intrinsic.

image


williballenthin (approved):
thanks!

mr-tz:
@re-fox have you observed this in malware? The hits I've seen so far are all on CRT library code and I'm thinking about removing (or adjusting) the rule.

PR #144, opened 2020-10-05
Create query-processor-features.yml

re-fox:
Another rule that uses the Ryuk sample.  

If you're wondering why the constants are out of order, that's the order in in the MSDN documentation. 
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent

### DIFF
--- a/host-interaction/hardware/cpu/query-processor-features.yml
+++ b/host-interaction/hardware/cpu/query-processor-features.yml
@@ -1,0 +1,44 @@
+rule:
+  meta:
+    name: query processor features
+    namespace: host-interaction/hardware/cpu
+    author: "@_re_fox"
+    scope: basic block
+    att&ck:
+      - Discovery::System Information Discovery [T1082]
+    references:
+      - https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent
+    examples:
+      - b87e9dd18a5533a09d3e48a7a1efbcf6:0x140007468
+  features:
+    - and:
+      - api: kernel32.IsProcessorFeaturePresent
+      - or:
+        - number: 25 = PF_ARM_64BIT_LOADSTORE_ATOMIC
+        - number: 24 = PF_ARM_DIVIDE_INSTRUCTION_AVAILABLE
+        - number: 26 = PF_ARM_EXTERNAL_CACHE_AVAILABLE
+        - number: 27 = PF_ARM_FMAC_INSTRUCTIONS_AVAILABLE
+        - number: 18 = PF_ARM_VFP_32_REGISTERS_AVAILABLE
+        - number: 7 = PF_3DNOW_INSTRUCTIONS_AVAILABLE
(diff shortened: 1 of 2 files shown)


host-interaction/hardware/cpu/query-processor-features.yml
```diff
@@ -0,0 +1,44 @@
+rule:
+  meta:
+    name: query processor features
```
mr-tz, line 3: thoughts on changing to singular: `query processor feature`? otherwise looks good, thanks!

re-fox: That sounds good to me.  I made the changes in `0b8cc29`

mr-tz:
Hm, in this sample this is part of `__report_gsfailure` which is called by `__security_check_cookie`.



I'm a little hesitant adding it due to many expected hits which could confuse users.

re-fox:
I understand completely.

The worst scenario would be adding too many features where the sheer number would distract/confuse users.  If a request (or issue) comes in we can always revive this PR, and potentially restrict the usage of it.  

As always, I appreciate the feedback and the time reviewing.  Until this comes back up, I'm good with closing the PR and not merging.

williballenthin:
how about removing the constant `17h` from the feature list?

or is this just a game of whack-a-mole and we should keep the rule disabled?

mr-tz:
I'd wait till we find a sample that uses this function maliciously and then narrow it down to those constants.

re-fox:
I'll close this out for now.  I can re-open if an issue or request comes in related to this API.

doomedraven:
hello guys, i have a small suggestion as i see that some malware started to use this as "sandbox" detection by checking if some specific feature is presented, we for sure don't need the whole list in, but some specific things could be useful 
```
push    PF_XMMI64_INSTRUCTIONS_AVAILABLE ; ProcessorFeature
call    IsProcessorFeaturePresent
```
Screenshot 2022-01-10 at 18 46 08

https://unit42.paloaltonetworks.com/matanbuchus-malware-as-a-service/

Screenshot 2022-01-10 at 18 48 55

https://unit42.paloaltonetworks.com/vatet-pyxie-defray777/3/


so i have simplified that rule to not match on everything that uses that call

```
# generated using capa explorer for IDA Pro
rule:
  meta:
    name: query processor feature
    namespace: host-interaction/hardware/cpu
    author: "@_re_fox"
    scope: basic block
    att&ck:
      - Discovery::System Information Discovery [T1082]
    references:
      - https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent
    examples:
      - 7fbaf7420943d4aa327bb82a357cd31ca92c7c83277f73a195d45bd18365cfce
  features:
    - and:
      - api: kernel32.IsProcessorFeaturePresent
      - number: 10 = PF_XMMI64_INSTRUCTIONS_AVAILABLE
```


more samples can be found here https://bazaar.abuse.ch/browse/signature/Matanbuchus/

PR #145, opened 2020-10-05
add rule delete volume shadow copies

mr-tz:


re-fox (approved):
You were faster than I am.  Looks good to me.  I was thinking when looking at these that It'd be nice if there was an immediate shellexecute after the string, but in this case I believe they are written to a .bat and executed in bulk.

mr-tz:
Thanks for adding the sample! The API calls are also resolved at run-time so I opted for just the strings here.

williballenthin (approved):
nice use of the ATT&CK namespace

PR #149, opened 2020-10-08
Initial set of packer rules

re-fox:
Handful of packer rules based on section names or strings.  

Submitting for cursory detection, the rules could be made more resilient with function features.  All supporting files are in https://github.com/fireeye/capa-testfiles/pull/25


williballenthin (approved):
looks great

PR #153, opened 2020-10-19
CI: Run rules with latest release

mr-tz:
Ensure that rules and most recent release are compatible.

Broken after enabling new `- description` entries, reported in https://github.com/fireeye/capa/issues/339.

Works with `- description` entries removed, see https://github.com/fireeye/capa-rules/actions/runs/315925603.

Plan to reenable descriptions and merge after new capa release.


williballenthin (approved):
good idea

PR #154, opened 2020-10-19
update mappings for mbc (part 3)

dzbeck:
Added Malware Behavior Catalog (MBC) and ATT&CK mappings in the data-manipulation namespace (plus a couple in anti-analysis and executable namespaces).


williballenthin (approved):
outstanding!

i love how MBC is growing in such a way that we can highlight the results in capa

williballenthin:
@dzbeck have you noticed any areas of MBC that has light coverage within capa? we can prioritize the development of rules there.

mr-tz (approved):
Great, thank you!

dzbeck:
@williballenthin nothing has jumped out as far as light coverage, but thinking about it now, you might want to look at some of the anti-status analysis behaviors (e.g., executable code obfuscation).

PR #158, opened 2020-10-22
add reference to DNS over HTTPS

yt0ng:


nursery/reference-doh.yml
```diff
@@ -0,0 +1,75 @@
+rule:
+  meta:
+    name: reference of DoH
```
williballenthin, line 3: lets rename this to "references DNS over HTTPS endpoints"

nursery/reference-doh.yml
```diff
@@ -0,0 +1,75 @@
...
+    namespace: communication/dns
+    author: markus.neis@swisscom.com
+    scope: file
+    references:
+      - https://github.com/curl/curl/wiki/DNS-over-HTTPS
+    examples:
+      # - 749e7becf00fccc6dff324a83976dc0d
```
williballenthin, line 10: can you update this line to be:
  1. not commented, and
  2. include the function offset in which the string is found?
    a. you can use `capa -v` to figure out what its recognizing
    b. format it like `749e7...:0x401000`

nursery/reference-doh.yml
```diff
@@ -0,0 +1,75 @@
...
+  features:
+    - or:
+      - string: /.*https://doh.seby.io:8443/dns-query.*/i
```
williballenthin, line 13: based on your example, i see that we do need to use regex here to account for application data being passed in the url. 

do you think its reasonable to remove the wildcard at the start of each of the strings here? regex matching is one area that capa is not very performant, and i think having a static anchor at the start of each pattern can maybe maintain performance as we add lots of patterns. 

image

williballenthin (changes requested):
i think this rule is really neat! i'm excited to get it merged. a little bit of feedback included inline. let me know if anything is unreasonable and we're happy to work with you.

nursery/reference-doh.yml
williballenthin: will need to rename the file, too

nursery/reference-doh.yml
williballenthin: and finally, would you add the example file to https://github.com/fireeye/capa-testfiles/

nursery/reference-doh.yml
yt0ng: Done!

nursery/reference-doh.yml
yt0ng: > can you update this line to be:
> 
> 1. not commented, and
> 2. include the function offset in which the string is found?
>    a. you can use `capa -v` to figure out what its recognizing
>    b. format it like `749e7...:0x401000`

Done!

yt0ng: > and finally, would you add the example file to https://github.com/fireeye/capa-testfiles/

Added in https://github.com/fireeye/capa-testfiles/pull/32

nursery/reference-doh.yml
mr-tz: We decided to use `re.search` to make rule writing more convenient.
https://github.com/fireeye/capa/blob/d5186f160dd461eb4fa779d9dc75f0766aedbf4f/capa/features/__init__.py#L151-L155

yt0ng: removed the wildcard at the start but apparently not necessary thanks @mr-tz

nursery/reference-doh.yml
yt0ng: Correction: had to close the previous PR and added now in https://github.com/fireeye/capa-testfiles/pull/33 adjusted filename

PR #159, opened 2020-10-22
Adler32 checksum rule

mwilliams31:


williballenthin:
would you add the example file to the https://github.com/fireeye/capa-testfiles/ repo?

then we can move this rule out of the nursery and merge the PR

williballenthin:
thanks @mwilliams31 !

PR #166, opened 2020-10-26
add resolve-function-by-hash

williballenthin:
adds most common lib/export hashes for ROR13. in theory would could add rules for each of the algos described in the blog post, though im concerned there would be hundreds of thousands of new rule statements that would slow things down.

https://www.fireeye.com/blog/threat-research/2012/11/precalculated-string-hashes-reverse-engineering-shellcode.html

### DIFF
--- a/nursery/resolve-function-by-hash.yml
+++ b/nursery/resolve-function-by-hash.yml
@@ -1,0 +1,22 @@
+rule:
+  meta:
+    name: resolve function by hash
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
These look good, although we probably don't want to do all possible hash/API combinations.

Instead I'd love to see a generic rule for detecting shellcode hashing functions.

PR #172, opened 2020-10-28
Create migrate-process-to-active-window-station.yml

williballenthin:
need example. will provide when i have a moment.


mr-tz:
Should we remove the `example` section and make this a nursery rule for now?

williballenthin:
yeah let’s do that

PR #183, opened 2020-11-18
Create read-virtual-disk.yml

re-fox:
This addresses https://github.com/fireeye/capa-rules/issues/177 which handles reading Virtual Disk files.  

I included the constants that make up the guid (shown below) as optional.  Due to the number of variations that could exist.  
```c
    static GUID VIRTUAL_STORAGE_TYPE_VENDOR_MICROSOFT_EX = { 0xEC984AEC ,0xA0F9, 0x47e9, 0x901F, 0x71415A66345B };
```

### DIFF
--- a/host-interaction/file-system/read/read-virtual-disk.yml
+++ b/host-interaction/file-system/read/read-virtual-disk.yml
@@ -1,0 +1,22 @@
+rule:
+  meta:
+    name: read virtual disk
+    namespace: host-interaction/file-system/read
+    author: "@_re_fox"
+    scope: function
+    references:
+      - https://github.com/vxunderground/VXUG-Papers/blob/main/Weaponizing%20Windows%20Virtualization/src.cpp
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
looks great, thanks for the references!

williballenthin:
> Due to the number of variations that could exist.

thoughts on a `guid: {....}` feature that expands into various ways of encoding it?

pro: makes it easier for rule authors
pro: could hook into an encyclopedia of GUIDs to enable use of symbolic names, too
con: single feature would expand into multiple features to support this, which could be complex to implement/display (thing about single string reference vs bunch of immediate constants).

re-fox:
That's a great idea.  

Rather than searching for string regexes related to GUIDs/CLSIDs/IIDs, mapping them to a feature would facilitate rapid capability discovery.  This would also assist in ID'ing COM objects as well.

PR #184, opened 2020-11-18
Graduate rule nursery/access-the-windows-event-log.yml

re-fox:
:mortar_board: 

Graduating nursery rule into namespace folder, and updated with example.


mr-tz:
Awesome, thanks! 👨‍🏫

PR #191, opened 2020-11-30
adding new rules for Gh0st

mike-hunhoff:


host-interaction/process/list/enumerate-processes.yml
```diff
@@ -11,9 +11,11 @@ rule:
...
-        - number: 0x2 = TH32CS_SNAPPROCESS
-      - api: kernel32.CreateToolhelp32Snapshot
       - api: kernel32.Process32First
       - api: kernel32.Process32Next
+      - optional:
+        - and:
+          - or:
+            - number: 0xF = TH32CS_SNAPALL
+            - number: 0x2 = TH32CS_SNAPPROCESS
+          - api: kernel32.CreateToolhelp32Snapshot
```
mike-hunhoff, line 21: making optional to increase rule hits for samples that obfuscate or do not directly call `CreateToolhelp32Snapshot`.

collection/screenshot/scan-screen.yml
```diff
@@ -0,0 +1,13 @@
+rule:
+  meta:
+    name: scan screen
```
mr-tz, line 3: What does `scan` refer to? How is this different to `capture screenshot`?

mr-tz (approved):
scan screen could use a description or new name, otherwise looks great

collection/screenshot/scan-screen.yml
mike-hunhoff: I used the word `scan` because of the way this sample uses `BitBlt` to loop through and record segments of the source screen. However, `capture screenshot` is probably a better fit. Should I merge this rule with `capture screenshot`?

mr-tz: If this is essentially screenshot capturing I'd say merge. If it's a different mechanism a separate rule with according doc works as well for me.

re-fox: I agree.  There will probably be another rule at somepoint that is dealing with screen recording (using the `capCaptureSequence` or the other multimedia libraries)

mike-hunhoff: I'll merge the two rules for now and then merge the PR. Thank you for the responses!

PR #196, opened 2020-12-11
Graduate compiled-with-go.yml

re-fox:
:mortar_board: 

Graduating `compiled-with-go` now that a sample is in `capa-testfiles`


williballenthin:
🎈

PR #202, opened 2020-12-16
update mappings for mbc (part 4)

dzbeck:
Added Malware Behavior Catalog (MBC) mappings in the host-interaction and impact namespaces (plus update to one micro objective name).

### DIFF
--- a/host-interaction/gui/set-application-hook.yml
+++ b/host-interaction/gui/set-application-hook.yml
@@ -4,12 +4,6 @@ rule:
     namespace: host-interaction/gui
     author: michael.hunhoff@fireeye.com
     scope: function
-    mbc:
-      - Collection::Hooking [F0003]
-      - Credential Access::Hooking [F0003]
-      - Defense Evasion::Hooking [F0003]
-      - Persistence::Hooking [F0003]
-      - Privilege Escalation::Hooking [F0003]
     examples:
       - Practical Malware Analysis Lab 12-03.exe_:0x401000
   features:
--- a/host-interaction/gui/set-application-hook.yml
+++ b/host-interaction/gui/set-application-hook.yml
@@ -4,6 +4,12 @@ rule:
     namespace: host-interaction/gui
     author: michael.hunhoff@fireeye.com
     scope: function
+    mbc:
+      - Collection::Hooking [F0003]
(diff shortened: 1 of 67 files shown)


host-interaction/process/allocate-thread-local-storage.yml
```diff
@@ -4,6 +4,11 @@ rule:
     namespace: host-interaction/process
     author: michael.hunhoff@fireeye.com
     scope: function
+    att&ck:
+      - Privilege Escalation::Process Injection::Thread Local Storage [1055.005]
+      - Defense Evasion::Process Injection::Thread Local Storage [1055.005]
```
williballenthin, line 9: i dont think this strictly indicates process injection. TLS has a valid use within a single process.

host-interaction/process/set-thread-local-storage-value.yml
```diff
@@ -4,6 +4,11 @@ rule:
     namespace: host-interaction/process
     author: michael.hunhoff@fireeye.com
     scope: function
+    att&ck:
+      - Privilege Escalation::Process Injection::Thread Local Storage [1055.005]
+      - Defense Evasion::Process Injection::Thread Local Storage [1055.005]
```
williballenthin, line 9: same TLS vs injection comment here.

host-interaction/gui/set-application-hook.yml
```diff
@@ -4,6 +4,12 @@ rule:
     namespace: host-interaction/gui
     author: michael.hunhoff@fireeye.com
     scope: function
+    mbc:
+      - Collection::Hooking [F0003]
+      - Credential Access::Hooking [F0003]
+      - Defense Evasion::Hooking [F0003]
+      - Persistence::Hooking [F0003]
+      - Privilege Escalation::Hooking [F0003]
```
williballenthin, line 12: the hook used here is a "normal" part of implementing a GUI program, so i think we should remove these references to privesc/persistence/etc.

williballenthin (changes requested):
three comments to tighten up some of the conclusions. open discussion if you feel differently.

also interested in feedback from any other maintainers (@mr-tz @mike-hunhoff @re-fox)

williballenthin:
should have said this first - thank you! tons of great updates in here

PR #204, opened 2020-12-17
Merging FNV hash rules

re-fox:
Per https://github.com/fireeye/capa-rules/issues/203.

Apologies on any confusion.  The primary difference between the rules were the use of the `optional` with the `offset_basis`.  Since @mr-tz's rule was more inclusive and didn't require these by default it would catch FNV-0.  I changed the numbers to `optional` in my rule and changed the name to catch the other variants.  

Thanks for the catch @mr-tz


data-manipulation/hashing/fnv/hash-data-using-fnv.yml
```diff
@@ -1,25 +1,33 @@
...
     scope: function
     references:
       - https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+      - http://isthe.com/chongo/tech/comp/fnv/
+      - https://create.stephan-brumme.com/fnv-hash/
     examples:
       - ad4229879180e267f431ac6666b6a0a2:0x14007B4D4
   features:
     - and:
       - characteristic: loop
```
mr-tz, line 18: I remember a sample where the loop was unrolled, so I left this out. Thoughts on making it optional? The constants are pretty unique features anyway.

mr-tz:
Thanks for the quick merge! This looks good to me.

data-manipulation/hashing/fnv/hash-data-using-fnv.yml
re-fox: That's great feedback, I hadn't seen that in a sample.  Changes made.

mr-tz:
On point as always! Thanks @re-fox!!!

Issue #174, opened 2020-11-02
enumerate-domain-trust-relationships


recvfrom:
## Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your rule idea isn't already filed: [search](https://github.com/fireeye/capa-rules/issues?q=is%3Aissue+is%3Aopen+)

## Summary

We could write a rule that looks for an EXE importing `DsEnumerateDomainTrustsA` / `DsEnumerateDomainTrustsW` or maybe that has the strings `nltest` / `nltest.exe`) `/domain_trusts`).

## Examples

38a41e8128ae3955d541c8a00a93de1cd10a01c58368c8254a35659f8627ba30

```
.text:0040E33D                 push    eax             ; DomainCount
.text:0040E33E                 lea     eax, [ebp+Domains]
.text:0040E341                 push    eax             ; Domains
.text:0040E342                 push    dword_46D1DC[esi] ; Flags
.text:0040E348                 push    0               ; ServerName
.text:0040E34A                 call    ds:DsEnumerateDomainTrustsA
.text:0040E350                 cmp     eax, 6BAh
.text:0040E355                 jz      short loc_40E3AA
.text:0040E357                 xor     ebx, ebx
.text:0040E359                 cmp     [ebp+DomainCount], ebx
.text:0040E35C                 jbe     short loc_40E398
.text:0040E35E                 xor     edi, edi
```

## Features

See above

## Additional context

N/A

### Namespace

collection/network

### References

https://attack.mitre.org/techniques/T1482/
https://twitter.com/sixdub/status/1321979931098812416/photo/1

### Other rule meta information

T1482

re-fox:
Providing another data point

I was recently looking at sample `3400a7df9ec3dc8283d5ac7accb6935691e93feda066cc46c6c04d67f7f87b2b` and it contains the following string which could also be used in the rule.

```
nltest.exe /domain_trusts /all_trusts
```

johnk3r:
Hello,

I will write this rule.

johnk3r:
Hey @mr-tz ,

after creating the rule, how does the validation and submission process work?

I didn't understand how to correctly fill in the examples. Would it be **md5: entrypoint**?

`
rule:
  meta:
    name: Detect Domain Trust Discovery 
    namespace: collection/network
    author: johnk3r
    scope: function
    att&ck:
      - Discovery::Domain Trust Discovery  [T1482]
    examples:
[...]

PR #206, opened 2020-12-21
Rule: Domain Trust Discovery and new IP checking domain

johnk3r:
#174


collection/network/domain-trust.yml
```diff
@@ -0,0 +1,15 @@
+rule:
+  meta:
+    name: Detect Domain Trust Discovery
```
mr-tz, line 3: Ideally this rules completes the sentence "this function may...". How about `get domain trust relationships`?

collection/network/domain-trust.yml
```diff
@@ -0,0 +1,15 @@
...
+    scope: function
+    att&ck:
+      - Discovery::Domain Trust Discovery  [T1482]
+    examples:
+      - 0796f1c1ea0a142fc1eb7109a44c86cb:0x40222F
+      - 0731679c5f99e8ee65d8b29a3cabfc6b:0x40408E
+  features:
+    - or:
+      - api: DsEnumerateDomainTrusts
+      - string: /nltest.exe/i
```
mr-tz, line 15: How about removing the `.exe` to be a little more flexible?
Also should we add `DOMAIN_TRUSTS` as `nltest` can be used for various tasks?

collection/network/capture-public-ip.yml
```diff
@@ -23,3 +23,4 @@ rule:
           - string: /checkip\.amazonaws\.com/
           - string: /icanhazip\.com/
           - string: /wtfismyip\.com/text/
+          - string: /api\.myip\.com/
```
mr-tz, line 26: 👍

mr-tz (changes requested):
Great, thank you. I only have a few comments.

johnk3r:
Hello @mr-tz ,

tks for the suggestions. I changed the logic of the rule to improve detection and avoid false positives.

The only thing is to note that leaving only "nltest" detection did not occur. So I followed some examples of rules that use the character "*".

Also, if the malware contains two spaces (or more)  between commands the detection does not occur. So I left the strings separately. Do you have any smarter idea to avoid this "bypass"?

mr-tz:
That looks great. I tried the below modified rule on the test samples:
- combines `nltest` and domain arguments under an `and` statement
- removed trailing `.*` from `nltest` (capa assumes leading and trailing wildcards)
- escaped `/` character

```yaml
  features:
    - or:
      - and:
        - string: /nltest/i
        - or:
          - string: /\/domain_trusts/i
          - string: /\/dclist/i
          - string: /\/all_trusts/i
      - api: DsEnumerateDomainTrusts
```

What do you think of these changes?

johnk3r:
Great.

I adjusted the rule again.

mr-tz (approved):
lgtm! thank you!

PR #207, opened 2020-12-21
Improve persistence detection

0x534a:
Hi all!

I suggest to improve the persistence detection capa rules with this PR.

The suggested improvements include the following:

- expanding the scope to also detect typical command lines (AppInit DLLs, GinaDLL, Windows service)
- adding more registry keys (AppInit DLLs, Run keys)

This changes will also make the rules more consistent.

In addition, I added two more rules to detect persistence via

- Winlogon Helper DLLs [T1547.004]
- Scheduled task by using the command line [T1053.005]

The test-files repository will be updated accordingly with the added example files of the suggested rules afterwards.

Best,
Simon


0x534a:
According PR for updating the test files can be found [here](https://github.com/fireeye/capa-testfiles/pull/49)

persistence/registry/appinitdlls/persist-via-appinit_dlls-registry-key.yml
```diff
@@ -9,11 +9,18 @@ rule:
     examples:
       - Practical Malware Analysis Lab 11-02.dll_:0x1000158b
   features:
```
mr-tz, line 11: Inline comments got messy on this? What you think of this?
```yaml
  features:
    - or:
      - and:
        - or:
          - match: set registry value
          - number: 0x80000002 = HKEY_LOCAL_MACHINE
        - or:
          - string: /Software\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
          - string: /Software\\Wow6432Node\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
        - string: /AppInit_DLLs/i
      - and:
        - match: create process
        - or:
          - string: /REG ADD[\s\"\\a-z_]{5,20}Software\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
          - string: /REG ADD[\s\"\\a-z_]{5,20}Software\\Wow6432Node\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
        - string: /AppInit_DLLs/i

```

persistence/registry/ginadll/persist-via-ginadll-registry-key.yml
```diff
@@ -9,11 +9,13 @@ rule:
     examples:
       - Practical Malware Analysis Lab 11-01.exe_:0x401000
   features:
-    - and:
-      - or:
+    - or:
```
mr-tz, line 12: I think there's a similar problem to the appinit rule here, i.e. that the `create process` block misses the `GinaDLL` string feature

persistence/registry/run/persist-via-run-registry-key.yml
```diff
@@ -9,20 +9,35 @@ rule:
...
-            - api: advapi32.RegOpenKey
-            - api: advapi32.RegOpenKeyEx
-          - or:
-            - api: advapi32.RegSetValue
-            - api: advapi32.RegSetValueEx
-          - or:
-            - number: 0x80000001 = HKEY_CURRENT_USER
-            - number: 0x80000002 = HKEY_LOCAL_MACHINE
-        - string: /Software\\Microsoft\\Windows\\CurrentVersion\\Run/i
+          - match: set registry value
```
mr-tz, line 17: nice!

persistence/registry/run/persist-via-run-registry-key.yml
```diff
@@ -9,20 +9,35 @@ rule:
...
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\Run/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\RunOnce/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\RunOnceEx/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Shell Folders/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\User Shell Folders/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\RunServices/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\RunServicesOnce/i
+          - string: /Software\\Microsoft\\Windows\\CurrentVersion\\Policies\\Explorer\\Run/i
+          - string: /Software\\Microsoft\\Windows NT\\CurrentVersion\\Windows\\load/i
+          - string: /System\\CurrentControlSet\\Control\\Session Manager\\BootExecute/i
```
mr-tz, line 30: These now capture more than just the Run keys (which is consistent with the att&ck technique). Should we rename the rule to `persist via registry key` or similar?

persistence/registry/appinitdlls/persist-via-appinit_dlls-registry-key.yml
```diff
@@ -9,11 +9,18 @@ rule:
...
-      - string: /SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
-      - string: AppInit_DLLs
+          - string: /Software\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
+          - string: /Software\\Wow6432Node\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
+        - string: /AppInit_DLLs/i
+      - and:
+        - match: create process
+        - or:
+          - string: /REG ADD[\s\"\\a-z_]{5,20}Software\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
+          - string: /REG ADD[\s\"\\a-z_]{5,20}Software\\Wow6432Node\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
```
mr-tz, line 26: Also, I wonder if we should include the `Software...` strings once instead of duplicating them. Adding a new block with `- string: /REG ADD/i` and the `create process` should work for this.

persistence/registry/run/persist-via-run-registry-key.yml
```diff
@@ -9,20 +9,35 @@ rule:
...
       - and:
         - match: create process
-        - string: /REG ADD[\s\"\\a-z_]{5,20}Software\\Microsoft\\Windows\\CurrentVersion\\Run/i
+        - or:
+          - string: /REG ADD[\s\"\\a-z_]{5,20}Software\\Microsoft\\Windows\\CurrentVersion\\Run/i
```
mr-tz, line 34: see previous comment on duplicating these strings

persistence/scheduled-tasks/schedule-task-via-commandline.yml
```diff
@@ -0,0 +1,18 @@
...
+      - Persistence::Scheduled Task/Job::Scheduled Task [T1053.005]
+    examples:
+      - 79cde1aa711e321b4939805d27e160be:0x401440
+  features:
+    - and:
+        - match: create process
+        - or:
+          - and:
+            - string: /schtasks/i
+            - string: //create /i
```
mr-tz, line 17: `/` should be escaped?

persistence/scheduled-tasks/schedule-task-via-itaskscheduler.yml
```diff
@@ -5,7 +5,7 @@ rule:
     author: moritz.raabe@fireeye.com
     scope: function
     att&ck:
-      - Persistence/Scheduled Task/Job/Scheduled Task [T1053.005]
+      - Persistence::Scheduled Task/Job::Scheduled Task [T1053.005]
```
mr-tz, line 8: good catch

mr-tz (changes requested):
Great, thanks a lot. I've added a couple of comments. This is a good opportunity to clean these rules up a bit / optimize the logic.
If you're not as thrilled about refactoring these, I'm happy to do it and discuss my proposed changes with you afterwards 😄

persistence/registry/ginadll/persist-via-ginadll-registry-key.yml
0x534a: Thank you for your comments @mr-tz! The `GinaDLL` string is already part of the Regex pattern `/REG ADD[\s\"\\a-z_]{5,20}Software\\Microsoft\\Windows NT\\CurrentVersion\\Winlogon\\GinaDLL/i`. I think I wanted to do the same for the AppInit rule but missed it somehow. Nevertheless, let's keep it consistent and I will change it in the upcoming commit.

persistence/scheduled-tasks/schedule-task-via-commandline.yml
0x534a: Based on my understanding and on the current capa implementation this works as expected without escaping the `/` (see [capa code](https://github.com/fireeye/capa/blob/master/capa/features/__init__.py#L129)).

mr-tz: right, my bad

persistence/registry/appinitdlls/persist-via-appinit_dlls-registry-key.yml
0x534a: Yeah, this makes sense. The commits 7d550ec and 13e454a should implement your suggestion for all of the rules that are part of my PR.

mr-tz (commented):
Great, thank you! These look good to me. A small improvement would be to pull up the duplicated strings, e.g.

```yaml
    - and:
      - or:
        - string: /Software\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
        - string: /Software\\Wow6432Node\\Microsoft\\Windows NT\\CurrentVersion\\Windows/i
      - string: /AppInit_DLLs/i
      - or:
        - match: set registry value
        - number: 0x80000002 = HKEY_LOCAL_MACHINE
        - and:
          - match: create process
          - string: /reg(|.exe) add /i
```

Potentially we could also add the `create process` and `reg add` block to the `set registry value`. Would you mind taking another look?

Otherwise, I'm ok with merging as is. Unless another reviewer objects, of course.

re-fox:
No additional comments here.  Appreciate all the work @mr-tz @0x534a

0x534a:
> Potentially we could also add the `create process` and `reg add` block to the `set registry value`. Would you mind taking another look?

I think this is a great idea that will lead to cleaner rules which follow the DRY principle. I catched up on this and provided the commit 13a4bf1 which implements your suggestion. We could also think about adding the `sc.exe` command to the `create-service.yml` rule in the future.

mr-tz (approved):

mr-tz:
Thanks a lot!

Issue #199, opened 2020-12-11
Additional shlwapi API's for get/set/open/create registry key operations 


re-fox:
## Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your rule idea isn't already filed: [search](https://github.com/fireeye/capa-rules/issues?q=is%3Aissue+is%3Aopen+)

## Summary

The rules under `host-interaction/registry` could benefit from adding the APIs from shlwapi for registry operations.

A large list of them can be found here-> https://docs.microsoft.com/en-us/windows/win32/sysinfo/registry-functions

One of the rules `host-interaction/registry/delete/` already has `SHDeleteKey` and `SHDeleteValue` in it.  So this one could most likely remain as-is.

Some quick examples would be:

For Open Key:
https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shregopenuskeya

For Query Key:
https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shqueryvalueexa

Get/Set Value:
https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shsetvaluea
https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shreggetvaluea

## Features
```
SHCopyKey
SHDeleteEmptyKey
SHEnumKeyEx
SHEnumValue
SHGetValue
SHQueryInfoKey
SHQueryValueEx
SHRegCloseUSKey
SHRegCreateUSKey
SHRegDeleteEmptyUSKey
SHRegDeleteUSValue
SHRegDuplicateHKey
SHRegEnumUSKey
SHRegEnumUSValue
SHRegGetBoolUSValue
SHRegGetIntW
SHRegGetPath
SHRegGetUSValue
SHRegOpenUSKey
SHRegQueryInfoUSKey
SHRegQueryUSValue
SHRegSetPath
SHRegSetUSValue
SHRegWriteUSValue
SHSetValue
```

### Namespace
`host-interaction/registry`

### References
https://docs.microsoft.com/en-us/windows/win32/sysinfo/registry-functions
https://docs.microsoft.com/en-us/windows/win32/shell/shlwapi-reg
https://www.geoffchappell.com/studies/windows/shell/shlwapi/api/index.htm

mike-hunhoff:
@re-fox with #210 I added all of these except the following:

- `SHCopyKey`
[...]

PR #210, opened 2020-12-30
new rules and rule updates inspired by B5F85C26D7AA5A1FB4AF5821B6B5AB9B

mike-hunhoff:


host-interaction/registry/query/query-registry-entry.yml
```diff
@@ -1,19 +0,0 @@
...
-      - BFB9B5391A13D0AFD787E87AB90F14F5:0x13145B5A
-      - Practical Malware Analysis Lab 03-02.dll_:0x100047AD
-  features:
-    - and:
-      - or:
-        - match: create registry key
-        - match: open registry key
-      - or:
-        - api: advapi32.RegQueryValue
-        - api: advapi32.RegQueryValueEx
```
mike-hunhoff, line 19: appeared to be iteration of `query-registry-value.yml`; removed to eliminate duplication

williballenthin:
image
classic @mike-hunhoff to drop a huge improvement PR like this!

host-interaction/file-system/create/create-file.yml
```diff
@@ -0,0 +1,15 @@
+rule:
+  meta:
+    name: create file
```
williballenthin, line 3: love the idea of factoring out this logic.

host-interaction/registry/create/set-registry-value.yml
```diff
@@ -2,17 +2,29 @@ rule:
...
         - api: advapi32.RegSetValue
         - api: advapi32.RegSetValueEx
+        - api: advapi32.RegSetKeyValue
+        - api: RtlWriteRegistryValue
+        - api: ZwSetValueKey
+        - api: SHSetValue
+        - api: SHRegSetPath
+        - api: SHRegSetValue
+        - api: SHRegSetUSValue
+        - api: SHRegWriteUSValue
```
williballenthin, line 30: wow these are becoming quite thorough

williballenthin (approved):
i think this is an outstanding PR. makes many existing rules much more thorough. 

interesting kernel-driver callback rules, too. should be interesting to hunt with these.

host-interaction/registry/open/open-registry-key.yml
```diff
@@ -3,14 +3,29 @@ rule:
...
       - Operating System::Registry::Open Registry Key [C0036.003]
     examples:
       - Practical Malware Analysis Lab 03-02.dll_:0x10004706
       - 493167E85E45363D09495D0841C30648:0x404D60
+      - B5F85C26D7AA5A1FB4AF5821B6B5AB9B:0x4045F2
   features:
     - or:
       - api: advapi32.RegOpenKey
       - api: advapi32.RegOpenKeyEx
+      - api: advapi32.RegCreateKey
```
mr-tz, line 17: overlap with `create registry key`?

host-interaction/registry/query/query-registry-key.yml
```diff
@@ -3,13 +3,25 @@ rule:
...
     att&ck:
       - Discovery::Query Registry [T1012]
     mbc:
       - Operating System::Registry::Query Registry Key [C0036.005]
     examples:
       - 493167E85E45363D09495D0841C30648:0x404930
+      - B5F85C26D7AA5A1FB4AF5821B6B5AB9B:0x402608
   features:
     - or:
+      - api: advapi32.RegEnumKey
```
mr-tz, line 16: I think enum and query provide different functionality. Is it worth splitting these up?

host-interaction/thread/create/create-thread.yml
```diff
@@ -2,18 +2,24 @@ rule:
...
   features:
     - or:
-      - api: kernel32.CreateThread
-      - api: _beginthread
-      - api: _beginthreadex
-      - basic block:
-        - and:
-          - api: kernel32.GetCurrentProcess
-          - api: kernel32.CreateRemoteThread
+      - or:
```
mr-tz, line 16: can remove one `or`

mr-tz (changes requested):
Nice job, a couple of comments open to discuss!

host-interaction/registry/create/set-registry-value.yml
re-fox: Thanks for adding these in.  Once these are added in #199 can probably be closed!  :+1:

host-interaction/file-system/open/open-file.yml
```diff
@@ -0,0 +1,18 @@
+rule:
+  meta:
+    name: open file
```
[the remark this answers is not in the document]
mike-hunhoff: yeah I struggled with the best way to go about adding create/open file rules; adding these as a library rule is a great idea. maybe we can combine to a single library rule named `create or open file`?

I have seen a few rules check for one or more of these APIs and think it would be great to capture all of the possible APIs in one or two rules to help tighten up the checks.

host-interaction/registry/open/open-registry-key.yml
mike-hunhoff: `RegCreateKey` can, per MSDN:

> Creates the specified registry key. If the key already exists in the registry, the function opens it.

Because of its dual purpose I included a check for `RegCreateKey` in both `open registry key` and `create registry key`. I had a similar struggle placing dual purpose APIs like this when creating the `create file` and `open file` rules.

Maybe there needs to be a bigger discussion on how dual purpose APIs should be expressed in capa rules e.g. including `RegCreateKey` in both `create registry key` and `open registry key` versus creating a separate rule named something like `create or open registry key` that only includes the dual purpose APIs like `RegCreateKey`. The latter IMO is more concise.

host-interaction/registry/query/query-registry-key.yml
mike-hunhoff: I see both enum and query as providing information about a given key, in the case of enum the name of one subkey each time it is called, so IMO it's not worth creating a separate rule just to split these up. But if you think they should still be split I'd be happy to update the PR.

host-interaction/file-system/start-kernel-mode-file-system-filter.yml
```diff
@@ -0,0 +1,13 @@
+rule:
+  meta:
+    name: start kernel-mode file system filter
```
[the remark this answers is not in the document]
mike-hunhoff: see comment `register kernel-mode file system filter`.

host-interaction/file-system/open/open-file.yml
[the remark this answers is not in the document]
mr-tz: > maybe we can combine to a single library rule named `create or open file`?

I like this. Please then replace `CreateFile` etc. with the new library rule or open an issue and I can do it later.

host-interaction/registry/open/open-registry-key.yml
mr-tz: Good points. All our current rules use `create` and `open` combined (or should), so combining into a `create or open registry key` rule makes sense in this case.

host-interaction/registry/query/query-registry-key.yml
mr-tz: A few descriptions from MSDN. Maybe we change to `query or enumerate registry key`? 😄 

- Enumerates the subkeys of the specified open registry key. The function retrieves the name of one subkey each time it is called.
- The ZwQueryKey routine provides information about the class of a registry key, and the number and sizes of its subkeys.
- Retrieves information about the specified registry key.

host-interaction/thread/create/create-thread.yml
mike-hunhoff: good catch!

host-interaction/registry/query/query-registry-key.yml
mike-hunhoff: renamed to `query or enumerate registry key`.

mr-tz (approved):
great, looking forward to seeing these in action!

PR #214, opened 2021-01-05
lint modified rules thoroughly

mr-tz:
Gets modified files and lints them thoroughly. All rules are also linted regularly before that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,15 +24,28 @@ jobs:
       with:
         repository: fireeye/capa-testfiles
         path: tests/data
-    - name: Set up Python 3.8
+    - name: Set up Python 2.7
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
-    # We don't need vivisect, so we can install capa using Python3
+        python-version: 2.7
     - name: Install capa
       run: pip install -e .
-    - name: Run rule linter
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -24,15 +24,27 @@ jobs:
...
+      uses: jitterbit/get-changed-files@v1
+    - name: Get modified rule names
+      run: |
+        for changed_file in ${{ steps.files.outputs.added_modified }}; do
+          tags+="\"$(grep 'name:' "${changed_file}" | sed 's/^.*: //')\" "
+        done
+        echo "$tags"
+        echo "tags=$tags" >> $GITHUB_ENV
+    - name: Run rule linter thoroughly
+      run: python scripts/lint.py -t ${{ env.tags }} rules/
```
williballenthin, line 44: lets use the --thorough flag here

.github/workflows/tests.yml
```diff
@@ -24,15 +24,27 @@ jobs:
...
+    # TODO removeme?
+#    - name: Run rule linter
+#      run: python scripts/lint.py rules/
```
williballenthin, line 47: lets keep these around in non-thorough mode - only takes a few seconds to run i think

williballenthin (changes requested):
good idea!

PR #215, opened 2021-01-05
update ATT&CK mappings

williballenthin:
thanks to R.E. for lots of (internal) mapping work that led to these clarifications.


williballenthin:
@dzbeck wouldn't mind having your thoughts on these changes if you notice anything seems weird/wrong.

c2/file-transfer/write-and-execute-a-file.yml
```diff
@@ -4,8 +4,6 @@ rule:
     namespace: c2/file-transfer
     author: moritz.raabe@fireeye.com
     scope: function
-    att&ck:
-      - Command and Control::Ingress Tool Transfer [T1105]
```
williballenthin, line 8: we have this under C2, but are we sure that this rule encodes command and control?

collection/credit-card/parse-credit-card-information.yml
```diff
@@ -4,8 +4,6 @@ rule:
     namespace: collection/credit-card
     author: "@_re_fox"
     scope: function
-    att&ck:
-      - Credential Access::Data from Local System [T1005]
```
williballenthin, line 8: CC data may not necessarily be from local system, nor  be credential data.

nursery/read-and-send-data-from-client-to-server.yml
```diff
@@ -4,8 +4,6 @@ rule:
     namespace: c2/file-transfer
     author: william.ballenthin@fireeye.com
     scope: function
-    att&ck:
-      - Exfiltration::Exfiltration Over C2 Channel [T1041]
```
williballenthin, line 8: are we sure this is C2?

nursery/receive-and-write-data-from-server-to-client.yml
```diff
@@ -4,8 +4,6 @@ rule:
     namespace: c2/file-transfer
     author: william.ballenthin@fireeye.com
     scope: function
-    att&ck:
-      - Command and Control::Ingress Tool Transfer [T1105]
```
williballenthin, line 8: are we sure this is C2?

c2/file-transfer/write-and-execute-a-file.yml
dzbeck: I don't think it's possible to know it's C2... I think something similar can be said for many rules/mappings, but I think there is value in suggesting such objectives/behaviors as possibilities (might be useful to associate confidence values with mappings). Could replace the current att&ck and mbc mapping with mbc micro behaviors (or add for context), which are literal/certain -- File System::Write File [C0052] and Process::Create Process [C0017].

mr-tz (approved):
Thanks, great changes!

PR #216, opened 2021-01-06
document example syntax

### LINKED ISSUE #205: Document example syntax format hash:fva
It's `md5:function offset`, so from the match above `0796f1c1ea0a142fc1eb7109a44c86cb:0x40222F`. I'll open an issue to update the documentation. Thanks!

If you include sample references, please upload the files to https://github.com/fireeye/capa-testfiles so we can test the matches.

_Originally posted by @mr-tz in https://github.com/fireeye/capa-rules/issues/174#issuecomment-748510524_

mr-tz:
closes #205

### DIFF
--- a/doc/format.md
+++ b/doc/format.md
@@ -131,6 +131,7 @@ like the ATT&CK list.
   - `examples` is a *required* list of references to samples that the rule should match.
 The linter verifies that each rule correctly fires on each sample referenced in a rule's `examples` list.
 These example files are stored in the [github.com/fireeye/capa-testfiles](https://github.com/fireeye/capa-testfiles) repository.
+`function` and `basic block` scope rules must contain offsets to the respective match locations using the format `<sample name>:<function or basic block offset>`.
 
   - `references` lists related information found in a book, article, blog post, etc.
 


doc/format.md
```diff
@@ -131,6 +131,8 @@ like the ATT&CK list.
   - `examples` is a *required* list of references to samples that the rule should match.
 The linter verifies that each rule correctly fires on each sample referenced in a rule's `examples` list.
 These example files are stored in the [github.com/fireeye/capa-testfiles](https://github.com/fireeye/capa-testfiles) repository.
+Example references use file names without extensions.
```
williballenthin, line 134: i dont think this is right: the linter hashes all the files it finds in the testdata directory and uses this to pick the right file. it may also use the literal filename (see the PMA examples). i don't think it strips the extension anywhere.

williballenthin (changes requested):
see inline comment

doc/format.md
williballenthin: https://github.com/fireeye/capa/blob/master/scripts/lint.py#L480-L484

williballenthin: image

mr-tz: Oh, good catch!

mr-tz:
Thanks, Willi!

PR #218, opened 2021-01-07
Create compiled-with-perl2exe.yml

re-fox:
Sample uploaded in https://github.com/fireeye/capa-testfiles/pull/54

This rule looks for a sample that is compiled with perl2exe, focusing on the function that loads up the `P2XDLL` and loading the `RunPerl` function.

### DIFF
--- a/compiler/perl2exe/compiled-with-perl2exe.yml
+++ b/compiler/perl2exe/compiled-with-perl2exe.yml
@@ -1,0 +1,17 @@
+rule:
+  meta:
+    name: compiled with perl2exe
+    namespace: compiler/perl2exe
+    author: "@_re_fox"
+    scope: function
+    examples:
+      - 873275ce8bf88ef66e9fa0c74b5c2a1e:0x4011C9
+  features:
+    - and:
+      - api: LoadLibrary
+      - api: FreeLibrary
+      - string: /^p2x[a-z0-9]{1,10}\.dll/i 
+      - basic block:
+        - and:
+          - api: GetProcAddress
+          - string: RunPerl


mr-tz (approved):
looks good, thank you!

compiler/perl2exe/compiled-with-perl2exe.yml
```diff
@@ -0,0 +1,17 @@
+rule:
+  meta:
+    name: compiled with perl2exe
+    namespace: compiler/perl2exe
+    author: "@_re_fox"
+    scope: function
```
williballenthin, line 6: seems that this should be a file-scope rule?

re-fox: Hi @williballenthin I checked the rule against 2 samples 

`3828d1da2e9774117e309fa92d1cb5c40e5d0728c4c29058b7785066a7729f8f`
`58e71c8e2a2712a6b1526134a8e656ca3f0d520ccd4041c2ffbd28d8030f120a`

In both samples the rule logic appeared within the context of a single function.  I'll copy and paste the decompiled output from Ghidra (for readability).  The function from both samples is roughly similar.

```c
int __cdecl FUN_004011c9(int param_1,int param_2,undefined4 param_3)

{
  int *piVar1;
  HMODULE hModule;
  char *pcVar2;
  int iVar3;
  char local_108 [260];
  
  piVar1 = FUN_00401000(&param_1,&param_2,local_108);
  if (piVar1 == (int *)0x0) {
    pcVar2 = "p2x5163.dll";
  }
  else {
    pcVar2 = local_108;
  }
  hModule = LoadLibraryA(pcVar2);
  if (hModule == (HMODULE)0x0) {
    printf("ERROR: LoadLibrary p2x5163.dll failed\n");
    iVar3 = 1;
  }
  else {
    _DAT_00404160 = GetProcAddress(hModule,"RunPerl");
    if (_DAT_00404160 == (FARPROC)0x0) {
      printf("ERROR: GetProcAddress p2x5163.dll failed\n");
      iVar3 = 2;
    }
    else {
      iVar3 = (*_DAT_00404160)(param_1,param_2,param_3);
      FreeLibrary(hModule);
      if (piVar1 != (int *)0x0) {
        remove(local_108);
        pcVar2 = FUN_00401bf1();
        _rmdir(pcVar2);
      }
    }
  }
  return iVar3;
}
```

With that said, If you think the rule is specific enough to increase the scope to file-level, I'd be happy to make adjustments.

mr-tz: `function` scope makes sense to me in this case

Issue #88, opened 2020-08-06
rule: thorough detection of injection


Ana06:
@williballenthin
> ![image](https://ghe.eng.fireeye.com/storage/user/969/files/811aad00-b377-11e9-99ca-c9577ae68959)
> 
> 
> http://struppigel.blogspot.com/2017/07/process-injection-info-graphic.html

@mr-tz
> https://www.blackhat.com/us-19/briefings/schedule/#process-injection-techniques---gotta-catch-them-all-16010

@mike-hunhoff
> More techniques described here: https://www.endgame.com/blog/technical-blog/ten-process-injection-techniques-technical-survey-common-and-trending-process.

PR #225, opened 2021-01-16
Improve injection detection and some refactoring

0x534a:
I suggest to improve the detection of code injection techniques with this PR. It fixes issue #88 partly by contributing new rules for certain injection techniques and by refactoring existing ones.

The suggested changes in this PR are mainly based on https://www.elastic.co/de/blog/ten-process-injection-techniques-technical-survey-common-and-trending-process and include:

- New rule for DLL injection (ATT&CK ID T1055.001)
- New rule for PE injection (ATT&CK ID T1055.002)
- New rules for MBC micro-behavior (thread suspend and resume)
- Refactoring of existing injection detection rules

The test-files repository will be updated accordingly with the added example files of the suggested rules afterwards.

### DIFF
--- a/host-interaction/process/inject/allocate-rwx-memory.yml
+++ b/host-interaction/process/inject/allocate-rwx-memory.yml
@@ -4,8 +4,6 @@ rule:
     namespace: host-interaction/process/inject
     author: moritz.raabe@fireeye.com
     scope: basic block
-    att&ck:
-      - Defense Evasion::Process Injection [T1055]
     mbc:
       - Memory::Allocate Memory [C0007]
     examples:
@@ -14,13 +12,5 @@ rule:
       - 563653399B82CD443F120ECEFF836EA3678D4CF11D9B351BB737573C2D856299:0x140001ABA
   features:
     - and:
-      - or:
-        - api: kernel32.VirtualAlloc
-        - api: kernel32.VirtualAllocEx
-        - api: kernel32.VirtualAllocExNuma
-        - api: kernel32.VirtualProtect
-        - api: kernel32.VirtualProtectEx
-        - api: ntdll.NtAllocateVirtualMemory
-        - api: ZwAllocateVirtualMemory
-      - or:
-        - number: 0x40 = PAGE_EXECUTE_READWRITE
+      - match: allocate memory
+      - number: 0x40 = PAGE_EXECUTE_READWRITE
--- a/host-interaction/process/inject/use-process-replacement.yml
+++ b/host-interaction/process/inject/use-process-replacement.yml
@@ -8,11 +8,11 @@ rule:
       - Defense Evasion::Process Injection::Process Hollowing [T1055.012]
     examples:
       - Practical Malware Analysis Lab 12-02.exe_:0x4010EA
+    references:
+      - http://www.autosectools.com/process-hollowing.pdf
+      - https://www.andreafortuna.org/2017/10/09/understanding-process-hollowing/
   features:
     - and:
       - match: create process suspended
-      - and:
-        - match: write process memory
-        - or:
-          - api: kernel32.ResumeThread
-          - api: ntdll.NtResumeThread
+      - match: write process memory
+      - match: resume a thread
--- a/host-interaction/process/inject/use-process-replacement.yml
+++ b/host-interaction/process/inject/use-process-replacement.yml
@@ -15,4 +15,4 @@ rule:
     - and:
       - match: create process suspended
       - match: write process memory
-      - match: resume a thread
+      - match: resume thread
--- a/host-interaction/thread/create/create-thread.yml
+++ b/host-interaction/thread/create/create-thread.yml
@@ -19,6 +19,9 @@ rule:
       - api: PsCreateSystemThread
       - api: SHCreateThread
       - api: SHCreateThreadWithHandle
-      - and:
-        - api: kernel32.GetCurrentProcess
-        - api: kernel32.CreateRemoteThread
+      - api: kernel32.CreateRemoteThread
+      - api: ntdll.RtlCreateUserThread
+      - api: ntdll.NtCreateThread
+      - api: ntdll.NtCreateThreadEx
+      - api: ntdll.ZwCreateThread
+      - api: ntdll.ZwCreateThreadEx
--- a/host-interaction/process/inject/hijack-thread-execution.yml
+++ b/host-interaction/process/inject/hijack-thread-execution.yml
@@ -1,0 +1,23 @@
+rule:
+  meta:
+    name: hijack thread execution
+    namespace: host-interaction/process/inject
+    author: 0x534a@mailbox.org
+    scope: function
+    att&ck:
+      - Defense Evasion::Process Injection::Thread Execution Hijacking [T1055.003]
+    examples:
+      - 787cbc8a6d1bc58ea169e51e1ad029a637f22560660cc129ab8a099a745bd50e:0x4044C7
+  features:
+    - and:
+      - optional:
+        - or:
+          - match: open a thread
+          - match: create thread
+      - match: suspend a thread
+      - api: kernel32.GetThreadContext
+      - match: allocate RWX memory
+      - optional:
+        - match: write process memory
+      - api: kernel32.SetThreadContext
+      - match: resume a thread
--- a/host-interaction/process/inject/hijack-thread-execution.yml
+++ b/host-interaction/process/inject/hijack-thread-execution.yml
@@ -7,17 +7,17 @@ rule:
     att&ck:
       - Defense Evasion::Process Injection::Thread Execution Hijacking [T1055.003]
     examples:
-      - 787cbc8a6d1bc58ea169e51e1ad029a637f22560660cc129ab8a099a745bd50e:0x4044C7
+      - 77d87e9937546aebc1595039d730352b15fab32c72a76913f04262c6802d098f:0x401000
   features:
     - and:
       - optional:
         - or:
-          - match: open a thread
+          - match: open thread
           - match: create thread
-      - match: suspend a thread
+      - match: suspend thread
       - api: kernel32.GetThreadContext
       - match: allocate RWX memory
       - optional:
         - match: write process memory
       - api: kernel32.SetThreadContext
-      - match: resume a thread
+      - match: resume thread
--- a/host-interaction/process/inject/inject-a-dll.yml
+++ b/host-interaction/process/inject/inject-a-dll.yml
@@ -1,0 +1,32 @@
+rule:
+  meta:
+    name: inject a dll
+    namespace: host-interaction/process/inject
+    author: 0x534a@mailbox.org
+    scope: function
+    att&ck:
+      - Defense Evasion::Process Injection::Dynamic-link Library Injection [T1055.001]
+    examples:
+      - Practical Malware Analysis Lab 17-02.dll_:0x1000D10D
+    references:
+      - Practical Malware Analysis, p. 676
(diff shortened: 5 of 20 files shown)


0x534a:
Corresponding PR to add test files in the capa-testfiles repo is [#56](https://github.com/fireeye/capa-testfiles/pull/56)

host-interaction/thread/create/create-thread.yml
```diff
@@ -19,6 +19,12 @@ rule:
...
         - api: SHCreateThreadWithHandle
+        - api: kernel32.CreateRemoteThread
+        - api: ntdll.RtlCreateUserThread
+        - api: ntdll.NtCreateThread
+        - api: ntdll.NtCreateThreadEx
+        - api: ntdll.ZwCreateThread
+        - api: ntdll.ZwCreateThreadEx
         - and:
           - api: kernel32.GetCurrentProcess
           - api: kernel32.CreateRemoteThread
```
mr-tz, line 30: We capture all thread creations here now, so can remove this block?

host-interaction/process/inject/use-thread-execution-hijacking.yml
```diff
@@ -0,0 +1,25 @@
+rule:
+  meta:
+    name: use thread execution hijacking
```
mr-tz, line 3: propose: `hijack thread execution`

host-interaction/process/inject/use-thread-execution-hijacking.yml
```diff
@@ -0,0 +1,25 @@
...
+      - Defense Evasion::Process Injection::Thread Execution Hijacking [T1055.003]
+    examples:
+      - 787cbc8a6d1bc58ea169e51e1ad029a637f22560660cc129ab8a099a745bd50e:0x4044C7
+  features:
+    - and:
+      - optional:
+        - or:
+          - api: kernel32.OpenThread
+          - api: ntdll.NtOpenThread
+          - api: ntdll.ZwOpenThread
```
mr-tz, line 17: looks like another good (library) rule candidate `open thread`

host-interaction/process/inject/use-process-replacement.yml
```diff
@@ -8,11 +8,11 @@ rule:
...
   features:
     - and:
       - match: create process suspended
-      - and:
-        - match: write process memory
-        - or:
-          - api: kernel32.ResumeThread
-          - api: ntdll.NtResumeThread
+      - match: write process memory
+      - match: resume thread
```
mr-tz, line 18: awesome!

host-interaction/process/inject/use-pe-injection.yml
```diff
@@ -0,0 +1,29 @@
+rule:
+  meta:
+    name: use pe injection
```
mr-tz, line 3: propose: `inject PE`

host-interaction/process/inject/use-pe-injection.yml
```diff
@@ -0,0 +1,29 @@
...
+    references:
+      - https://www.elastic.co/de/blog/ten-process-injection-techniques-technical-survey-common-and-trending-process
+  features:
+    - and:
+      - characteristic: loop
+      - optional:
+        - or:
+          - api: kernel32.OpenProcess
+          - api: ntdll.NtOpenProcess
+          - api: ntdll.ZwOpenProcess
```
mr-tz, line 20: new (library) rule candidate: `open process`

host-interaction/process/inject/use-pe-injection.yml
```diff
@@ -0,0 +1,29 @@
...
+          - match: create process
+      - match: allocate RWX memory
+      - basic block:
+        - description: virtual address offset calculation
+        - and:
+          - mnemonic: and
+          - number: 0x0FFF
```
mr-tz, line 27: I'm not sure if this is a key part in every implementation of `pe injection`

host-interaction/process/inject/use-dll-injection.yml
```diff
@@ -0,0 +1,34 @@
+rule:
+  meta:
+    name: use dll injection
```
mr-tz, line 3: propose: `inject DLL`

host-interaction/process/inject/allocate-rwx-memory.yml
```diff
@@ -21,6 +21,8 @@ rule:
         - api: kernel32.VirtualProtect
```
mr-tz, line 21: should we remove the `att&ck` mapping as this general functionality is not necessarily process injection?

host-interaction/process/inject/allocate-rw-memory.yml
```diff
@@ -0,0 +1,26 @@
+rule:
+  meta:
+    name: allocate RW memory
+    namespace: host-interaction/process/inject
+    author: 0x534a@mailbox.org
+    scope: basic block
+    att&ck:
+      - Defense Evasion::Process Injection [T1055]
```
mr-tz, line 8: Propose to remove this as not specific enough to process injection?

host-interaction/process/inject/allocate-rw-memory.yml
```diff
@@ -0,0 +1,26 @@
...
+      - or:
+        - api: kernel32.VirtualAlloc
+        - api: kernel32.VirtualAllocEx
+        - api: kernel32.VirtualAllocExNuma
+        - api: kernel32.VirtualProtect
+        - api: kernel32.VirtualProtectEx
+        - api: ntdll.NtAllocateVirtualMemory
+        - api: ntdll.ZwAllocateVirtualMemory
+        - api: ntdll.NtMapViewOfSection
+        - api: ntdll.ZwMapViewOfSection
```
mr-tz, line 24: could move these to a library rule `allocate memory` and use a match in the RW and RWX rules

mr-tz (changes requested):
Nice additions, thank you! Added a couple of improvement ideas and comments.
I prefer that rule names start with a verb other than `use` although we set the example with `process replacement` ourselves ;)

If possible, please rebase to the current master, so the linter action can run successfully using the newly merged test files.

host-interaction/process/inject/use-pe-injection.yml
williballenthin: "this function may...."
"inject *a* PE"

0x534a:
Thanks for all your comments. I addressed most of them in the newly provided commits. 

The linter currently fails with the following error:
```
allocate memory 


  FAIL: lib rule not found in lib directory: Move the rule to the `lib` subdirectory of the rules path
```

The `allocate-memory.yml` is already placed in the `lib` directory. Can you help me with fixing this linter error?

host-interaction/process/inject/allocate-rw-memory.yml
0x534a: Fixed in aa94a33 but there is a linter error. See comment below.

host-interaction/process/inject/allocate-rw-memory.yml
0x534a: Fixed in aa94a33

host-interaction/process/inject/allocate-rwx-memory.yml
0x534a: Fixed in aa94a33

host-interaction/process/inject/use-dll-injection.yml
0x534a: Fixed in aa94a33. Also added _a_ as suggested by @williballenthin for the PE injection rule.

host-interaction/process/inject/use-pe-injection.yml
0x534a: Fixed in aa94a33

host-interaction/process/inject/use-pe-injection.yml
0x534a: Fixed in aa94a33

host-interaction/process/inject/use-thread-execution-hijacking.yml
0x534a: Fixed in aa94a33

host-interaction/process/inject/use-thread-execution-hijacking.yml
0x534a: Fixed in aa94a33

host-interaction/thread/create/create-thread.yml
0x534a: I just left it since I thought it has a deeper sense for a certain rule. It is removed in aa94a33.

host-interaction/process/inject/use-pe-injection.yml
0x534a: I am also unsure regarding this block of the rule. I simply followed https://www.elastic.co/de/blog/ten-process-injection-techniques-technical-survey-common-and-trending-process in this case in order to write the rule. 

I think, you are the more proficient reverse engineers than me and maybe have more experience regarding pe injection. Do have certain samples in mind that implement pe injection in another way?

I will also do some research on my own but this will take some time.

0x534a:
I noticed that the usage of indefinite articles in rule names is quite inconsistent in the current rule base. E.g. `create-process.yml` is not using an article but `create-a-process-with-modified-io-handles-and-window.yml` is using one. This can be addressed in the future to make the rule base more consistent and to not confuse rule contributors. ;D

mr-tz:
> 
> 
> Thanks for all your comments. I addressed most of them in the newly provided commits.
> 
> The linter currently fails with the following error:
> 
> ```
> allocate memory 
> 
> 
>   FAIL: lib rule not found in lib directory: Move the rule to the `lib` subdirectory of the rules path
> ```
> 
> The `allocate-memory.yml` is already placed in the `lib` directory. Can you help me with fixing this linter error?

running this locally, I see the following:
```
allocate RW memory
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example

allocate user process RWX memory
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example
```

I'll investigate the missing matches and look into fixing the check linter output.

host-interaction/process/inject/use-pe-injection.yml
mr-tz: I say we remove this part.

0x534a: I think, we could then remove the whole rule since there is no real difference to `inject-thread.yml`.

mr-tz: Ah, I see. Let's keep the rule as is then.

mr-tz:
Reran this after the capa fix and getting
```
hijack thread execution 
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example
```

The offset doesn't match the sample. Could you take another look, please?

0x534a:
Thanks for the update!

Unfortunately, the sample referenced by the rule was not suitable since the memory allocation happens in a separate function. Therefore, I added another sample in the test files repository which matches the rule. Sorry for the extra work.

I also reworked the usage of articles by removing them from the rules and resolved the merge conflict.

0x534a:
There was another problem with the example files of the rule `inject-thread.yml`. The example file `Practical Malware Analysis Lab 12-01.exe_:0x4010D0` had no detection since the rule only matched on the allocation of RWX memory but this sample only allocates RW memory. This is fixed in commit cf84ebe.

Thanks for your comments!

mr-tz (approved):
Awesome, thanks a lot @0x534a! Great additions and improvements.
Happy to see that the linter helps us even more now to keep the rules quality high.

0x534a:
That's great news! :)

Thanks for merging my PR and for all your effort in fixing the linter and reviewing my stuff, @mr-tz!

PR #227, opened 2021-01-21
update mappings for mbc (part 5)

dzbeck:
Added MBC mappings to lib and linking namespaces, as well as for rules in other namespaces that were added after those namespaces were originally mapped to MBC. Where possible, all but nursery namespace have been mapped.


williballenthin (approved):
lgtm

williballenthin:
awesome, thank you!

dzbeck:
😃 Thanks, Willi!

PR #229, opened 2021-01-27
capafmted rules

mr-tz:
In conjunction with https://github.com/fireeye/capa/pull/401


lib/validate-payment-card-number-using-luhn-algorithm-with-lookup-table.yml
```diff
@@ -49,15 +49,15 @@ rule:
           - and:
             - description: Conversion of chr to int (LEA REG,[REG+ -0x30])
             - mnemonic: lea
-            - offset: -0x30
+            - offset: !!int '0x-30'
```
mr-tz, line 52: TODO

PR #230, opened 2021-01-27
reformated using capafmt

mr-tz:
mostly affects:
- meta reorder
- whitespace
- indents

with these changes and https://github.com/fireeye/capa/pull/401 the linter should be happy


mr-tz:
currently depends on #398

williballenthin (approved):
outstanding!

PR #235, opened 2021-02-01
create patch-process-command-line

williballenthin:


nursery/patch-process-command-line
```diff
@@ -0,0 +1,30 @@
...
+    name: patch process command line
+    namespace: anti-analysis/anti-forensic
+    author: william.ballenthin@fireeye.com
+    scope: function
+    references:
+      - https://stackoverflow.com/q/24754844/87207
+      - https://blog.xpnsec.com/how-to-argue-like-cobalt-strike/
+    examples:
+      # - CostaRicto
+  features:
```
mr-tz, line 12:
```suggestion  features:
    - and:
      - basic block:
        # example:
        #   mov     rbx, gs:60h
        #   lea     r9, [rsp+4A0h+flOldProtect] ; lpflOldProtect
        #   mov     edx, 8          ; dwSize
        #   mov     rcx, [rbx+20h]
        #   add     rcx, 70h ; 'p'  ; lpAddress
        #   lea     r8d, [rdx-4]    ; flNewProtect
        #   call    cs:VirtualProtect
        #   test    eax, eax
        - and:
          - characteristic: gs access
          - offset/x64: 0x60 = PEB
          - offset/x64: 0x20 = PEB->ProcessParameters
          - offset/x64: 0x70 = PEB->ProcessParameters->CommandLine
          - api: VirtualProtect
      - count(api(VirtualProtect)): 2 or more

```

mr-tz (changes requested):
also needs renaming (add `.yml`)

re-fox:
Interesting technique.

The binary from the writeup is `d6a8bbe44fcf20c573b137dd0aef7b7862eb4bc845fc0157ed0a6de1197e6a21`.  I'll download that and load it up into a disassembler to validate the signature.

williballenthin:
@re-fox ah, good idea! i saw this technique in a sample that isn't public, and haven't yet found another case.

mr-tz:
I've cleaned up the git history and fixed the indent. Feel free to merge or we can wait for @re-fox's validation.

re-fox:
The version in the write-up uses `NtQueryInformationProcess` to get PEB and then performs a Read/WriteProcessMemory to modify CommandLine.  It looks like the sample @williballenthin analyzed is manually enumerating the PEB.  

The above rule is certainly worth a push, and we can come back to add the NtQueryInformationProcess -> ReadProcessMemory -> WriteProcessMemory technique mentioned.

PR #237, opened 2021-02-04
patch-process-command-line updates

re-fox:
@williballenthin's PR provided a link (https://blog.xpnsec.com/how-to-argue-like-cobalt-strike/) demonstrating how to override command line arguments.  

I was able to find 2 samples.  A 64b sample (`d6a8bbe44fcf20c573b137dd0aef7b7862eb4bc845fc0157ed0a6de1197e6a21`) that is essentially the same as the blog post, and a second 32b (`cb73bb7630febc2f49f6e812d7b50fb67cbafec65d57ae4711299114cf794f52`)  sample that is similar, but slightly different. 

The 32b sample has been uploaded to `capa-testfiles` https://github.com/fireeye/capa-testfiles/pull/58

`WriteProcessMemory` is called at the end of function, which is responsible for the patching.  However it's called in such a way that the only feature is a `call`.   
```
insn: 0x00401a86: mnemonic(call)
insn: 0x00401abd: mnemonic(call)
```
Unfortunately this does make the signature weaker, but I believe the use of `NtQueryInformationProcess` + `ReadProcessMemory` + offsets of CommandLine make the signature less prone to false positives.


mr-tz:
looks good, can you please fix the whitespace issues (https://github.com/fireeye/capa-rules/pull/237/checks?check_run_id=1829087879#step:7:49)?

re-fox:
Fixed the linter errors.  Will be working capafmt.py into my workflow.

mr-tz (approved):
lgtm

mr-tz:
@williballenthin anything to add here or can we merge?

PR #238, opened 2021-02-04
connect-tcp-socket: add ConnectEx via WSAID_CONNECTEX

williballenthin:
image
image

sorry, non-public sample


re-fox:
This is an example where having a `guid` feature could be useful.  (https://github.com/fireeye/capa-rules/pull/183#issuecomment-730618016)  

I'll search my local repo to see if I have a sharable sample.  I should have enough information from the screenshot to dig.

mr-tz (approved):
Neat! The formatting lint seems to work well 😉

PR #242, opened 2021-02-06
Update and rename get-os-version.yml to check-os-version.yml

johnk3r:
https://github.com/fireeye/capa-rules/issues/28


johnk3r:
Review.

Issue #28, opened 2020-07-09
rule idea: check OS version new/update


mike-hunhoff:
Add new rule or update existing to detect OS version checks via following `kernel32` APIs:

- `VerSetConditionMask`
- `VerifyVersionInfo`

johnk3r:
Hello @mike-hunhoff .

I will work on that rule.

johnk3r:
sample: 8ed9a60127aee45336102bf12059a850

mike-hunhoff:
Hi @johnk3r! Thank you for finding a sample. I will assign this task to you, thank you!

johnk3r:
- [ ]     VerSetConditionMask
- [ ]     VerifyVersionInfo

johnk3r:
@mike-hunhoff ,

Would it be interesting to leave in the same rule?

I ask why the API GetVersion (Ex) was deprecated, too, we will need to make adjustments to the rule that is in production.

https://github.com/fireeye/capa-rules/blob/master/host-interaction/os/version/get-os-version.yml

mike-hunhoff:
@johnk3r looking back at the existing rule [get-os-version.yml](https://github.com/fireeye/capa-rules/blob/master/host-interaction/os/version/get-os-version.yml) I feel it should instead be named `check-os-version.yml` to better fit the intention. With this change the rule you added in fdf3afe could be combined with the existing rule.

Does it sound reasonable to rename the existing rule [get-os-version.yml](https://github.com/fireeye/capa-rules/blob/master/host-interaction/os/version/get-os-version.yml) to `check-os-version.yml` and combine it with your rule committed in fdf3afe? I feel this is a better fit all around and improves the existing rule.

johnk3r:
Hello @mike-hunhoff ,

I have an open PR. Once it's completed, I'll update this rule. 

Grateful

johnk3r:
Hi,

I will work on that rule again.

PR #243, opened 2021-02-06
#28

johnk3r:
https://github.com/fireeye/capa-rules/issues/28


mr-tz (approved):
Good addition, thanks.
We can merge this after the v1.5.0 release.

PR #244, opened 2021-02-07
Create encrypt-data-using-vest.yml

re-fox:
Rule addresses ->  https://github.com/fireeye/capa-rules/issues/231 

The bytes in the rule are VEST sbox and tables.  All values were pulled from www.ecrypt.eu.org/stream/p2ciphers/vest/vest_p2source.zip and truncated at 256 bytes.  

Sample using VEST is uploaded in https://github.com/fireeye/capa-testfiles/pull/60


williballenthin (approved):
lgtm

PR #258, opened 2021-02-11
Update resolve-dns.yml

johnk3r:
https://github.com/fireeye/capa-rules/issues/239

### DIFF
--- a/host-interaction/network/dns/resolve/resolve-dns.yml
+++ b/host-interaction/network/dns/resolve/resolve-dns.yml
@@ -2,13 +2,15 @@ rule:
   meta:
     name: resolve DNS
     namespace: host-interaction/network/dns/resolve
-    author: william.ballenthin@fireeye.com
+    author:
+      - william.ballenthin@fireeye.com
+      - johnk3r
     scope: function
     mbc:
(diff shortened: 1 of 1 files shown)


host-interaction/network/dns/resolve/resolve-dns.yml
```diff
@@ -2,13 +2,15 @@ rule:
...
+      - william.ballenthin@fireeye.com
+      - johnk3r
     scope: function
     mbc:
       - Communication::DNS Communication::Resolve [C0011.001]
     examples:
-      - 82986A4302421ABA1DB9DB34A8BC2BA1:0x401530  # winternals networking lab
-      - Practical Malware Analysis Lab 01-01.dll_:0x10001010
+      - 17264e3126a97c319a6a0c61e6da951e:0x5FDC25D0
+      - bab17c3bc9d0231a5c154f8a5899fc4f:0x100089F0
```
mr-tz, line 13: Are these the files added in https://github.com/fireeye/capa-testfiles/pull/67? See the comments there on the run time of one sample. Also it would be best to use the same hash type for the examples here and the file names.

mr-tz:
Tested locally, thanks!

PR #262, opened 2021-02-12
adding inspect-load-icon-resource.yml

mike-hunhoff:
interesting anti-analysis technique most likely meant to evade sandbox/emulation.

from MS docs for `LoadIcon`: _To use one of the predefined icons, set the hInstance parameter to NULL and the lpIconName parameter to one of the following values.....If the function fails, the return value is NULL._

observed numerous calls to `LoadIcon` passing `NULL` for first argument and random identifier for second argument, with a check for `NULL` return following each call.

### DIFF
--- a/nursery/inspect-load-icon-resource.yml
+++ b/nursery/inspect-load-icon-resource.yml
@@ -1,0 +1,26 @@
+# generated using capa explorer for IDA Pro
+rule:
+  meta:
+    name: inspect load icon resource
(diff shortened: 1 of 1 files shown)


mr-tz:
Looks good, would love to see this run with https://github.com/fireeye/capa/issues/424

williballenthin (approved):
lgtm

interested to see the efficacy here - any chance for it to FP?

Issue #240, opened 2021-02-05
allocate user process RWX memory: doesn't match on referenced example: Fix the rule logic or provide a different example


williballenthin:
from `lint.py --thorough`

mike-hunhoff:
fixed with #265

PR #265, opened 2021-02-17
removing ntdll requirement from function features exported by both ntdll and ntoskrnl

mike-hunhoff:
removing the `ntdll` requirement from function features that are exported by both `ntdll` and `ntoskrnl` will improve hits for kernel drivers e.g. #240.

### DIFF
--- a/executable/resource/extract-resource-via-kernel32-functions.yml
+++ b/executable/resource/extract-resource-via-kernel32-functions.yml
@@ -15,7 +15,7 @@ rule:
         - or:
           - api: kernel32.LoadResource
           - api: kernel32.LockResource
-          - api: ntdll.LdrAccessResource
+          - api: LdrAccessResource
         - optional:
           - match: contain a resource (.rsrc) section
           - api: kernel32.GetModuleHandle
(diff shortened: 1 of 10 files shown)


williballenthin (approved):
good idea

should we leave a note somewhere that we discovered this? or... there are 10 different files, so maybe it doesn't make sense copy the same comment everywhere.

mr-tz:
Great! I think documenting this would be a very good idea. Either at https://github.com/fireeye/capa-rules/blob/master/doc/format.md#api or even better as a linter suggestion?

mike-hunhoff:
> Great! I think documenting this would be a very good idea. Either at https://github.com/fireeye/capa-rules/blob/master/doc/format.md#api or even better as a linter suggestion?

From my analysis of `ntdll.dll` and `ntoskrnl.exe` found on Windows 7 there is overlap of ~625 exported symbols. I think it would be really neat to include this as a linter suggestion; less the user needs to worry about when building out their rules.

I'll merge this PR and create a new issue to discuss further how to best handle export overlap in `ntdll.dll` and `ntoskrnl.exe`.

PR #273, opened 2021-02-18
Update compiled-with-borland-delphi.yml

re-fox:
Addresses https://github.com/fireeye/capa-rules/issues/272


williballenthin (approved):
thanks for noticing and fixing this!

PR #276, opened 2021-02-18
add compiled-from-visual-basic.yml

### LINKED ISSUE #274: visual basic exe
library: `msvbvm60.dll`
api: `ThunRTMain`
string: `VB5!6&*`

williballenthin:
closes #274

### DIFF
--- a/compiler/vb/compiled-from-visual-basic.yml
+++ b/compiler/vb/compiled-from-visual-basic.yml
@@ -5,7 +5,7 @@ rule:
     author: "@williballenthin"
     scope: file
     examples:
-      -  9bca6b99e7981208af4c7925b96fb9cf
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
rule logic looks good
is this a file limitation rule like `compiler/autoit` or `runtime/dotnet`?

mike-hunhoff (approved):
rule logic is good, looks like the linter caught a small formatting error. other than that lgtm

alongside @mr-tz's comment we may want to open a separate issue to include this as rule limitation following the merge.

williballenthin:
yeah, this should result in a file limitation message (though it doesnt right now).

i think the best fix would be to implement https://github.com/fireeye/capa/issues/390 first

PR #277, opened 2021-02-19
compiled-with-exe4j.yml

### LINKED ISSUE #261: EXE4J
https://www.ej-technologies.com/download/exe4j/files
filemd5=0a29569004ceb84e86e3a551b485dc1c

johnk3r:
https://github.com/fireeye/capa-rules/issues/261

Closes #261

### DIFF
--- a/compiler/exe4j/compiled-with-exe4j.yml
+++ b/compiler/exe4j/compiled-with-exe4j.yml
@@ -10,10 +10,17 @@ rule:
       - 6b25f1e754ef486bbb28a66d46bababe:0x404EDE
   features:
     - or:
-      - string: /exe4j_log/i
-      - string: /install4j_log/i
-      - string: /exe4j_java_home/i
-      - string: /install4j/i
-      - string: /exe4j.isinstall4j/i
+      - string: exe4j_log
+      - string: install4j_log
+      - string: exe4j_java_home
+      - string: install4j
+      - string: exe4j.isinstall4j
       - string: /com/exe4j/runtime/exe4jcontroller/i
       - string: /com/exe4j/runtime/winlauncher/i
+      - string: EXE4J_LOG
+      - string: INSTALL4J_LOG
+      - string: EXE4J_JAVA_HOME
+      - string: INSTALL4J
+      - string: EXE4J.ISINSTALL4J
+      - string: /COM/EXE4J/RUNTIME/EXE4JCONTROLLER/i
+      - string: /COM/EXE4J/RUNTIME/WINLAUNCHER/i
--- a/compiler/exe4j/compiled-with-exe4j.yml
+++ b/compiler/exe4j/compiled-with-exe4j.yml
@@ -1,0 +1,19 @@
+rule:
+  meta:
+    name: compiled with exe4j
+    namespace: compiler/exe4j
+    author: johnk3r
+    scope: function
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information::Compile After Delivery [T1027.004]
+    examples:
+      - 6b25f1e754ef486bbb28a66d46bababe:0x404EDE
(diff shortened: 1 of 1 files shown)


compiler/exe4j/compiled-with-exe4j.yml
```diff
@@ -0,0 +1,19 @@
...
+      - 6b25f1e754ef486bbb28a66d46bababe:0x404EDE
+  features:
+    - or:
+      - string: /exe4j_log/i
+      - string: /install4j_log/i
+      - string: /exe4j_java_home/i
+      - string: /install4j/i
+      - string: /exe4j.isinstall4j/i
+      - string: /com/exe4j/runtime/exe4jcontroller/i
+      - string: /com/exe4j/runtime/winlauncher/i
```
williballenthin, line 19: i *think* we can use string literals (not regex) for these. did this not work for you?

we should prefer string literals because they will match faster.

compiler/exe4j/compiled-with-exe4j.yml
```diff
@@ -0,0 +1,19 @@
+rule:
+  meta:
+    name: compiled with exe4j
+    namespace: compiler/exe4j
+    author: johnk3r
+    scope: file
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information::Compile After Delivery [T1027.004]
```
williballenthin, line 8: lets remove this because exe4j is not specifically for obfuscation

williballenthin (changes requested):
come minor changes suggested

compiler/exe4j/compiled-with-exe4j.yml
johnk3r: I didn't know that (match faster), thanks for reporting. So, with regex the result was more accurate. For this sample, the first session is with regex, the second with literal string. What is your opinion? 

regex

compiler/exe4j/compiled-with-exe4j.yml
johnk3r: Ok. I took as reference other rules in the namespace compiler ...

compiler/exe4j/compiled-with-exe4j.yml
mr-tz: If they're always upper or lower case, changing accordingly (adding both variants) would work. If there's more combination, e.g. `Exe4j`, `EXE4j`, ... I'd prefer the regex.

johnk3r: Ok. I added the words in uppercase and regex just where it was really needed.

compiler/exe4j/compiled-with-exe4j.yml
```diff
@@ -0,0 +1,26 @@
...
+      - string: exe4j.isinstall4j
+      - string: /com/exe4j/runtime/exe4jcontroller/i
+      - string: /com/exe4j/runtime/winlauncher/i
+      - string: EXE4J_LOG
+      - string: INSTALL4J_LOG
+      - string: EXE4J_JAVA_HOME
+      - string: INSTALL4J
+      - string: EXE4J.ISINSTALL4J
+      - string: /COM/EXE4J/RUNTIME/EXE4JCONTROLLER/i
+      - string: /COM/EXE4J/RUNTIME/WINLAUNCHER/i
```
mr-tz, line 26: These are covered by the two regexes above. The `i` suffix indicates case-insensitive matching. Please remove the duplicates and then we can merge. Thanks for your great work here!

williballenthin:
```suggestion
```

williballenthin (approved):
🚀

williballenthin:
thank you @johnk3r !

PR #287, opened 2021-03-04
Create offline-registry-manipulation.yml

### LINKED ISSUE #211: offline registry manipulation with offreg.dll
originally suggested here: https://twitter.com/0gtweet/status/1344363283063701504?s=20

documentation:
- https://docs.microsoft.com/en-us/windows/win32/devnotes/offline-registry-library-portal
- https://docs.microsoft.com/en-us/windows/win32/devnotes/offline-registry-library-functions

johnk3r:
https://github.com/fireeye/capa-rules/issues/211

resolve #211


host-interaction/registry/offline-registry-manipulation.yml
```diff
@@ -0,0 +1,43 @@
...
+    authOR: johnk3r
+    scope: file
+    att&ck:
+      - Discovery::Query Registry [T1012]
+    mbc:
+      - Operating System::Registry::Query Registry Value [C0036.006]
+      - Operating System::Registry::Create Registry Key [C0036.004]
+      - Operating System::Registry::Open Registry Key [C0036.003]
+    examples:
+      - 3edc06a4784d1038c999b984d558b60d:0xEE20
```
williballenthin, line 14: is this the loaded virtual address?

williballenthin:
#211 suggests that we create a rule to match on executables that *use* the offline registry routines, rather than implement them. the rule you've proposed here seems to match on `offreg.dll`, instead.

can you please change this rule to match on functions that call routines provided by `offreg.dll`?

williballenthin (changes requested):
see discussion

johnk3r:
The rule sample was the only malicious file I found with these routines. And the creation of the rule was based on the "finds" in that file.

I will dedicate myself to find a suitable sample and soon update this rule.

recvfrom:
@johnk3r here's one: https://www.virustotal.com/gui/file/9a292192453b1746b23e26560d04bde422b44e48c1186e93ffb518fc8210225e/details

https://cisco.box.com/s/woy5cce2szd42gpf2yt67x5vhosgvlog (password: infected)

host-interaction/registry/offline-registry-manipulation.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: offline registry manipulation
+    namespace: host-interaction/registry
+    authOR: johnk3r
+    scope: function
+    att&ck:
+      - Discovery::Query Registry [T1012]
+    mbc:
+      - Operating System::Registry::Query Registry Value [C0036.006]
+      - Operating System::Registry::Create Registry Key [C0036.004]
+      - Operating System::Registry::Open Registry Key [C0036.003]
```
mr-tz, line 12: I think we should split the apis and respective techniques create, query, open, etc. up into separate rules. Would you mind doing that? It's fine if we don't have samples for all and add those to the nursery.

johnk3r: Good idea. Can I use the same sample if possible?

johnk3r: Hello. Please evaluate the submitted rules. 

How can I do when a sample of a rule takes too long to process? As the pipeline does not accept samples that take a long time to process, should I create this rule as a nursery?

williballenthin: yeah, putting the rule in the nursery is a fine choice.

host-interaction/registry/offline-set-registry.yml
```diff
@@ -0,0 +1,16 @@
+rule:
+  meta:
+    name: offline set registry
```
mr-tz, line 3:
```suggestion
    name: set registry value via offline registry library

```

mr-tz (changes requested):
thanks for splitting these up!

johnk3r:
Add nursery rule.

mr-tz:
Hey @johnk3r, is this ready for another review? Looks like there's only minor formatting issues (line endings).

johnk3r:
> Hey @johnk3r, is this ready for another review? Looks like there's only minor formatting issues (line endings).

Sorry man, I forgot to fix this.

mr-tz:
tested locally and only got the following warning

```
$ python2 ../scripts/lint.py . -t offline --thorough

    (nursery) delete registry key via offline registry library
      WARN: missing examples: Add meta.examples so that the rule can be tested and verified
```

mr-tz:
Thank you!

PR #288, opened 2021-03-04
Update Luhn's algorithm to catch TinyPOS implementation

re-fox:
Updated to trigger on a variant of the Luhn checksum used in a TinyPOS sample of `3679ae3462d5005a1967f278b1c40f39`. 

Rather than using an `add` or `shl` for multiplying by 2, this version explicitly uses a `imul`.  In the final step `mod 10`, the authors use a `div`, rather than an `idiv/cdq` or an optimized `shr/imul` combination. 

I checked against the `capa-testfiles` repo and the rule doesn't introduce any false negatives.  This version of the rule is less restrictive (doesn't have a `not: characteristic: nzxor`) so there may be a slight risk of false positives, but I haven't seen any yet.

### DIFF
--- a/lib/validate-payment-card-number-using-luhn-algorithm-with-no-lookup-table.yml
+++ b/lib/validate-payment-card-number-using-luhn-algorithm-with-no-lookup-table.yml
@@ -10,19 +10,19 @@ rule:
       - 6fcc13563aad936c7d0f3165351cb453:0x4026C0
   features:
     - and:
-      - not:
-        - characteristic: nzxor
       - characteristic: loop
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
lgtm
another motivator for #424

Ana06 (approved):
cool! 👍  I have added the `dont merge` label, so that we can release capa v1.6.0 with the [v1.6.0 label I have just created in capa-rules](https://github.com/fireeye/capa-rules/releases/tag/v1.6.0). But we can merge it afterwards. 😉

PR #298, opened 2021-03-18
ensure yml files use LF for EOL

mr-tz:
references: https://stackoverflow.com/a/10855862/10548020, https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings


mr-tz:
let me try this again :(

PR #305, opened 2021-03-22
enforce string formatting

mike-hunhoff:
includes changes resulting from string format enforcement.

see https://github.com/fireeye/capa/pull/468 for more information.


doc/format.md
```diff
@@ -317,6 +317,7 @@ This could also be an obfuscated string, for example a stack string.
 
 The parameter is a string describing the string.
 This can be the verbatim value, or a regex matching the string.
+Verbatim values containing special characters must be surrounded by double quotes and each special character must be escaped.
```
williballenthin, line 320: i think we need more detail here. what is a special character? lets enumerate those now before we forget.
```suggestion
Verbatim values containing special characters must be surrounded by double quotes and each special character must be escaped.
A special character is one of:
  - a backslash, which should be represented as `string: "\\"`
  - a newline or other non-space whitespace (e.g. tab, CR, LF, etc), which should be represented like `string: "\n"`
  - a double quote, which should be represented as `string: "\""`
  - a leading slash (not for a regex), which should be represented as `string: "/"`
```

doc/format.md
```diff
@@ -328,11 +329,12 @@ Examples:
...
 ```
 - string: Firefox 64.0
+- string: "Hostname:\t\t\t%s\nIP adress:\t\t\t%s\nOS version:\t\t\t%s\n"
 - string: This program cannot be run in DOS mode.
   description: MS-DOS stub message
-- string: '{3E5FC7F9-9A51-4367-9063-A120244FBEC7}'
+- string: "{3E5FC7F9-9A51-4367-9063-A120244FBEC7}"
   description: CLSID_CMSTPLUA
-- string: '/SELECT.*FROM.*WHERE/'
+- string: /SELECT.*FROM.*WHERE/
```
williballenthin, line 337: is this a regex? if its verbatim, needs double quotes.

williballenthin: oh, yeah, looks like a regex, sorry.

doc/format.md
williballenthin: what about `:` and `#`?

williballenthin (changes requested):
its close, but would like a bit more documentation to specify what a "special character" is and examples of how to escape them.

nursery/debug-build.yml
```diff
@@ -7,4 +7,4 @@ rule:
   features:
     - or:
       - string: Assertion failed!
```
mr-tz, line 9: shouldn't strings with white space be quoted as well?

doc/format.md
mr-tz: what about spaces, see comment below?

mike-hunhoff: I think the current list is comprehensive now that we enforce all strings must start/end with double quotes.

nursery/debug-build.yml
mike-hunhoff: covered not that we enforce all strings start/end with double quotes.

doc/format.md
```diff
@@ -316,7 +316,15 @@ This is probably a pointer to an ASCII or Unicode string.
 This could also be an obfuscated string, for example a stack string.
 
 The parameter is a string describing the string.
-This can be the verbatim value, or a regex matching the string.
+This can be the verbatim value or a regex matching the string.
+
+Verbatim values should be surrounded by double quotes and special characters should be escaped.
```
mr-tz, line 321:
```suggestion
Verbatim values must be surrounded by double quotes and special characters must be escaped.
```

doc/format.md
```diff
@@ -435,13 +444,21 @@ An ASCII or UTF-16 LE string present in the file.
 
 The parameter is a string describing the string.
 This can be the verbatim value, or a regex matching the string.
+
```
mr-tz, line 447: To not repeat ourselves I'd prefer to refer to the above section.

mr-tz (approved):
`- string: "awesome!"`

Ana06 (approved):
LGTM

PR #308, opened 2021-03-24
add load-code/pe/copy-pe-sections.yml

Ana06:
Add rules related to reflective dll injection.

references:
- https://0x00sec.org/t/reflective-dll-injection/3080
- https://www.ired.team/offensive-security/code-injection-process-injection/reflective-dll-injection
- Practical Malware Analysis, Chapter 19

I'll send a PR for uploading E4C33AC3638EEF68311F8AC0D72483C7 in a moment.


load-code/pe/copy-pe-sections.yml
```diff
@@ -0,0 +1,32 @@
...
+            - description: IMAGE_FIRST_SECTION(nt_header)
+            - offset: 0x14 = IMAGE_NT_HEADERS.FileHeader.SizeOfOptionalHeader
+            - offset: 0x18 = FileHeader.SizeOfOptionalHeader
+          - and:
+            - description: (DWORD)dll_raw + dos_header->e_lfanew + sizeof(IMAGE_NT_HEADERS) + sizeof(IMAGE_SECTION_HEADER) * i
+            - number: 0x28 = sizeof(IMAGE_SECTION_HEADER)
+            - or:
+              - offset/x32: 0xF8 = sizeof(IMAGE_NT_HEADERS32)
+              - offset/x64: 0x108 = sizeof(IMAGE_NT_HEADERS64)
+      - basic block:
```
Ana06, line 27: a `loop` scope would be useful here. I'll open an issue in capa for discussion.

load-code/pe/copy-pe-sections.yml
```diff
@@ -0,0 +1,32 @@
...
+    author: "@Ana06"
+    scope: function
+    references:
+      - https://0x00sec.org/t/reflective-dll-injection/3080
+      - https://www.ired.team/offensive-security/code-injection-process-injection/reflective-dll-injection
+    examples:
+      - E4C33AC3638EEF68311F8AC0D72483C7:0x401510
+  features:
+    - and:
+      - offset: 0x6 = IMAGE_NT_HEADERS.FileHeader.NumberOfSections
```
williballenthin, line 14: could we pull a bunch of this logic into "enumerate pe sections"?

load-code/pe/copy-pe-sections.yml
```diff
@@ -0,0 +1,32 @@
...
+            - number: 0x28 = sizeof(IMAGE_SECTION_HEADER)
+            - or:
+              - offset/x32: 0xF8 = sizeof(IMAGE_NT_HEADERS32)
+              - offset/x64: 0x108 = sizeof(IMAGE_NT_HEADERS64)
+      - basic block:
+        - and:
+          - offset: 0xC = IMAGE_SECTION_HEADER.VirtualAddress
+          - offset: 0x14 = IMAGE_SECTION_HEADER.PointerToRawData
+          - offset: 0x10 = IMAGE_SECTION_HEADER.SizeOfRawData
+          - mnemonic: call = call CopyMemory, memcpy or similar function
```
williballenthin, line 32: this seems pretty loose to me. maybe we wait until we have FLIRT that can match memcpy or something?

this would just as easily match `VirtualProtect` that you might expect to see nearby, but would not be related to copying memory.

load-code/pe/parse-pe-export-data.yml
```diff
@@ -0,0 +1,20 @@
+rule:
+  meta:
+    name: parse PE export data
```
williballenthin, line 3:
```suggestion
    name: parse PE exports
```

load-code/pe/perform-reflective-dll-injection.yml
```diff
@@ -0,0 +1,22 @@
...
+      - E4C33AC3638EEF68311F8AC0D72483C7:0x401510
+  features:
+    - and:
+      - match: copy PE sections
+      - match: rebuild import table
+      - basic block:
+        - and:
+          - offset: 0x28 = IMAGE_NT_HEADERS.OptionalHeader.AddressOfEntryPoint
+          - number: 0x1 = DLL_PROCESS_ATTACH
+          - characteristic: indirect call = call entry point
```
williballenthin, line 20: nice

load-code/pe/translate-characteristic-into-permission.yml
```diff
@@ -0,0 +1,22 @@
+rule:
+  meta:
+    name: translate characteristic into permission
```
williballenthin, line 3: what do you mean by "characteristic"?
```suggestion
    name: manipulate section memory permissions
```
maybe?

load-code/pe/translate-characteristic-into-permission.yml
```diff
@@ -0,0 +1,22 @@
...
+    namespace: load-code/pe
+    author: "@Ana06"
+    scope: function
+    examples:
+      - E4C33AC3638EEF68311F8AC0D72483C7:0x401480
+  features:
+    - and:
```
Ana06, line 10: Introducing a `match at least X` scope would be helpful here, as we could write something like that:
```
- match at least 3:
      - and:
	    - number: 0x40000000 = IMAGE_SCN_MEM_READ
	    - number: 0x2 = PAGE_READONLY
	  - and:
	    - number: 0x20000000 = IMAGE_SCN_MEM_EXECUTE
        - number: 0x10 = PAGE_EXECUTE
      - and:
        - number: 0x60000000 = IMAGE_SCN_MEM_READ | IMAGE_SCN_MEM_EXECUTE
        - number: 0x20 = PAGE_EXECUTE_READ
      - and:
        - number: 0xC0000000 = IMAGE_SCN_MEM_READ | IMAGE_SCN_MEM_WRITE
        - number: 0x4 = PAGE_READWRITE
      - and:
        - number: 0xE0000000 = IMAGE_SCN_MEM_READ | IMAGE_SCN_MEM_WRITE | IMAGE_SCN_MEM_EXECUTE
        - number: 0x40 = PAGE_EXECUTE_READWRITE
      - optional:
        - number: 0x1 = PAGE_NOACCESS

```

I know I can write an equivalent rule already but it is not a very beautiful one.

I can open an issue for this one as well.

load-code/pe/copy-pe-sections.yml
Ana06: are you proposing to rename this rule and remove `- mnemonic: call = call CopyMemory, memcpy or similar function`? 😕

williballenthin (changes requested):
most concerned about the call to `memcpy` or similar

load-code/pe/copy-pe-sections.yml
williballenthin: im proposing to factor out the section looping into its own rule, and have this rule `match: enumerate PE sections`

load-code/pe/translate-characteristic-into-permission.yml
williballenthin: does `3 or more` not work here?

williballenthin: i think it should

load-code/pe/parse-pe-export-data.yml
```diff
@@ -0,0 +1,20 @@
...
+  features:
+    - and:
+      - offset: 0x3C = IMAGE_DOS_HEADER.e_lfanew
+      - or:
+        - offset/x32: 0x78 = IMAGE_NT_HEADERS32.OptionalHeader.DataDirectory.VirtualAddress
+        - offset/x64: 0x88 = IMAGE_NT_HEADERS64.OptionalHeader.DataDirectory.VirtualAddress
+      - offset: 0x18 = IMAGE_EXPORT_DIRECTORY.NumberOfNames
+      - offset: 0x20 = IMAGE_EXPORT_DIRECTORY.AddressOfNames
+      - offset: 0x24 = IMAGE_EXPORT_DIRECTORY.AddressOfNameOrdinals
+      - offset: 0x1C = IMAGE_EXPORT_DIRECTORY.AddressOfFunctions
```
mr-tz, line 20: For large functions I could imagine there to be false-positives. Maybe grouping some into basic block subscopes could help.
I'm only guessing though since we don't have https://github.com/fireeye/capa/issues/424 yet.

load-code/pe/perform-reflective-dll-injection.yml
```diff
@@ -0,0 +1,22 @@
+rule:
+  meta:
+    name: perform reflective dll injection
```
mr-tz, line 3:
```suggestion
    name: inject DLL reflectively
```

load-code/pe/translate-characteristic-into-permission.yml
mr-tz: I don't think so as we only do that for `count(<feature>)...` but not across multiple features?

mr-tz (changes requested):
5 new rules, nice! 😉 With some minor tweaks, these should be great!

load-code/pe/parse-pe-export-data.yml
Ana06: This is most likely implemented with an `if-else`, so I don't think we can split it in basic blocks:

Screenshot 2021-03-25 at 19 14 53

load-code/pe/translate-characteristic-into-permission.yml
Ana06: The characteristics of an image section header which can take values such as `IMAGE_SCN_MEM_READ`, `IMAGE_SCN_MEM_EXECUTE` and `IMAGE_SCN_MEM_WRITE`: https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_section_header

This rule is not to detect manipulation section memory permissions, but the translation of the section permissions specified in the image section header to page protection constants used in functions such as `CreateFileMapping` and `VirtualProtect`.

I am not sure how to make this clearer in the rule. Suggestions?

load-code/pe/copy-pe-sections.yml
Ana06: Discussed offline to remove `mnemonic: call = call CopyMemory, memcpy or similar function` and call this rule `enumerate PE sections`

load-code/pe/copy-pe-sections.yml
Ana06: Discussed offline to remove `mnemonic: call = call CopyMemory, memcpy or similar function` and call this rule `enumerate PE sections`

load-code/pe/translate-characteristic-into-permission.yml
Ana06: It could be called `3 or more`, but still I think this is something we don't currently support. 🤔

williballenthin: i dont think so, i think `3 or more` is parsed like `optional` which supports arbitrary children terms:
https://github.com/fireeye/capa/blob/master/capa/rules.py#L329-L331

let me pull up the explorer and make a test case

williballenthin: image

williballenthin: i agree there's also a term like `count(mnem(mov)): 3 or more` but that's separate

williballenthin: btw that was soooo easy with the explorer @mike-hunhoff

load-code/pe/translate-characteristic-into-permission.yml
williballenthin: "characteristic" and "permission" are general terms. there's no indication from the rule name that we're talking about memory nor sections. im ok with keeping "translate" in there if we can clarify what types of characteristics and permissions we are talking about.

also, "translate" is an ok term, but its not great, since im not sure i've actually seen the phrase "translate memory characteristics" used before (though i agree it makes sense). "inspect section memory permissions" might also work.

load-code/pe/translate-characteristic-into-permission.yml
mr-tz: 🙈 wow, this needs to be documented better

Ana06: You are right @williballenthin, it works! 😮 although `optional` doesn't work as I expected. The following two seem to be equivalent:

```
- 2 or more: 
	- number: 0x1
	- number: 0x2
	- number: 0x3
	- optional:
		- number: 0x4
```

```
- 2 or more: 
	- number: 0x1
	- number: 0x2
	- number: 0x3
	- number: 0x4
```

but I had actually expected this behaviour:
```
- and:
	- 2 or more: 
		  - number: 0x1
		  - number: 0x2
		  - number: 0x3
	- optional:
		- number: 0x4
```

load-code/pe/translate-characteristic-into-permission.yml
Ana06: I have updated to `inspect section memory permissions` in 9db6392b3803b6f1be988c4003af597620bc5f59. IS that ok @williballenthin?

load-code/pe/translate-characteristic-into-permission.yml
mr-tz: `optional` always evaluates to true which has been a point of confusion before

load-code/pe/translate-characteristic-into-permission.yml
williballenthin: 👍🏼

PR #309, opened 2021-03-24
hash-data-using-fnv: removing arch requirement from number constants

mike-hunhoff:
observed `64-bit` sample referencing `32-bit` number constants:

```asm
mov     ecx, 811C9DC5h
lea     rdx, [rax+rbx]
jmp     short loc_2002EED9
movzx   eax, byte ptr [rbx]
imul    ecx, 1000193h
xor     ecx, eax
add     rbx, 1
```

removing arch requirement from number constants to up the matches.


mr-tz (approved):

mr-tz:
thanks, good find 😄

PR #311, opened 2021-03-25
updating rules + new rules

mike-hunhoff:


nursery/compare-security-identifiers.yml
```diff
@@ -0,0 +1,10 @@
+# generated using capa explorer for IDA Pro
```
williballenthin, line 1: ❤️

nursery/query-remote-server-for-availible-data.yml
```diff
@@ -0,0 +1,10 @@
+# generated using capa explorer for IDA Pro
+rule:
+  meta:
+    name: query remote server for availible data
```
williballenthin, line 4:
```suggestion
    name: query remote server for available data
```

+rename file

williballenthin: i wonder if there's a more human-readable name for this, like what is the context in which the server "has available data"?

williballenthin (changes requested):
typo is the only blocker

PR #333, opened 2021-04-22
New rules related to the amber packer

gormaniac:
There is an amber specific rule and a second that detects the shellcode loading function in the amber PE stub but should hopefully be a useful generic detection.
https://github.com/EgeBalci/amber


anti-analysis/packer/amber/packed-with-amber.yml
```diff
@@ -0,0 +1,15 @@
+rule:
+  meta:
+    name: packed with amber
+    namespace: anti-analysis/packer/amber
+    author: "john.gorman@fireeye.com"
+    scope: file
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
```
williballenthin, line 8:
```suggestion
      - Defense Evasion::Obfuscated Files or Information [T1027]
    mbc:
      - Anti-Static Analysis::Software Packing [F0001]
```

williballenthin:
```suggestion
      - Defense Evasion::Obfuscated Files or Information::Software Packing [T1027.002]
    mbc:
      - Anti-Static Analysis::Software Packing [F0001]      
```

load-code/pe/potential-shellcode-loader.yml
```diff
@@ -0,0 +1,19 @@
...
+      - https://github.com/EgeBalci/amber
+    examples:
+      - bb7922d368a9a9c8d981837b5ad988f1
+  features:
+    - or:
+      - and:
+        - api: kernel32.LoadLibraryA
+        - api: kernel32.GetProcAddress
+        - api: kernel32.GetModuleHandleA
+        - api: kernel32.VirtualAlloc
```
williballenthin, line 19: i suspect that this rule may match on too many functions, especially those unrelated to shellcode loading.

i'd propose that we prefer the existing rules in `load-code/pe` https://github.com/fireeye/capa-rules/tree/master/load-code/pe

williballenthin (changes requested):
i think the second rule should be removed from this PR, otherwise looking great, esp with minor tweaks to first rule metadata

gormaniac:
Removed the 2nd rule. I tried to fix the issues from the linter in `packed-with-amber.yml` but I am getting a different error this second time that I am not sure how to handle:
```
Error: Process completed with exit code 1.
```

williballenthin:
the (undocumented) trick is to look for "FAIL" in the linter output:

image

this is simply saying that the referenced test file is not present. thats expected, because we have another open PR for it. so we can ignore that for now. in other words, the linter is happy.

anti-analysis/packer/amber/packed-with-amber.yml
williballenthin, line 8:
```suggestion
      - Defense Evasion::Obfuscated Files or Information::Software Packing [T1027.002]
    mbc:
      - Anti-Static Analysis::Software Packing [F0001]
```

williballenthin: @gormaniac  you can hit "commit suggestion" above to easily merge this suggested change

gormaniac: @williballenthin  Oh see I left the `.002` sub-technique off because the language specifically calls out software packing as a `method of compressing or encrypting an executable`. While amber does "pack" the given PE into shellcode and subsequently a second PE, and refers to itself as a packer, I could not find evidence that it specifically encrypts or compresses any part of the original PE. Perhaps some encryption is done behind the scenes with SGN, but SGN documentation only refers to "encoding" when describing the transforms to supplied shellcode.

williballenthin: in the context of capa, we've used .002 to include reasonable forms of packing, such as amber that include encoding (but perhaps not compression nor encryption). this is also in line with contributions we've received from MITRE/ATT&CK authors.

so here, i think we should use .002, though i do appreciate your careful inspection of the ATT&CK technique definitions!

gormaniac: Good to know! Committed the change.

williballenthin (approved):

williballenthin:
thanks @gormaniac !

PR #338, opened 2021-04-24
Add rules from Egregor ransomware DLL

recvfrom:
Let me know if you'd like me to open tickets corresponding to these two new rules.

Test sample: https://github.com/fireeye/capa-testfiles/pull/86

Thanks!

### DIFF
--- a/host-interaction/process/terminate/terminate-process.yml
+++ b/host-interaction/process/terminate/terminate-process.yml
@@ -8,10 +8,12 @@ rule:
       - Process::Terminate Process [C0018]
     examples:
       - C91887D861D9BD4A5872249B641BC9F9:0x401A77
+      - 9B7CCAA2AE6A5B96E3110EBCBC4311F6:0x10010307
   features:
     - and:
       - optional:
         - api: kernel32.OpenProcess
+        - api: ntdll.NtOpenProcess
       - or:
         - api: kernel32.TerminateProcess
         - api: ntdll.NtTerminateProcess
--- a/host-interaction/network/domain/get-domain-information.yml
+++ b/host-interaction/network/domain/get-domain-information.yml
@@ -1,0 +1,13 @@
+rule:
+  meta:
+    name: get domain information
+    namespace: host-interaction/network/domain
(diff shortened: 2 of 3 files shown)


host-interaction/process/terminate/terminate-process.yml
```diff
@@ -8,10 +8,12 @@ rule:
       - Process::Terminate Process [C0018]
     examples:
       - C91887D861D9BD4A5872249B641BC9F9:0x401A77
+      - 9B7CCAA2AE6A5B96E3110EBCBC4311F6:0x10010307
   features:
     - and:
       - optional:
         - api: kernel32.OpenProcess
+        - api: ntdll.NtOpenProcess
```
williballenthin, line 16: nice catch

host-interaction/os/users/get-logon-sessions.yml
```diff
@@ -0,0 +1,16 @@
+rule:
+  meta:
+    name: get logon sessions
+    namespace: host-interaction/os/users
```
williballenthin, line 4: we have `host-interaction/session` already. perhaps that's a better existing namespace for this rule?

williballenthin (changes requested):
one namespace suggestion, otherwise looks great. thanks @recvfrom!

host-interaction/os/users/get-logon-sessions.yml
williballenthin: "when to use 'session' and when to use 'users'?" i don't know, but lets go for consistency right now :-)

williballenthin:
> Let me know if you'd like me to open tickets corresponding to these two new rules.

no need! we really don't want process to get in the way of progress. let us know if there's any way we can make contributions even easier

williballenthin (approved):
thank you!

host-interaction/process/terminate/terminate-process.yml
mr-tz: for reference, there's a `open process` library rule

mr-tz (commented):
thank you!

PR #374, opened 2021-05-07
ci: Update CHANGELOG automatically

Ana06:
Update new rules in capa repository's CHANGELOG with a GitHub Action automatically with every push (for example merge of PR).

You can see how the created commit looks like here: https://github.com/Ana06/capa/commit/7d7def34548434051af318197e736f815c7658c0

And the action running here: https://github.com/Ana06/capa-rules/runs/2528333012?check_suite_focus=true

Needs https://github.com/fireeye/capa/pull/549 to be merged to work.

Related: https://github.com/fireeye/capa/issues/457.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -42,10 +42,27 @@ jobs:
         # one which triggered the action)
         ref: master
         path: rules
+        fetch-depth: 100 # needed so that `git diff` finds `github.event.before`
     - name: Update rules number badge in README
       run: |
         num_rules=$(find rules -type f -name '*.yml' -not -path 'rules/.github/*' | wc -l)
         sed -i "s/rules-[0-9]*-blue\.svg/rules-$num_rules-blue.svg/" README.md
+    - name: Update number of new rules in CHANGELOG
+      run: |
+        new_rules=$(git -C rules diff -M --summary ${{ github.event.before }} | grep create | wc -l)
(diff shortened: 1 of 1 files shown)


.github/workflows/sync.yml
```diff
@@ -42,10 +42,27 @@ jobs:
...
+        old_rules=$(grep -m 1 "### New Rules.*" CHANGELOG.md | sed -r 's/.*\((.*)\)/\1/')
+        rules=$(($old_rules + $new_rules))
+        sed -ir "0,/New Rules.*/s//New Rules \($rules\)/" CHANGELOG.md
+    - name: Get modified files
+      id: files
+      uses: jitterbit/get-changed-files@v1
+    - name: Add new rules to CHANGELOG
+      run: |
+        for added_file in ${{ steps.files.outputs.added }}; do
+          author=$(grep 'author:' rules/$added_file | sed 's/^.*: //' | sed 's/"//g' )
```
Ana06, line 62: This is just adding whatever the rule author wrote in the author field. Till now we have always used the GitHub username for the CHANGELOG. I don't know an easy way to get the GitHub username (if someone know let me know!). But I think this is ok, we just show what people prefer that we show. We could document that GitHub user is preferred.

mr-tz (approved):
wow, that's some command line kung-fu, I say let's go with it!

PR #378, opened 2021-05-11
ci: support multiple author in sync GH

Ana06:
The sync GH action only works for a single author, producing unexpected results when using list of authors.

This change also prevent problems with new files that are not rules (`.yml`). (such as `.github/scripts/changelog_author.py`)

Closes https://github.com/fireeye/capa/issues/555

### DIFF
--- a/.github/scripts/changelog_author.py
+++ b/.github/scripts/changelog_author.py
@@ -1,0 +1,12 @@
+import yaml
+import sys
+
+rule_file = sys.argv[1]
+with open(rule_file, 'r') as stream:
+    rule_yaml = yaml.safe_load(stream)
+
+author_value = rule_yaml["rule"]["meta"]["author"]
+if isinstance(author_value, list): # list of authors
+    print(" ".join(author_value))
+else: # one author
+    print(author_value)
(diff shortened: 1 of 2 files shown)


.github/scripts/changelog_author.py
```diff
@@ -0,0 +1,12 @@
+import yaml
+import sys
+
+rule_file = sys.argv[1]
+with open(rule_file, 'r') as stream:
```
williballenthin, line 5: recommend: prefer to use `rb` with an explicit call to `f.read().decode("utf-8")`

no need to change here, just mentioning it

Ana06: can you please elaborate on why?

williballenthin: to be clear, this is a personal style thing.

basically, because the behavior here is platform-dependent:

image


while it probably doesn't matter in this case (CI is always going to be linux, and files never shared across platforms) doing things explicitly prevents subtle bugs for creeping in when files might be shared. therefore, i default to using explicit encodings in all cases.

Ana06: Thanks for the explanation @williballenthin!

PR #385, opened 2021-05-18
Update gather-leapftp-information.yml

re-fox:
Small patch to address https://github.com/fireeye/capa-rules/issues/382.


mr-tz:
thanks!

PR #386, opened 2021-05-19
remove single child statements

mr-tz:
see #382 and https://github.com/fireeye/capa/pull/563

PR #387, opened 2021-05-19
check-os-version: remove example that doesn't match anymore

williballenthin:
example is a library function that will be ignored due to FLIRT matching now.

https://github.com/fireeye/capa/issues/564


williballenthin:
closed with preference to fix in #386

PR #390, opened 2021-05-21
small fixes

ruppde:


compiler/go/compiled-with-go.yml
```diff
@@ -8,6 +8,6 @@ rule:
       - 49a34cfbeed733c24392c9217ef46bb6
   features:
     - or:
-      - string: /Go build ID:/
-      - string: /go.buildid/
-      - string: /Go buildinf:/
+      - string: "Go build ID:"
+      - string: "go.buildid"
+      - string: "Go buildinf:"
```
williballenthin, line 13: these need to be regexes so that they can match part of a string. typically the buildid string looks like `Go build ID: asdfafdsafdasfasd`

nursery/compiled-with-nim.yml
```diff
@@ -9,8 +9,8 @@ rule:
...
-      - string: /io.nim$/
-      - string: /fatal.nim$/
-      - string: /system.nim$/
-      - string: /alloc.nim$/
-      - string: /osalloc.nim$/
+      - string: /io\.nim$/
+      - string: /fatal\.nim$/
+      - string: /system\.nim$/
+      - string: /alloc\.nim$/
+      - string: /osalloc\.nim$/
```
williballenthin, line 16: all the rest of these fixes look great, thanks!

compiler/go/compiled-with-go.yml
williballenthin, line 13:
```suggestion
      - string: /^Go build ID:/
      - string: /go\.buildid/
      - string: /^Go buildinf:/
```

williballenthin (changes requested):
see inline suggestion

compiler/go/compiled-with-go.yml
williballenthin: these need to be regexes so that they can match part of a string. typically the buildid string looks like `Go build ID: asdfafdsafdasfasd`

mr-tz:
ping @ruppde 
Can I assist on this?

ruppde:
ok, it was new to me, that the regular strings in capa don't do a substring search. just came from doing lots of YARA and there it's default unless you use "fullword" ;)

maybe write that explicit in the documentation, just found it here for regex, but no that's the other way around for normal strings:
```
Regexes should be surrounded with / characters. By default, capa uses case-sensitive matching and assumes leading and trailing wildcards. 
```
from https://github.com/fireeye/capa-rules/blob/master/doc/format.md

mr-tz:
thanks, we're adding the clarification in #414

williballenthin (approved):

mr-tz:
thanks, @williballenthin

PR #395, opened 2021-05-24
Add packaged-as-an-inno-setup-installer

recvfrom:
Add `packaged-as-an-inno-setup-installer` which looks for simple strings associated with Inno Setup installers.  These strings appear to be present in installers produced using at least Inno Setup version 3 through the latest (Inno Setup version 6).

Related: https://github.com/fireeye/capa-testfiles/pull/95


williballenthin (approved):

williballenthin:
thank you!

PR #398, opened 2021-05-27
Add capture webcam image

### LINKED ISSUE #285: capture-webcam-image
## Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your rule idea isn't already filed: [search](https://github.com/fireeye/capa-rules/issues?q=is%3Aissue+is%3Aopen+)

## Summary

Let's have a rule that detects a system's webcam being used to capture a picture

## Examples

Ex: Calling `capCreateCaptureWindow` + calling `SendMessage` with `WM_CAP_DRIVER_CONNECT` + calling `SendMessage` with `WM_CAP_FILE_SAVEDIB` (and maybe optionally calling `capGetDriverDescription`)
 - Source for an implementation in meterpreter: https://github.com/rapid7/meterpreter/blob/d338f702ce8cb7f4e550f005ececaf5f3cadd2bc/source/extensions/espia/video.c#L19-L34
 - I think this might be an example of malware using this technique (it imports `capCreateCaptureWindowA` and `capGetDriverDescriptionA`) but haven't confirmed: https://www.virustotal.com/gui/file/a5deb825c9aa91d5f79a04dfa59d23340b34df3e598da1fbc6b7c2b39c9c074d/details
 - This technique is also highlighted in https://blog.talosintelligence.com/2021/02/obliquerat-new-campaign.html, but I'll need to ask which sample corresponds to the screenshot

## Features

See above

## Additional context

N/A

### Namespace
collection/webcam

### References

- https://docs.microsoft.com/en-us/windows/win32/multimedia/video-capture

### Other rule meta information

N/A

johnk3r:
https://github.com/fireeye/capa-rules/issues/285

@recvfrom , thanks for sharing this case. It was a new learning experience for me.

Study reference: https://docs.microsoft.com/en-us/windows/win32/multimedia/video-capture--a-minimal-approach

Closes #285


mr-tz (commented):
Should we add the constants `WM_CAP_DRIVER_CONNECT` and `WM_CAP_FILE_SAVEDIB` as suggested by @recvfrom? Otherwise, looking good!

mr-tz:
Please also address the linter fails:
```
capture webcam image 
  FAIL: file path doesn't match rule namespace: Move rule to appropriate directory or update the namespace
  FAIL: rule string quotes incorrect: add double quotes to "AVICAP32.dll"
```

johnk3r:
> 
> 
> Should we add the constants `WM_CAP_DRIVER_CONNECT` and `WM_CAP_FILE_SAVEDIB` as suggested by @recvfrom? Otherwise, looking good!

Ok.

johnk3r:
> 
> 
> Please also address the linter fails:
> 
> ```
> capture webcam image 
>   FAIL: file path doesn't match rule namespace: Move rule to appropriate directory or update the namespace
>   FAIL: rule string quotes incorrect: add double quotes to "AVICAP32.dll"
> ```

Should "capafmt.py" have detected the error in the string? He didn't signal any problems for me.

mr-tz:
No, this is currently only checked via the linter.

recvfrom:
@johnk3r `WM_CAP_DRIVER_CONNECT` is just a constant integer value, so to reference it in a capa rule you'd just figure out what its underlying value is and use it that way.  To walk you through this one, googling for `WM_CAP_DRIVER_CONNECT` leads to the following header file [1] which defines it:

```
#define WM_CAP_DRIVER_CONNECT           (WM_CAP_START+  10)
```

We'll need to figure out what `WM_CAP_START` is, and that's defined in that same file [2] as:

```
#define WM_CAP_START                    WM_USER
```
`WM_USER` is not defined in that file, so googling for the leads to [3]:

```
#define WM_USER                         0x0400
```

So `WM_CAP_DRIVER_CONNECT` is 0x0400 + 10 which equals 0x040a.  Then you can use it in capa like what's done in [4].  For example:

```
          - basic block:
            - and:
              - api: SendMessage
              - number: 0x40a = WM_CAP_DRIVER_CONNECT
```

[1] https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/um/Vfw.h#L3339
[2] https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/um/Vfw.h#L3312
[3] https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-user
[4] https://github.com/fireeye/capa-rules/blob/93ea28dd325c2db5e080272063c4dc12276c49cd/communication/socket/tcp/connect-tcp-socket.yml

recvfrom:
@johnk3r yep, those are them!

mr-tz:
@johnk3r can I provide further assistance here?

johnk3r:
> 
> 
> @johnk3r can I provide further assistance here?

Apologies for the delay. I wrote the rule again, I believe it's better now.

ReverseMalw_w10-2021-06-10-01-05-01

collection/webcam/capture-webcam-image.yml
```diff
@@ -0,0 +1,28 @@
...
+        - basic block:
+          - and:
+            - api: SendMessage
+            - number: 0x40B = WM_CAP_DRIVER_DISCONNECT
+        - optional:
+          - api: capGetDriverDescription
+      - basic block:
+        - and:
+          - api: SendMessage
+          - number: 0x419 = WM_CAP_FILE_SAVEDIB
```
mr-tz, line 28: is this block indented correctly? alone it seems to be not as reliable?

collection/webcam/capture-webcam-image.yml
```diff
@@ -0,0 +1,28 @@
...
+      - and:
+        - api: capCreateCaptureWindow
+        - basic block:
+          - and:
+            - api: SendMessage
+            - number: 0x40a = WM_CAP_DRIVER_CONNECT
+        - basic block:
+          - and:
+            - api: SendMessage
+            - number: 0x40B = WM_CAP_DRIVER_DISCONNECT
```
mr-tz, line 22: is this required or should it be optional?

mr-tz (commented):
thanks, the numbers looks good now.
two questions on the blocks inline

collection/webcam/capture-webcam-image.yml
johnk3r: I'm going to change it to optional.

collection/webcam/capture-webcam-image.yml
johnk3r: So, for the rule sample, if I leave this "block" in the main one, detection doesn't occur. In the screenshot I sent, you can see that this "block" is treated separately.

Note below:
  1º Only the constants and api were added. The conditioner used is an AND:
ReverseMalw_w10-2021-06-10-10-56-30

  2º Only the constants and api were added. The conditioner used is an OR:
ReverseMalw_w10-2021-06-10-10-55-10


I didn't understand why this is and how I can't use the CAPA plugin in IDA, I'm not getting a better logic for it. If you have any tips, I will be grateful!

mr-tz: My concern is if the
```yaml
      - basic block:
        - and:
          - api: SendMessage
          - number: 0x419 = WM_CAP_FILE_SAVEDIB
```
alone is enough to match on the capability

They are used in two separate functions so case 1 above doesn't match.

mr-tz:
bump 😃

johnk3r:
> 
> 
> bump 😃

Apologies for the delay.

I was reviewing this rule today and could not implement an adequate improvement following what was suggested in the ISSUE.

Analyzing the "WM_CAP_FILE_SAVEDIB" of the sample, I found only the use of "WM_CAP_GRAB_FRAME_NOSTOP" in the routine, but it wouldn't solve my problem if I add this contant.

The question is: For this case, is it possible to validate the two separate functions (AND)?  As in the example of the first image.

mr-tz:
let's keep both options WM_CAP_FILE_SAVEDIB and WM_CAP_DRIVER_CONNECT then, only left to do is to make the WM_CAP_DRIVER_DISCONNECT optional

johnk3r:
Hi, please check if the rule is compliant:

`  features:
    - or:
      - and:
        - api: capCreateCaptureWindow
        - basic block:
          - and:
            - api: SendMessage
            - number: 0x40a = WM_CAP_DRIVER_CONNECT
        - optional:
          - basic block:
            - and:
              - api: SendMessage
              - number: 0x40B = WM_CAP_DRIVER_DISCONNECT
      - basic block:
        - and:
          - api: SendMessage
          - number: 0x419 = WM_CAP_FILE_SAVEDIB`

Remembering that WM_CAP_FILE_SAVEDIB and WM_CAP_DRIVER_CONNECT are in different subroutines

mr-tz:
that looks good, thanks

mr-tz:
Thanks a lot!

PR #399, opened 2021-05-28
format: document function-name feature

williballenthin:
document `function-name` feature added in https://github.com/fireeye/capa/pull/568


mr-tz (approved):

Ana06 (approved):
👍

Issue #55, opened 2020-07-28
graduate nursery rules using existing examples


williballenthin:
A bunch of the nursery rules are there because they don't have an example yet, and I suspect that an example is already present in our existing set in capa-testfiles.  Run capa against all the testfiles and cross reference the matches with the rules found in the nursery. Graduate rules as appropriate.

williballenthin:
```
for F in tests/data/*.(exe_|dll_); do 
  echo "$F"; 
  time python -m capa.main -q -j "$F" > "$F".json; 
done

cat rules/nursery/*.yml | grep "name:" | cut -d ":" -f 2 | sed -e "s/^ //g" | while read -r "RULE"; do 
  echo "$RULE"; 
  for FILE in tests/data/*.json; do 
    if jq ".rules | keys" "$FILE" | grep -q "$RULE" ; then 
      echo "  $FILE"; 
    fi ; 
  done; 
done
```


```
access the Windows event log
  tests/data/39c05b15e9834ac93f206bc114d0a00c357c888db567ba8f5345da0529cbed41.dll_.json
  tests/data/5f66b82558ca92e54e77f216ef4c066c.exe_.json
  tests/data/6cc148363200798a12091b97a17181a1.exe_.json
  tests/data/b7841b9d5dc1f511a93cc7576672ec0c.dll_.json
  tests/data/mimikatz.exe_.json
acquire debug privileges
bypass UAC via ICMLuaUtil
  tests/data/08ac667c65d36d6542917655571e61c8.exe_.json
bypass UAC via scheduled task environment variable
connect network resource
  tests/data/55d77ab16377a8a314982f723fcc6fae.exe_.json
create shortcut via IShellLink
  tests/data/55d77ab16377a8a314982f723fcc6fae.exe_.json
  tests/data/6f99a2c8944cb02ff28c6f9ced59b161.exe_.json
debug build
decrypt data via SSPI
  tests/data/2b8bec5bcb1777eaa155d832f7afc797.exe_.json
empty the recycle bin
  tests/data/55d77ab16377a8a314982f723fcc6fae.exe_.json
encode data using Base64 via WinAPI
  tests/data/2b8bec5bcb1777eaa155d832f7afc797.exe_.json
  tests/data/5f66b82558ca92e54e77f216ef4c066c.exe_.json
  tests/data/bc577119d1a5b7da489e7b5817d3cc38.dll_.json
[...]

PR #413, opened 2021-06-08
2021 06 08 graduation

williballenthin:
10 rule graduations. a tweak to the MD5 rule logic. small reorganization of a couple other rules.

ref #55


mr-tz (approved):
fantastic!

PR #414, opened 2021-06-10
clarify verbatim string matching

mr-tz:
per #390

> ok, it was new to me, that the regular strings in capa don't do a substring search. just came from doing lots of YARA and there it's default unless you use "fullword" ;)
> 
> maybe write that explicit in the documentation, just found it here for regex, but no that's the other way around for normal strings:
> 
> ```
> Regexes should be surrounded with / characters. By default, capa uses case-sensitive matching and assumes leading and trailing wildcards. 
> ```
> 
> from https://github.com/fireeye/capa-rules/blob/master/doc/format.md

### DIFF
--- a/doc/format.md
+++ b/doc/format.md
@@ -326,9 +326,12 @@ A special character is one of:
   - a newline or other non-space whitespace (e.g. tab, CR, LF, etc), which should be represented like `string: "\n"`
   - a double quote, which should be represented as `string: "\""`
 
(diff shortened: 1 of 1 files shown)


williballenthin (approved):
thanks! this does clarify how the features match.

williballenthin:
@ruppde  thoughts?

ruppde:
> @ruppde thoughts?

looks good (but also means that I need to change capa2yara because that still does substring search. when I touch it next time ...)

PR #421, opened 2021-06-29
set save workspace env var to speed up testing

mr-tz:


.github/workflows/tests.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       # however, it still gets the modified files and we can continue
       # Error: The head commit for this pull_request event is not ahead of the base commit.
       continue-on-error: true
+    # save workspaces to speed up testing
+    - name: set CAPA_SAVE_WORKSPACE environment variable
```
mr-tz, line 45: could slightly improve if multiple rules per file are submitted at once

williballenthin (approved):

.github/workflows/tests.yml
```diff
@@ -41,6 +41,9 @@ jobs:
...
+      run: echo "CAPA_SAVE_WORKSPACE=True" >> $GITHUB_ENV
```
williballenthin, line 46: i think can maybe use a top level element to set the env var for all steps, like:

```yaml
env:
    CAPA_SAVE_WORKSPACE: "True"
```

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#env

williballenthin (approved):

.github/workflows/tests.yml
mr-tz: nice find, updated in 441550d

PR #429, opened 2021-07-19
adding gopacker capa

jaredscottwilson:
Added the capa for files packed with gopacker (https://github.com/petercunha/gopacker/).


anti-analysis/packer/gopacker/gopacker_capa.yml
```diff
@@ -0,0 +1,29 @@
+rule:
+  meta:
+    name: packed with GoPacker
```
williballenthin, line 3: filename should match the rule name, so probably `packed-with-gopacker.yml`

anti-analysis/packer/gopacker/gopacker_capa.yml
```diff
@@ -0,0 +1,29 @@
...
+    namespace: anti-analysis/packer/gopacker
+    author: jared.wilson@fireeye.com
+    description: The sample appears to be packed with GoPacker.
+    scope: file
+    att&ck:
+    - Defense Evasion::Obfuscated Files or Information [T1027.002]
```
williballenthin, line 9: formatting here is inconsistent, need another two spaces

anti-analysis/packer/gopacker/gopacker_capa.yml
```diff
@@ -0,0 +1,29 @@
...
+    mbc:
+      - Anti-Static Analysis::Software Packing
```
williballenthin, line 11: this should have an ID with it, i think

anti-analysis/packer/gopacker/gopacker_capa.yml
```diff
@@ -0,0 +1,29 @@
...
+      - ebfba63891471ced641b60e0403777da
+      - 03ae008aa48e52d32517a60e2ce1a1d2
+    references:
+      - https://github.com/petercunha/gopacker
+  features:
+    - and:
+        - string: "Anonymous"
+        - string: "<autogenerated>"
+        - string: "github.com/nirhaas/gopacker/"
+        - string: ".packed"
```
williballenthin, line 22: is this a section name? if so, use feature `section: .packed`

anti-analysis/packer/gopacker/gopacker_capa.yml
```diff
@@ -0,0 +1,29 @@
...
+        - string: "ZSTDCompression"    
+        - string: "gopacker"
+        - string: "LALALALA"
+        - description: LALALALA is a unique and default magic footer for this packing. This footer can be changed, however, by default it is not.
```
williballenthin, line 26: for the description to be associated with the string feature `LALALALA`, then need to make this not-a-list element
```suggestion
          description: LALALALA is a unique and default magic footer for this packing. This footer can be changed, however, by default it is not.
```

anti-analysis/packer/gopacker/gopacker_capa.yml
```diff
@@ -0,0 +1,29 @@
...
+        - string: /\$f64\.[a-zA-Z0-9]{16}/
```
williballenthin, line 27: regular expressions are a bit expensive to match (we have to scan *every* string for each regex). do you think this feature (and the associated `count: ...`) are strictly required to match this packer? if so - no problem! if not, then i'd recommend removing this feature as a performance tweak.

williballenthin (changes requested):
awesome! some fairly minor feedback and then we'll be happy to merge

anti-analysis/packer/gopacker/gopacker_capa.yml
jaredscottwilson: It's not, it's actually the default extension given to packed files. I'll add a Description.

anti-analysis/packer/gopacker/gopacker_capa.yml
jaredscottwilson: It's not, it's actually the default extension given to packed files. I'll add a Description.

anti-analysis/packer/gopacker/gopacker_capa.yml
jaredscottwilson: Great question. I would think it's not. I just was trying to get as specific as possible to avoid FPs. However, let me remove it and run some tests to see if we can move forward without.

anti-analysis/packer/gopacker/gopacker_capa.yml
williballenthin:
```suggestion
      - Anti-Static Analysis::Software Packing [F0001]
```

anti-analysis/packer/gopacker/packed-with-gopacker.yml
```diff
@@ -0,0 +1,28 @@
...
+      - string: "<autogenerated>"
+      - string: "github.com/nirhaas/gopacker/"
+      - string: ".packed"
+        description: This is the default file extension given to packed files.
+      - string: "ZSTDCompression"
+        description: This packer is set to use ZSTD.
+      - string: "gopacker"
+      - string: "LALALALA"
+        description: LALALALA is a unique and default magic footer for this packing. This footer can be changed, however, by default it is not.
+      - count(string(LALALALA)): 2
```
williballenthin, line 28:
```suggestion
      - optional:
        - string: "LALALALA"
          description: LALALALA is a unique and default magic footer for this packing. This footer can be changed, however, by default it is not.
```

if this is optional, let's mark it as such.

anti-analysis/packer/gopacker/packed-with-gopacker.yml
```diff
@@ -0,0 +1,28 @@
...
+      - https://github.com/petercunha/gopacker
+    examples:
+      - 4122acca2f9ea98fc3f3ad040688e4ce
+  features:
+    - and:
+      - string: "Anonymous"
+      - string: "<autogenerated>"
+      - string: "github.com/nirhaas/gopacker/"
+      - string: ".packed"
+        description: This is the default file extension given to packed files.
```
williballenthin, line 22: thanks for the in-line clarification here!

jaredscottwilson:
Sounds good. I think it's value added to include it, but from what I can tell, moving it to optional doesn't impact TP/FP rate.

williballenthin:
thank you!

PR #443, opened 2021-08-16
document os/arch/format features

### LINKED ISSUE #446: doc: change arch to bitness
e.g. "If the number is only relevant for a particular **architecture**, then you can use one of the architecture flavors: number/x32 or number/x64." in doc/format.md

williballenthin:
related to https://github.com/fireeye/capa/pull/723

closes #446


mr-tz:
nice!

PR #456, opened 2021-08-27
transmitpackets callback function via wsaioctl

jlepore-fe:


communication/socket/tcp/send/wsaioctl_transmitpackets_callbackfunc.yml
```diff
@@ -0,0 +1,28 @@
+rule:
+  meta:
+    name: obtains TransmitPackets callback function via WSAIoctl
```
williballenthin, line 3: the filename should match the rule name, lowercase and using dashes instead of spaces

williballenthin, line 3: the filename should match the rule name, lowercase and using dashes instead of spaces

communication/socket/tcp/send/wsaioctl_transmitpackets_callbackfunc.yml
```diff
@@ -0,0 +1,28 @@
+rule:
+  meta:
+    name: obtains TransmitPackets callback function via WSAIoctl  
+    description: The TransmitPackets function transmits in-memory data or file data over a connected socket. The TransmitPackets function uses the operating system cache manager to retrieve file data, locking memory for the minimum time required to transmit and resulting in efficient, high-performance transmission.
```
williballenthin, line 4: for long line, use YAML folded style:
```suggestion
    description: >-
      The TransmitPackets function transmits in-memory data or file data over a connected socket. 
      The TransmitPackets function uses the operating system cache manager to 
      retrieve file data, locking memory for the minimum time required to transmit and 
      resulting in efficient, high-performance transmission.
```

williballenthin: https://stackoverflow.com/a/3790497/87207

communication/socket/tcp/send/wsaioctl_transmitpackets_callbackfunc.yml
```diff
@@ -0,0 +1,28 @@
...
+    namespace: communication/socket/tcp/send
+    author: jonathan.lepore@mandiant.com
+    scope: function 
+    references: https://docs.microsoft.com/en-us/windows/win32/api/mswsock/nc-mswsock-lpfn_transmitpackets
+    examples:
+      - 138C71E94961FE9E31CEC5DFBA62A639:0x100FF55
```
williballenthin, line 10: please open a second PR against https://github.com/fireeye/capa-testfiles/ that adds this file

williballenthin (changes requested):
two formatting changes and then ready to merge!

mr-tz (commented):
I agree with Willi's suggestions. Otherwise, looks great!
You can test/lint the rule locally using `scripts/lint.py` as well.

PR #459, opened 2021-09-10
rule additions and updates

### LINKED ISSUE #418: check-http-status-code is a little loose
as noted in https://github.com/fireeye/capa/issues/641, communication/http/client/check-http-status-code is a bit loose. it will match with all of the following:

  - `cmp`
  - `0x13`
  - `0x200`

these numbers aren't very large, so they might reasonably seen in structure manipulation code, etc. we should consider tightening up this rule, at least with this combination of features.

mr-tz:
some additions and a fix for a linux FP
closes #418

### DIFF
--- a/communication/http/client/check-http-status-code.yml
+++ b/communication/http/client/check-http-status-code.yml
@@ -2,7 +2,7 @@ rule:
   meta:
     name: check HTTP status code
     namespace: communication/http/client
-    author: moritz.raabe@fireeye.com
+    author: "@mr-tz"
     scope: function
     mbc:
       - Communication::HTTP Communication::Read Header [C0002.014]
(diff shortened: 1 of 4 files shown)


communication/http/client/check-http-status-code.yml
```diff
@@ -2,14 +2,15 @@ rule:
   meta:
     name: check HTTP status code
     namespace: communication/http/client
-    author: moritz.raabe@fireeye.com
+    author: "@mr-tz"
```
williballenthin, line 5: 🤔

nursery/linked-against-cpp-http-library.yml
```diff
@@ -0,0 +1,12 @@
...
+  meta:
+    name: linked against CPP HTTP library
+    namespace: linking/static/httplib
+    author: "@mr-tz"
+    scope: file
+    references:
+      - https://github.com/yhirose/cpp-httplib
+  features:
+    - or:
+      - string: /cpp-httplib\//
```
williballenthin, line 11:
```suggestion
      - substring: "cpp-httplib/"
```

williballenthin (approved):

PR #463, opened 2021-09-15
Created rules for the ELF BEACON sample shared by Intezer

ramen0x3f:
Some rules leveraging the newly added ELF support in capa.

### DIFF
--- a/host-interaction/hardware/memory/manipulate-linux-process-memory.yml
+++ b/host-interaction/hardware/memory/manipulate-linux-process-memory.yml
@@ -1,0 +1,19 @@
+rule:
+  meta:
+    name: manipulate Linux process memory
+    namespace: host-interaction/hardware/memory
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - or:
+        - function:
+          - or:
+            - api: malloc
+            - api: memcpy
+            - api: memmove
+            - api: memset
+        - import: free
--- a/host-interaction/mutex/create-semaphore.yml
+++ b/host-interaction/mutex/create-semaphore.yml
@@ -1,0 +1,12 @@
+rule:
+  meta:
+    name: create semaphore
+    namespace: host-interaction/mutex
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - import: sem_init
--- a/host-interaction/mutex/lock-semaphore.yml
+++ b/host-interaction/mutex/lock-semaphore.yml
@@ -1,0 +1,14 @@
+rule:
+  meta:
+    name: lock semaphore
+    namespace: host-interaction/mutex
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - or:
+        - import: sem_timedwait
+        - import: sem_wait
--- a/host-interaction/network/create-network-socket.yml
+++ b/host-interaction/network/create-network-socket.yml
@@ -1,0 +1,17 @@
+rule:
+  meta:
+    name: create network socket
+    namespace: host-interaction/network
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - or:
+        - import: recv
+        - import: recvfrom
+        - import: setsockopt
+        - import: socket
+        - import: sendto
--- a/host-interaction/process/create/execute-command-through-syscall.yml
+++ b/host-interaction/process/create/execute-command-through-syscall.yml
@@ -1,0 +1,12 @@
+rule:
+  meta:
+    name: execute command through syscall
+    namespace: host-interaction/process/create
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - import: syscall
--- a/lib/delay-execution.yml
+++ b/lib/delay-execution.yml
@@ -1,9 +1,11 @@
 rule:
   meta:
     name: delay execution
-    author: michael.hunhoff@fireeye.com
+    author: 
+      - michael.hunhoff@fireeye.com
+      - alyssa.rahman@mandiant.com
     lib: true
-    scope: basic block
+    scope: file
     mbc:
       - Anti-Behavioral Analysis::Dynamic Analysis Evasion::Delayed Execution [B0003.003]
     references:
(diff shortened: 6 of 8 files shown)


host-interaction/hardware/memory/manipulate-linux-process-memory.yml
```diff
@@ -0,0 +1,19 @@
+rule:
+  meta:
+    name: manipulate Linux process memory
```
williballenthin, line 3: the APIs referenced here are not linux specific (they're part of the c standard library) so lets loosen that restriction

host-interaction/hardware/memory/manipulate-linux-process-memory.yml
```diff
@@ -0,0 +1,19 @@
...
+    - and:
+      - os: linux
+      - or:
+        - function:
+          - or:
+            - api: malloc
+            - api: memcpy
+            - api: memmove
+            - api: memset
+        - import: free
```
williballenthin, line 19: i think we should prefer API functions to imports, unless the feature makes more sense at the file level. when we use an API feature, capa can tell us exactly which function has the capability. when we use imports, we just get a general sense of the capabilities.

so, lets make this a function (or basic block) scoped rule and rely on API features

host-interaction/mutex/create-semaphore.yml
```diff
@@ -0,0 +1,12 @@
+rule:
+  meta:
+    name: create semaphore
```
williballenthin, line 3:
```suggestion
    name: create semaphore on Linux
```

host-interaction/mutex/create-semaphore.yml
```diff
@@ -0,0 +1,12 @@
...
+    namespace: host-interaction/mutex
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - import: sem_init
```
williballenthin, line 12:
```suggestion
      - api: sem_init
```

and change scope to bb or function

host-interaction/mutex/create-semaphore.yml
williballenthin: same comment on other semaphore rules

host-interaction/mutex/create-semaphore.yml
williballenthin: same comment on other semaphore rules

host-interaction/network/create-network-socket.yml
```diff
@@ -0,0 +1,17 @@
...
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - or:
+        - import: recv
+        - import: recvfrom
+        - import: setsockopt
+        - import: socket
+        - import: sendto
```
williballenthin, line 17: i think these routines are referenced in more specific ways by the rules found here: https://github.com/fireeye/capa-rules/tree/master/communication/socket

host-interaction/process/create/execute-command-through-syscall.yml
```diff
@@ -0,0 +1,12 @@
...
+    name: execute command through syscall
+    namespace: host-interaction/process/create
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - import: syscall
```
williballenthin, line 12: `syscall` is an API that directly invokes the system call provided as an argument (i think, as documented here: https://man7.org/linux/man-pages/man2/syscall.2.html). so, this rule needs more specificity to disambiguate execution of a command from the other system calls (like recv, write, etc.).

lib/delay-execution.yml
```diff
@@ -14,17 +16,23 @@ rule:
...
+              - api: kernel32.WaitForMultipleObjects
+              - api: kernel32.WaitForMultipleObjectsEx
+              - api: kernel32.RegisterWaitForSingleObject
+              - api: WaitOnAddress
+              - api: user32.MsgWaitForMultipleObjects
+              - api: user32.MsgWaitForMultipleObjectsEx
+              - api: NtDelayExecution
+              - api: KeWaitForSingleObject
+              - api: KeDelayExecutionThread
+      - import: signal
```
williballenthin, line 37: `signal` is used to configure the handling of many types of notifications, beyond sleeping (i think, at least as documented here: https://man7.org/linux/man-pages/man2/signal.2.html). so, i think this portion needs more logic to specify the specific type of signal that is being configured to effect a delay.

lib/delay-execution.yml
```diff
@@ -14,17 +16,23 @@ rule:
...
+              - api: kernel32.WaitForSingleObjectEx
+              - api: kernel32.WaitForMultipleObjects
+              - api: kernel32.WaitForMultipleObjectsEx
+              - api: kernel32.RegisterWaitForSingleObject
+              - api: WaitOnAddress
+              - api: user32.MsgWaitForMultipleObjects
+              - api: user32.MsgWaitForMultipleObjectsEx
+              - api: NtDelayExecution
+              - api: KeWaitForSingleObject
+              - api: KeDelayExecutionThread
```
williballenthin, line 36: these should be restricted with `os: windows` while the other block restricted with `os: linux`. we haven't quite gotten to all the rules to specify the os, so when we come across them during updates, we should add those conditions in to make clear which APIs apply to which OSes.

williballenthin (changes requested):
thank you @ramen0x3f! i'm excited that you're contributing these eight new rules that identify behaviors found in a real (interesting) malware sample.

as you can see from my inline comments, my biggest request is to prefer to use the "most specific scope" possible when identifying features, as this gives us the most information about where and how a behavior is referenced. with this in mind, please use `api` features at the BB or function scope rather than `import` features at the file scope, when possible. this will, e.g., let us associated capa matches with symfuncs or other function level systems.

feel free to ask any questions or for clarification on any of these points - happy to help get this stuff merged in!

host-interaction/hardware/memory/manipulate-linux-process-memory.yml
mr-tz: I wonder how useful this rule will be as one of these apis will be used in 99% of samples? I vote to remove this generic rule.

host-interaction/mutex/create-semaphore.yml
```diff
@@ -0,0 +1,12 @@
+rule:
+  meta:
+    name: create semaphore
+    namespace: host-interaction/mutex
+    author: alyssa.rahman@fireeye.com
+    scope: file
```
mr-tz, line 6:
```suggestion
    scope: function
```

mr-tz: see Willi's comment below

host-interaction/mutex/lock-semaphore.yml
```diff
@@ -0,0 +1,14 @@
...
+    author: alyssa.rahman@fireeye.com
+    scope: file
+    examples:
+      - 3db3e55b16a7b1b1afb970d5e77c5d98
+  features:
+    - and:
+      - os: linux
+      - or:
+        - import: sem_timedwait
+        - import: sem_wait
```
mr-tz, line 14: the linter will complain about a missing empty line at the end of the file, ensure to run `lint.py`/`lint.py --thorough` on all rule files

host-interaction/network/create-network-socket.yml
mr-tz: agreed, vote to remove this generic rule

lib/delay-execution.yml
```diff
@@ -14,17 +16,23 @@ rule:
...
-      - api: WaitOnAddress
-      - api: user32.MsgWaitForMultipleObjects
-      - api: user32.MsgWaitForMultipleObjectsEx
-      - api: NtDelayExecution
-      - api: KeWaitForSingleObject
-      - api: KeDelayExecutionThread
+      - function:
+        - and:
+          - basic block:
+            - or:
```
mr-tz, line 22: our current preference is to limit the scope via the file meta data, so I'd keep this as is.

mr-tz:
Thanks, Alyssa, I've taken your work, made changes, and opened a new PR at #654.

PR #467, opened 2021-09-30
Add FNV-specific hash checking rule

Still34:
## Summary

This PR adds a new rule that detects known FNV hashes that correspond to common import names (e.g., `GetProcAddress` FNV'd ->0xf8f45725. This technique was spotted in the wild and was adopted by a Chinese APT group.


nursery/resolve-function-by-fnv-hash.yml
```diff
@@ -0,0 +1,21 @@
+rule:
+  meta:
+    name: resolve function by FNV hash
```
mr-tz, line 3: can you please add a `description:` that this is `FNV-1a`
alternatively modify the name (and rename the file)

mr-tz (changes requested):
great, thanks, one small clarification request and then we can merge

nursery/resolve-function-by-fnv-hash.yml
Still34: Done. Hope that's enough!

mr-tz: almost, the file needs to be renamed to `resolve-function-by-fnv-1a-hash.yml` and I think reformatted, you can test it locally using `scripts/lint.py`

Still34: Oh, didn't notice there was an enforced linting rule. Re-saved as LF-only and whatnot. Should pass the linting now.

Still34: ...wait, forgot to add the renamed file.

mr-tz:
Awesome, thanks for the quick fixes and your contribution!

PR #488, opened 2021-10-28
Small fix for aPLib detection

cdong1012:
Remove `count(characteristic(calls from)): 2 or more` to detect aPLib compression used in BlackMatter v3.0.

Also the linting fails with B43FCA5283BFC7022553EFF663683834 cause capa throws UnsupportedFormatError, so I removed it from the rule. Lmk if you guys want me to add it back.

### DIFF
--- a/data-manipulation/compression/decompress-data-using-aplib.yml
+++ b/data-manipulation/compression/decompress-data-using-aplib.yml
@@ -12,7 +12,6 @@ rule:
       - https://ibsensoftware.com/files/aPLib-1.1.1.zip
     examples:
       - DAA13AE302FE8B618DDBF590537443EF:0x419F50
-      - B43FCA5283BFC7022553EFF663683834:0x424
       - 757139E76FAE876AE50DD2C3AC11D5D8:0x413074
   features:
     - and:
--- a/data-manipulation/compression/decompress-data-using-aplib.yml
+++ b/data-manipulation/compression/decompress-data-using-aplib.yml
@@ -12,18 +12,34 @@ rule:
       - https://ibsensoftware.com/files/aPLib-1.1.1.zip
     examples:
       - DAA13AE302FE8B618DDBF590537443EF:0x419F50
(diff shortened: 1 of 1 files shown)


mr-tz:
Thanks for the contribution!

The B43FCA5283BFC7022553EFF663683834 lint seems to pass for me, where are you getting the error?

Do you know if the expected calls are inlined here or if the algorithm works differently?

cdong1012:
Hmm it's not really inlined cause the the patterns from the rules are inside a function, and that function is definitely called more than 2 times. Not sure why the existing rule doesn't detect it unless I remove that.

williballenthin:
I'm not sure why this rule isn't matching either and `show-features` lists all the features that we expect to see in this function. So, I'd like to triage why matching isn't working before merging this PR.

image
image

williballenthin:
ah, nevermind, the rule specifies 2x "call from" (not 2x "call to" that i screenshotted above).

confirmed that there are no calls from function 0x413074 in 757139E76FAE876AE50DD2C3AC11D5D8. i will next find out if the expected function calls have been inlined.

williballenthin:
with calls vs inlined variants:

image

mr-tz:
nice investigation, can we pull some features from the called/inlined functions maybe?

williballenthin:
recommend that we make the calls from feature optional.

i think we can also add the following features:

  - shl 8
  - shr 1

image
image

image
image

these features are found in all 3 sample binaries that we have.

while some compilations use pusha/popa, the dll distributed with aplib 1.1.1 does not use this.

williballenthin:
added logic, seems to work

image

please review @mr-tz  @cdong1012 and then i'd be happy to merge

cdong1012:
Just pulled it down and tested it. The new logic looks good to me, and everything works nicely! Thank you for your work on this!

mr-tz (approved):
awesome, thanks! only misses an author ;)

Issue #470, opened 2021-10-18
add coverage for registry manipulation via WMI StdRegProv


mike-hunhoff:
> The StdRegProv class contains methods that manipulate system registry keys and values. StdRegProv is preinstalled in the WMI namespaces root\default and root\cimv2.

https://docs.microsoft.com/en-us/previous-versions/windows/desktop/regprov/stdregprov

mike-hunhoff:
based on this: `hxxps://microsoft.public.win32.programmer.wmi.narkive[.]com/e5KHEMoC/using-stdregprov-setdwordvalue-in-c`

`GetObject` must be used to get a referenced to the class, in this case `StdRegProv`, and `GetMethod` must be used to get a reference to the method e.g. `SetStringValue`.

so we could write a rule for `registry manipulation via StdRegProv` containing something like the following:

```yml
- and:
  - string: "StdRegProv"
  - or:
    - string: "SetStringValue"
    - string: "CreateKey"
    - string: "DeleteKey"
    - string: "SetDwordValue"
    - string: "DeleteValue"
```

PR #498, opened 2021-11-15
adding registry manipulation via StdRegProv

mike-hunhoff:
see #470 for more information. essentially, we hit on the strings passed to `GetObject`, e.g. `StdRegProv`, and `GetMethod`, e.g. `CreateKey`.

### DIFF
--- a/nursery/query-or-enumerate-registry-key-via-stdregprov.yml
+++ b/nursery/query-or-enumerate-registry-key-via-stdregprov.yml
@@ -9,5 +9,4 @@ rule:
   features:
     - and:
       - string: "StdRegProv"
-      - or:
-        - string: "EnumKey"
+      - string: "EnumKey"
--- a/nursery/query-or-enumerate-registry-key-via-stdregprov.yml
+++ b/nursery/query-or-enumerate-registry-key-via-stdregprov.yml
@@ -1,0 +1,13 @@
+# generated using capa explorer for IDA Pro
+rule:
+  meta:
+    name: query or enumerate registry key via StdRegProv
+    namespace: host-interaction/registry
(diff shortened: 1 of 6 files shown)


williballenthin (approved):
would love to see an example for these rules. would you (or shall i) find a candidate on VT perhaps?

williballenthin:
I can look for a sample on VT

williballenthin:
not a lot of samples out there that have all the capabilities in one:
image

williballenthin:
image
no matches in current set of samples. i give up for right now :-)

nursery/query-or-enumerate-registry-key-via-stdregprov.yml
```diff
@@ -0,0 +1,13 @@
...
+    name: query or enumerate registry key via StdRegProv
+    namespace: host-interaction/registry
+    author: michael.hunhoff@mandiant.com
+    scope: function
+    references: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/regprov/stdregprov#methods
+  features:
+    - and:
+      - string: "StdRegProv"
+      - or:
+        - string: "EnumKey"
```
mr-tz, line 13:
```suggestion
      - string: "EnumKey"
```

mr-tz (commented):
I agree that 1-2 example samples would be good.

mike-hunhoff:
thanks for checking @williballenthin! do we want to hold a merge until we can find one or two public samples?

mike-hunhoff:
@williballenthin @mr-tz bump

williballenthin:
i dont have any examples, so send it!

PR #502, opened 2021-11-30
Update reference-anti-vm-strings-targeting-vmware.yml

### LINKED ISSUE #495: rule idea: anti-vm check for VMWare via registry check
identified malware checking if registry key `HKCR\Applications\VMwareHostOpen.exe` exists.

we can use this to expand https://github.com/mandiant/capa-rules/blob/master/anti-analysis/anti-vm/vm-detection/reference-anti-vm-strings-targeting-vmware.yml

johnk3r:
Rule updated as recommended. I found a valid sample for the new adjustment.

closes #495


williballenthin (approved):
thank you!

PR #504, opened 2021-12-03
Add linked-against-wolfssl.yml and linked-against-wolfcrypt.yml

jtothej:
This PR adds two new rules that match on strings found in binaries statically linked with wolfSSL or wolfCrypt.
wolfCrypt is a backend library used by wolfSSL but it can also be used separately.

### DIFF
--- a/linking/static/wolfcrypt/linked-against-wolfcrypt.yml
+++ b/linking/static/wolfcrypt/linked-against-wolfcrypt.yml
@@ -1,0 +1,21 @@
+rule:
+  meta:
+    name: linked against wolfCrypt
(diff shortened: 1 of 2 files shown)


mr-tz (approved):
Looks good to me! Thanks.

williballenthin (approved):

williballenthin:
thank you!

williballenthin:
at first I wondered about using a more specific scope, e.g. `function`, and trying to identify specific routines. however, unless the routine indicates something about what the program is definitely doing (e.g. "encrypting with AES") then I agree its better to describe the linked library. so, no concerns here.

PR #513, opened 2021-12-12
Adding inject-shellcode-using-mapping-injection.yml

jtothej:
New rule that detects process injection using the Mapping Injection technique.

### DIFF
--- a/host-interaction/process/inject/inject-shellcode-using-mapping-injection.yml
+++ b/host-interaction/process/inject/inject-shellcode-using-mapping-injection.yml
@@ -1,0 +1,21 @@
+rule:
+  meta:
+    name: inject shellcode using Mapping Injection
+    namespace: host-interaction/process/inject
+    author: jakub.jozwiak@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Process Injection [T1055]
+    mbc:
+      - Defense Evasion::Process Injection [E1055]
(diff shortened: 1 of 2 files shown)


host-interaction/process/inject/inject-shellcode-using-mapping-injection.yml
```diff
@@ -0,0 +1,21 @@
+rule:
+  meta:
+    name: inject shellcode using Mapping Injection
```
williballenthin, line 3:
```suggestion
    name: inject shellcode using a file mapping object
```

imho, "Mapping Injection" isn't a a well known term and the technique seems to be a fairly straightforward application of file mapping objects, so let's just describe it as that

williballenthin (approved):
logic looks good, though some overlap with here:
https://github.com/mandiant/capa-rules/blob/26d3a7937f826169588d356f5af1a7bc1211e365/host-interaction/process/inject/inject-apc.yml

this rule is more succinct, though.

one recommendation around the rule name. let me know what you think.

host-interaction/process/inject/inject-shellcode-using-mapping-injection.yml
jtothej: Sure, that makes perfect sense. I submitted the renamed rule.

williballenthin:
thank you!

PR #514, opened 2021-12-12
Adding execute-shellcode-via-copyfile2.yml

jtothej:
New rule that detects shellcode execution via callback function in CopyFile2

### DIFF
--- a/load-code/shellcode/execute-shellcode-via-copyfile2.yml
+++ b/load-code/shellcode/execute-shellcode-via-copyfile2.yml
@@ -1,0 +1,28 @@
+rule:
+  meta:
+    name: execute shellcode via CopyFile2
+    namespace: load-code/shellcode
+    author: jakub.jozwiak@mandiant.com
+    scope: function
+    references:
+      - https://github.com/S4R1N/AlternativeShellcodeExec/blob/master/CopyFile2/CopyFile2.cpp
+    examples:
+      - c2bb17c12975ea61ff43a71afd9c3ff111d018af161859abae0bdb0b3dae98f9:0x140001010
+  features:
+    - and:
+      - or:
+        - match: allocate RWX memory
+        - basic block:
+          - and:
+            # xor     ecx, ecx        ; lpAddress
+            # mov     edx, 31Fh       ; dwSize
+            # mov     r8d, 1000h      ; flAllocationType
+            # lea     r9d, [rcx+40h]  ; flProtect
+            # call    cs:VirtualAlloc
+            - match: allocate memory
+            - mnemonic: lea
+            - offset: 0x40
+      - api: CopyFile2
+      - api: DeleteFileW
+      - number: 0x00000001 = COPY_FILE_FAIL_IF_EXISTS
+      - number: 0x20 = sizeof(COPYFILE2_EXTENDED_PARAMETERS)


load-code/shellcode/execute-shellcode-via-copyfile2.yml
```diff
@@ -0,0 +1,28 @@
...
+        - basic block:
+          - and:
+            # xor     ecx, ecx        ; lpAddress
+            # mov     edx, 31Fh       ; dwSize
+            # mov     r8d, 1000h      ; flAllocationType
+            # lea     r9d, [rcx+40h]  ; flProtect
+            # call    cs:VirtualAlloc
+            - match: allocate memory
+            - mnemonic: lea
+            - offset: 0x40
```
williballenthin, line 24: i'd prefer to see this merged into the "allocate RWX memory" rule so other rules get the benefit.

williballenthin (approved):
logic looks good aside from consolidating RWX logic in the other rule

load-code/shellcode/execute-shellcode-via-copyfile2.yml
jtothej: I tried that approach in the beginning but it looks like it would generate a lot of FPs.

Using below logic for "allocate RWX memory":

```
  features:
    - and:
      - match: allocate memory
      - or:
        - number: 0x40 = PAGE_EXECUTE_READWRITE
        - and:
          - mnemonic: lea
          - offset: 0x40
```

and running it against example file from "allocate RWX memory" triggers following match:

```
basic block @ 0x140001499 in function 0x140001364
  and:
    match: allocate memory @ 0x140001499
      or:
        api: NtAllocateVirtualMemory @ 0x1400014C9
    or:
      and:
        mnemonic: lea @ 0x14000149D, 0x1400014AC, 0x1400014B9
        offset: 0x40 @ 0x1400014B9, 0x1400014CF
```

where offset 0x40 is just used to reference a variable:


```
.text:00000001400014B9 48 8D 54 24 40          lea     rdx, [rsp+138h+var_F8] ; BaseAddress
.text:00000001400014BE C7 44 24 20 00 30 00 00 mov     dword ptr [rsp+138h+var_118], 3000h
.text:00000001400014C6 49 8B CC                mov     rcx, r12
.text:00000001400014C9 FF 15 99 5D 00 00       call    cs:NtAllocateVirtualMemory
.text:00000001400014CF 48 8B 4C 24 40          mov     rcx, [rsp+138h+var_F8] ; BaseAddress
```

Similarly running it against example file c2bb17c12975ea61ff43a71afd9c3ff111d018af161859abae0bdb0b3dae98f9 from "execute shellcode via CopyFile2" also generates additional matches in the same basic block:

```
basic block @ 0x14000109F in function 0x140001010
  and:
    match: allocate memory @ 0x14000109F
      or:
        api: kernel32.VirtualAlloc @ 0x1400010B0
    or:
      and:
        mnemonic: lea @ 0x1400010AC, 0x1400010CA, 0x1400010FE, 0x140001103, and 1 more...
        offset: 0x40 @ 0x1400010AC, 0x140001119
```

```
.text:000000014000110A 48 8D 0D 77 11 00 00          lea     rcx, aCWindowsDirect ; "C:\\Windows\\DirectX.log"
.text:0000000140001111 FF 15 F9 0E 00 00             call    cs:CopyFile2
.text:0000000140001117 33 C0                         xor     eax, eax
.text:0000000140001119 48 8B 4C 24 40                mov     rcx, [rsp+58h+var_18]

```

I couldn't think of way to make it less FP prone so I used that logic only in this specific rule. Use of CopyFile2 seems to be rather sparse and impact of potential FPs would be low.

williballenthin: ah, ok, that makes sense. we'd had a conversation last year about this problem: https://github.com/mandiant/capa/issues/320 We don't have a great solution at the moment, so lets merge the rule as-is, and keep in mind refactoring it when we have a good solution.

PR #516, opened 2021-12-14
update mappings for mbc (part 7)

dzbeck:
Looked at capa rules and found a few newer rules without MBC/ATT&CK mappings. Please consider these suggested mappings.

### DIFF
--- a/host-interaction/bootloader/manipulate-boot-configuration.yml
+++ b/host-interaction/bootloader/manipulate-boot-configuration.yml
@@ -4,6 +4,8 @@ rule:
     namespace: host-interaction/bootloader
     author: william.ballenthin@mandiant.com
     scope: function
+    att&ck:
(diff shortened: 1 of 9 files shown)


host-interaction/bootloader/manipulate-boot-configuration.yml
```diff
@@ -4,6 +4,8 @@ rule:
     namespace: host-interaction/bootloader
     author: william.ballenthin@mandiant.com
     scope: function
+    att&ck:
+      - Defense Evasion::Impair Defenses [T1562]
```
williballenthin, line 8:
```suggestion
```

this rule we *had* mapped but decided the rule doesn't specifically imply an impact. there's another rule or two that are more related to disabling boot or other security features. so, i think we should keep this rule unmapped.

PR #518, opened 2021-12-20
Add golang runtime pattern

stevemk14ebr:


stevemk14ebr:
lint fails on examples, any golang binary would work. I will not push that though.

runtime/golang/compiled-to-the-standard-golang-runtime.yml
```diff
@@ -0,0 +1,14 @@
+rule:
+  meta:
+    name: compiled to the standard golang runtime
```
mr-tz, line 3: Would you mind merging this into https://github.com/mandiant/capa-rules/blob/master/compiler/go/compiled-with-go.yml?
That rule also has an example.

stevemk14ebr: Yes I can do that, would it make sense to move this to the runtime folder? These rules, especially after my updates, trigger on internals of the runtime rather than any compiler artifacts

stevemk14ebr: Thinking more, the distinction between runtime and compiler, in the context of rules like this seems ambiguous. Mostly the information of interest is more like 'language', if terms were to be changed?

mr-tz: Hm, yeah, there seem to be some inconsistencies under `compiler` and `runtime`. Should we add a `language` namespace?

I'd then see the following associations:

name | namespace
------|------------
autohotkey | language
d | language
exe4j | runtime?
mingw | compiler
perl2exe | runtime?
py2exe | runtime?
rust | language
autoit | language
delphi | language
go | language
nim | language
ps2exe | runtime?
pyarmor | obfuscator?
vb | language
dotnet | language

stevemk14ebr: imho I think perhaps just 'language' and 'obfuscator' the distinction between runtime/language/compiler I personally feel is kind of hard to distinguish between when you're just working with rules like this. As an example .NET, Go, Nim, and I'm sure more languages have 'runtime' things you can signature off of, but those runtimes are closely tied to the compiler used by that language, so my argument is that's not a super useful distinction. 

Obfuscators are a whole different breed though and emit artifacts that could be easily distinguished. Perhaps the name obfuscator/bundler/packager would be a good name, this would include bundlers like py2exe and ps2exe which aren't really 'runtimes'.

suggested terms:
* language
* obfuscator/bundler/packager (is there a unified term?)

williballenthin: - language (go)
  - compiler (msvc vs mingw)
  - packer (upx)
  - installer (inno)

im not sure where py2exe goes. i guess "packager" if we wanted yet another category.

however, for capa i dont think we really want to be in the business of building thorough taxonomies. sure, we want to be precise, but as you know we can debate this stuff for weeks and never be perfectly happy. so, im hesitant about creating a bunch of new categories that require documentation for users to figure out which is which.

stevemk14ebr: I'm suggesting less categories, not more. Ideally 1: language, maybe 2: language + obfuscator/bundler/packager.

If you don't want to do this then may i suggest .net be moved to compiler and the /runtime removed.

compiler/go/compiled-with-go.yml
```diff
@@ -11,3 +11,7 @@ rule:
       - string: /^Go build ID:/
       - substring: "go.buildid"
       - string: /^Go buildinf:/
+      - substring: "go1."
+      - substring: "runtime.main"
+      - substring: "main.main"
+      - substring: "runtime.gcWork"
```
mr-tz, line 17:
```suggestion
      - substring: "runtime.gcWork"

```

mr-tz:
thanks, we'll track the namespace cleanup in #519

PR #522, opened 2022-01-22
Updated and graduated spoof-parent-pid.yml

jtothej:
Added example and attack mapping to spoof-parent-pid.yml. Moved to anti-analysis/anti-forensic from nursery.


mr-tz (approved):
great, thanks!

williballenthin (approved):

PR #525, opened 2022-01-23
Adding block-operations-on-executable-memory-pages-using-arbitrary-code-guard.yml

jtothej:
New rule that identifies use of SetProcessMitigationPolicy to set dynamic code policy (Arbitrary Code Guard) of the process to prevent it from creating and modifying code pages in memory - possibly preventing injection of EDR DLLs.


williballenthin (approved):
very cool

PR #535, opened 2022-02-24
Update create-process-on-windows.yml mapping to ATT&CK T1106

erhlee-bird:
Just adding the att&ck category for T1106 to the create-process-on-windows.yml rule.
I think this makes sense for the Windows rule.

https://attack.mitre.org/techniques/T1106/

### DIFF
--- a/host-interaction/process/create/create-process-on-windows.yml
+++ b/host-interaction/process/create/create-process-on-windows.yml
@@ -5,6 +5,8 @@ rule:
     author:
       - moritz.raabe@mandiant.com
     scope: basic block
+    att&ck:
+      - Execution::Native API [T1106]
     mbc:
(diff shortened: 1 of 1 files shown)


mr-tz:
Hm, interesting. The way I read this is that we could assign this category to every rule using a Windows API call?
So I am not sure if this is really helpful.

mr-tz:
Hm, I'm still not convinced here.
https://attack.mitre.org/tactics/TA0002/ describes Execution as "The adversary is trying to run malicious code."
That also applies to potentially all rules we have.

erhlee-bird:
Cool, totally understand. Thank you for taking the time to consider and review.

mr-tz:
Of course, please keep ideas, suggestions, and feedback coming!

PR #536, opened 2022-03-01
call $+5 update rule and doc

mr-tz:
see https://github.com/mandiant/capa/pull/899 and https://github.com/mandiant/capa/pull/902


mr-tz:
depends on https://github.com/mandiant/capa/issues/903

williballenthin (approved):

PR #543, opened 2022-03-21
Adding compiled-with-v.yml and compiled-with-zig.yml

jtothej:
New rules that identify executables compiled with V(lang) and Zig(lang).

### DIFF
--- a/compiler/zig/compiled-with-zig.yml
+++ b/compiler/zig/compiled-with-zig.yml
@@ -10,16 +10,13 @@ rule:
     examples:
       - 74e0758e469ab87f38cda7925d696ea41a122364891860e1ae30fc9ff2f68e7d
   features:
-    - and:
-      - function:
-        - or:
-          - api: ntdll.NtLockFile
-          - api: ntdll.NtCreateFile
-          - api: ntdll.NtClose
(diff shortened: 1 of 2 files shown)


compiler/zig/compiled-with-zig.yml
```diff
@@ -0,0 +1,25 @@
...
+      - 74e0758e469ab87f38cda7925d696ea41a122364891860e1ae30fc9ff2f68e7d
+  features:
+    - and:
+      - function:
+        - or:
+          - api: ntdll.NtLockFile
+          - api: ntdll.NtCreateFile
+          - api: ntdll.NtClose
+          - api: ntdll.NtCreateKeyedEvent
+          - api: ntdll.NtWaitForKeyedEvent
```
williballenthin, line 20: this looks windows specific. should we add `os: windows` or include this characteristic in the rule name?

williballenthin (approved):
thank you!

compiler/zig/compiled-with-zig.yml
jtothej: Good point! I updated the rule and removed Windows specific entries. Strings seem to be specific enough so we should be good without that part.

mr-tz (approved):

PR #547, opened 2022-03-24
Adding packed-with-huan.yml

jtothej:
New rule that identifies files "packed" (encrypted) with Huan (https://github.com/frkngksl/Huan).


williballenthin (approved):
thank you!

mr-tz (approved):

mr-tz:
not sure why the thorough lint fails here

mr-tz:
I think it's because we grep for "name: " in the CI script and that's part of the rule text. So merging here.

PR #553, opened 2022-04-14
added parse-pe-export-data rule

sara-rn:
it detects manual parsing of the PE Export Data when resolving function RVAs dynamically.
lint.py fails to execute.


nursery/parse-pe-export-data.yml
```diff
@@ -0,0 +1,33 @@
...
+        - 3 or more:
+          - offset: 0x14 = IMAGE_EXPORT_DIRECTORY.NumberOfFunctions
+          - offset: 0x24 = IMAGE_EXPORT_DIRECTORY.AddressOfNameOrdinals
+          - offset: 0x20 = IMAGE_EXPORT_DIRECTORY.AddressOfNames
+          - offset: 0x18 = IMAGE_EXPORT_DIRECTORY.NumberOfNames
+          - offset: 0x1C = IMAGE_EXPORT_DIRECTORY.AddressOfFunctions
+      - optional:
+        - or:
+          - api: LoadLibrary
+          - api: strcmp
```
mr-tz, line 33:
```suggestion
          - api: strcmp

```

nursery/parse-pe-export-data.yml
```diff
@@ -0,0 +1,33 @@
+rule:
+  meta:
+    name: find RVA from a function by parsing the PE Export Data
```
mr-tz, line 3:
```suggestion
    name: resolve function by parsing PE exports
```

mr-tz: just an idea to make this more concise

nursery/parse-pe-export-data.yml
```diff
@@ -0,0 +1,33 @@
...
+    author: sara-rn
+    scope: function
+    examples:
+      - 73CE04892E5F39EC82B00C02FC04C70F
+  features:
+    - and:
+      - os: windows
+      - or:
+        - characteristic: loop
+        - mnemonic: movzx
```
mr-tz, line 14: could this also be `mov`?

mr-tz (changes requested):
Nice start, please see the linter output in CI:
```
     (nursery) find RVA from a function by parsing the PE Export Data 
      WARN: filename doesn't match the rule name: Rename rule file to match the rule name, expected: "find-rva-from-a-function-by-parsing-the-pe-export-data.yml", found: "parse-pe-export-data.yml"
      WARN: missing example offset: Add offset of example function
      WARN: EOF format: end file with a single empty line
      WARN: rule format incorrect: use scripts/capafmt.py or adjust as follows
```

nursery/parse-pe-export-data.yml
sara-rn: sounds good

nursery/parse-pe-export-data.yml
sara-rn: I have seen movzx in 2 samples so far.
mov     eax, [ebp+addressofNamesOrd]
movzx   eax, word ptr [eax+ebx*2]
In the chapter 19 of PMA the same is done with mov: mov  cx, word ptr [ebx+ecx*2]

sara-rn: by removing the movz there will be FP I believe but maybe there is less chance to miss hits, let me know what you think

mr-tz: sorry, I wasn't clear, I meant to suggest to add `mov`?

williballenthin, line 14:
```suggestion
        - mnemonic: movzx
        - mnemonic: mov
```

mr-tz:
Thanks!

PR #557, opened 2022-05-11
Update rules to use a list item instead of a string value for reference

idiom:
Several rules use a string value to define a reference. This can cause issues in calling code as the value may be a list or a string. Checking the defined format states that this field lists related information, so I believe this should always be a list of references. 

Updated format.md to define the field as a list of related info.  
Updated these rules to define the reference value as a list entry.


williballenthin:
thank you @idiom! great catches.

we'll also update the linter to assert that the references are the correct type.

williballenthin (approved):

mr-tz:
great, thanks!

PR #559, opened 2022-05-26
*: use meta.authors when there's more than one author

williballenthin:
https://github.com/mandiant/capa/issues/1027

so when there's a single author, we continue to use `author` but when there are multiple, then we do `authors`.
thoughts?

can easily use `authors` everywhere. see alternative in #560


williballenthin:
```
rg -l "author:\$" | while read LINE; sed -i -e "s/author: /authors:\n      - /g" $LINE; end
```

mr-tz:
I prefer #560 over this.

PR #560, opened 2022-05-26
*: use meta.authors everywhere

williballenthin:
https://github.com/mandiant/capa/issues/1027

use `meta.authors` everywhere. its a little less natural when there's a single  author. our code will be a bit simpler than whats needed in #559 to normalize `author`  -> `authors`.


williballenthin:
```
rg -l "author:\$" | while read LINE; sed -i -e "s/author: /authors:\n      - /g" $LINE; end
rg -l "author: " | while read LINE; sed -i -e "s/author: /authors:\n      - /g" $LINE; end
```

mike-hunhoff:
I prefer this over #559; let's keep it simple.

mr-tz (approved):
thanks! note: this won't break rule loading (only displaying, which is not a problem, because we want to omit the author in vverbose output anyway [see https://github.com/mandiant/capa/issues/993]).

williballenthin:
thanks!

PR #564, opened 2022-06-06
use new .NET features for existing .NET rules

williballenthin:
don't merge this until we're reading for breaking changes


mike-hunhoff (approved):
LGTM 🚀

PR #566, opened 2022-06-09
dotnet rule updates

mike-hunhoff:
adding new dotnet rules and updating existing rules with dotnet APIs.


communication/http/client/send-http-request.yml
```diff
@@ -12,6 +13,7 @@ rule:
       - 6A352C3E55E8AE5ED39DC1BE7FB964B1:0x100026E0
   features:
     - or:
+      - api: System.Net.WebRequest::GetResponse
```
mr-tz, line 16: shouldn't it be `Create` or similar here?

nursery/mixed-mode-dotnet.yml
```diff
@@ -0,0 +1,9 @@
+rule:
+  meta:
+    name: mixed mode dotnet
```
mr-tz, line 3: would be helpful to add a description of what this means

nursery/unmanaged-call-dotnet.yml
```diff
@@ -0,0 +1,10 @@
+rule:
+  meta:
+    name: unmanaged call dotnet
```
mr-tz, line 3: would be helpful to add a description of what this means

mr-tz (approved):
nice, added a few minor comments

nursery/unmanaged-call-dotnet.yml
williballenthin, line 3:
```suggestion
    name: unmanaged .NET call
```

williballenthin:
ping @mike-hunhoff for comments above by moritz

nursery/unmanaged-call-dotnet.yml
mike-hunhoff: updated w/ description

mike-hunhoff: We may want to make this more generic to open up the rule to other runtimes that may be supported in the future.

nursery/mixed-mode-dotnet.yml
mike-hunhoff: added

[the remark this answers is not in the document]
mike-hunhoff: We may want to make this more generic to open up the rule to other runtimes that may be supported in the future.

PR #567, opened 2022-06-23
Update compiled-with-borland-delphi.yml

mr-tz:
Extend Delphi detection


mike-hunhoff (approved):
looks great, thank you!

PR #568, opened 2022-06-23
Add load-dotnet-assembly.yml

anushkavirgaonkar:
Adding a new rule that detects when a sample loads a .NET assembly.

### DIFF
--- a/nursery/load-dotnet-assembly.yml
+++ b/nursery/load-dotnet-assembly.yml
@@ -1,6 +1,6 @@
 rule:
   meta:
-    name: load dotnet assembly
+    name: load .NET assembly
     namespace: load-code/dotnet
     authors:
       - anushka.virgaonkar@mandiant.com
--- a/nursery/load-dotnet-assembly.yml
+++ b/nursery/load-dotnet-assembly.yml
@@ -1,0 +1,19 @@
+rule:
+  meta:
+    name: load dotnet assembly
+    namespace: load-code/dotnet
+    authors:
+      - anushka.virgaonkar@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Reflective Code Loading [T1620]
+  features:
+    - or:
+      - api: System.Reflection.Assembly::Load
+      - api: System.AppDomain::Load
+      - api: System.Reflection.Assembly::LoadFile
+      - api: System.Reflection.Assembly::LoadFrom
+      - api: System.Reflection.Assembly::LoadWithPartialName
+      - api: System.Reflection.Assembly::ReflectionOnlyLoad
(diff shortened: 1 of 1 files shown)


mike-hunhoff (approved):
This is great, thank you! LGTM 🚀

nursery/load-dotnet-assembly.yml
```diff
@@ -0,0 +1,19 @@
+rule:
+  meta:
+    name: load dotnet assembly
```
mr-tz, line 3: final question before we merge, should we use .NET here instead of dotnet?

mike-hunhoff: I'd personally prefer `.NET` but then the filename would need to be `load-.net-assembly.yml`? which I don't like.

williballenthin: i think we strip out special characters, so it would be `load-net-assembly.yml`.

we could special case `.NET` to `dotnet` in the linter filename check.

williballenthin: also agree that `.NET` is the preferred term.

mr-tz: Yes, the `.` is stripped so the rule file would be `load-net-assembly.yml`

williballenthin: see https://github.com/mandiant/capa/pull/1068

williballenthin: currently have two other affected rules:

```console
❯ python scripts/lint.py rules/
INFO:lint:successfully loaded 774 rules
INFO:lint:collecting potentially referenced samples
                                                                                                                     
compiled to the .NET platform                                                                                        
  FAIL: filename doesn't match the rule name: Rename rule file to match the rule name, expected: "compiled-to-the-dotnet-platform.yml", found: "compiled-to-the-net-platform.yml"
                                                                                                                     
                                                                                                                     
encrypt data using AES via .NET                                                                                      
  FAIL: filename doesn't match the rule name: Rename rule file to match the rule name, expected: "encrypt-data-using-aes-via-dotnet.yml", found: "encrypt-data-using-aes-via-net.yml"
                                                                                                                     
linting rule: linked against OpenSSL            : 100%|█████████████████████████| 774/774 [00:06<00:00, 117.52rule/s]
rules with FAIL:
  - compiled to the .NET platform
  - encrypt data using AES via .NET
```

williballenthin: other affected rules fixed here: https://github.com/mandiant/capa-rules/pull/573

mr-tz, line 3:
```suggestion
    name: load .NET assembly
```

williballenthin (approved):

Issue #328, opened 2021-04-19
Steal browser credentials


Ana06:
Write rules to detect the ability to capture or mine credentials stored, cached, or used by the browser (Chrome, Opera, Firefox, Internet Explorer).

## Examples

Publicly available samples that contain the capability this rule should detect (MD5 hashes).

- 7beb638507574423893e89883bcb1161
- 1b8ee96f844c4633feb3f3bcdd43aa9c
- fae0fc8ee982b5998ae9d939777241fb

## References

Links or references to additional information on the capability (can also be included in the rule):

- https://attack.mitre.org/techniques/T1555

## Namespace

Proposed namespace: `collection/browser`. More details in https://github.com/fireeye/capa-rules/blob/master/doc/format.md#rule-namespace

## att&ck

`aka.mitre.att&ck.t1555.003`

PR #578, opened 2022-06-30
Create gather-chrome-based-browser-login-information.yml

re-fox:
Simple string-based rule to look for Chrome (and Chrome-based) browser login data being referenced.   I've included a stealer in https://github.com/mandiant/capa-testfiles/pull/135 that is referenced in this rule. 

May help contribute to https://github.com/mandiant/capa-rules/issues/328.  

I leaned on regex for case-insensitivity, if there are any performance concerns I'd be happy to use regular string operators.


williballenthin (approved):

williballenthin:
thank you!

PR #581, opened 2022-07-01
ci: lint thorough on PR text

mr-tz:
Lint all rules thoroughly when this is set:

- [X] lint thorough all


williballenthin:
ooo this is cool!

williballenthin:
only thing i'd ask is that you document what this does in the workflow. i had to study it for a moment to understand what was happening.

williballenthin (approved):

williballenthin:
thanks!

PR #590, opened 2022-07-07
Add dotnet rules

anushkavirgaonkar:
The following rules are updated with dotnet capabilities:
- `accept command line arguments`
- `get networking interfaces`
- `get session user name`


mike-hunhoff (approved):
very good, thank you @anushkavirgaonkar!

PR #594, opened 2022-07-18
Removed global-level feature format from function-level capabilities.

adamstorek:
Removed global-level feature format "dotnet" from function-level capabilities.


mr-tz:
Outdated?

PR #597, opened 2022-07-25
Add act-as-credential-manager-dll.yml and act-as-password-filter-dll.yml

jtothej:
Adding two new rules for DLLs designed to act as Windows password filter or credential manager


williballenthin (approved):

mike-hunhoff (approved):
thank you @jtothej!

PR #601, opened 2022-07-28
Add dotnet rules having property features

anushkavirgaonkar:


communication/ftp/send/send-file-using-ftp.yml
```diff
@@ -0,0 +1,28 @@
...
+        - optional:
+          - or:
+            - api: wininet.FtpSetCurrentDirectory
+            - and:
+              - api: wininet.InternetConnect
+              - number: 0x15 = IPPORT_FTP
+      - and:
+        - api: System.Net.WebRequest::Create
+        - class: System.Net.FtpWebRequest
+        - string: STOR
```
mike-hunhoff, line 27:
```suggestion
        - string: "STOR"
```

communication/socket/tcp/create-tcp-socket.yml
```diff
@@ -11,11 +11,13 @@ rule:
...
   features:
-    - and:
-      - number: 6 = IPPROTO_TCP
-      - number: 1 = SOCK_STREAM
-      - number: 2 = AF_INET
-      - or:
-        - api: ws2_32.socket
-        - api: ws2_32.WSASocket
-        - api: socket
+    - or:
```
mike-hunhoff, line 14: Don't forget to update the `authors` field

host-interaction/gui/enumerate-gui-resources.yml
```diff
@@ -18,3 +18,4 @@ rule:
       - api: EnumWindowStations
       - api: EnumDesktops
       - api: EnumWindows
+      - property: System.Windows.Forms.Screen::AllScreens
```
mike-hunhoff, line 21: Don't forget to update the `authors` field.

mike-hunhoff (changes requested):
Thank you @anushkavirgaonkar, great job! Just a few minor comments to address.

communication/socket/tcp/create-tcp-socket.yml
anushkavirgaonkar: Noted. https://github.com/mandiant/capa-rules/pull/601/commits/b3cd0abc6de717aaedda04b51804f1c562c8b472 updates the field.

host-interaction/gui/enumerate-gui-resources.yml
anushkavirgaonkar: Noted. https://github.com/mandiant/capa-rules/pull/601/commits/b3cd0abc6de717aaedda04b51804f1c562c8b472 updates the field.

mike-hunhoff (approved):
great, thank you! Following the v4.0.0 release we can get this merged.

anti-analysis/anti-debugging/debugger-detection/check-for-debugger-via-api.yml
```diff
@@ -18,3 +19,4 @@ rule:
       - api: WUDFPlatform.WudfIsAnyDebuggerPresent
       - api: WUDFPlatform.WudfIsKernelDebuggerPresent
       - api: WUDFPlatform.WudfIsUserDebuggerPresent
+      - property: System.Diagnostics.Debugger::IsAttached
```
mr-tz, line 22: add `matches: compiled to the .NET platform` to set a good example? (see https://github.com/mandiant/capa/discussions/1106#discussioncomment-3202916)

communication/ftp/send/send-file-using-ftp.yml
```diff
@@ -0,0 +1,28 @@
...
+          - or:
+            - api: wininet.FtpSetCurrentDirectory
+            - and:
+              - api: wininet.InternetConnect
+              - number: 0x15 = IPPORT_FTP
+      - and:
+        - api: System.Net.WebRequest::Create
+        - class: System.Net.FtpWebRequest
+        - string: "STOR"
+        - property: System.Net.WebRequest::Method
```
mr-tz, line 28: Are the `api`, `class`, and `property` expected to be extracted from the same instruction?
Is there duplication in this case?

anti-analysis/anti-debugging/debugger-detection/check-for-debugger-via-api.yml
mr-tz: this rule is pretty clear but we should come up with a consistent solution

communication/http/read-http-header.yml
```diff
@@ -4,11 +4,13 @@ rule:
...
     scope: function
     mbc:
       - Communication::HTTP Communication::Read Header [C0002.014]
     examples:
       - 6A352C3E55E8AE5ED39DC1BE7FB964B1:0x10002A30
   features:
-    - and:
+    - or:
       - api: winhttp.WinHttpQueryHeaders
+      - property: System.Net.WebClient::Headers
```
mr-tz, line 16: see above on diff between formats

communication/http/set-http-header.yml
```diff
@@ -4,11 +4,13 @@ rule:
...
     scope: function
     mbc:
       - Communication::HTTP Communication::Set Header [C0002.013]
     examples:
       - 6A352C3E55E8AE5ED39DC1BE7FB964B1:0x1000E230
   features:
-    - and:
+    - or:
       - api: winhttp.WinHttpAddRequestHeaders
+      - property: System.Net.WebClient::Headers
```
mr-tz, line 16: can we distinguish get vs. set of a property? if not, should we limit this to one rule?

mr-tz (commented):
nice updates, my only concern is around adding format specifications, see https://github.com/mandiant/capa/discussions/1106#discussioncomment-3202916

communication/http/set-http-header.yml
mike-hunhoff: Per discussion from https://github.com/mandiant/capa/issues/939 we've opted not to differentiate get v. set although I believe we would have all of the information to do so. Curious to @anushkavirgaonkar 's thoughts on the our ability to differentiate between get v. set using .NET metadata.

The alternative to differentiating get v. set is to label all property rules as "access" which I think is limiting and forces use to create more rules than ideally necessary. I prefer adding the ability to differentiate get v. set.

anti-analysis/anti-debugging/debugger-detection/check-for-debugger-via-api.yml
mike-hunhoff: I like the idea of `matches: xyz technology` where we expect the former to be included in all rule blocks matching `xyz` technology. We can then maintain the one`xyz technology` rule.

communication/ftp/send/send-file-using-ftp.yml
anushkavirgaonkar: No, they are extracted from different instructions.

communication/http/set-http-header.yml
anushkavirgaonkar: The `System.Net.WebClient.Headers` property is being used for both getting and setting a collection of headers. I understand that it would mean that both the rules `set http header` and `read http header` would be matched when the property is referenced in the sample. Since we are not distinguishing between get vs set, should I keep the rules as they are (and we'd have to be okay with having a false positive) or should I remove the rules for maintaining accuracy?

communication/ftp/send/send-file-using-ftp.yml
mr-tz: Ok, I'm having trouble following this. Can you include a comment of an example code snippet?

communication/http/set-http-header.yml
mr-tz: Mike raises a good point. @anushkavirgaonkar do you think it's worth adding the differentiation? Would you be able to get this feature in?
From what we learned with other features I think it makes sense to add it from the start, it's more powerful, and we can extract it "easily". We'd need to agree on a good syntax though.

communication/ftp/send/send-file-using-ftp.yml
anushkavirgaonkar: image

communication/http/set-http-header.yml
anushkavirgaonkar: Sure! Differentiating between read and write sounds like a good idea. I've added commits to https://github.com/mandiant/capa/pull/1122 for emitting the features and have updated the rules here as well. I've followed the option 2 for the rule syntax from here: https://github.com/mandiant/capa/issues/939#issuecomment-1210406078

mr-tz: Thanks, this is great!

communication/ftp/send/send-file-using-ftp.yml
mr-tz: Thanks, so we could remove `class: System.Net.FtpWebRequest` and still be good?

mr-tz: Unrelated, can we do this?
```yaml
instruction:
  - property: System.Net.WebRequest::Method
  - string: STOR
```

anushkavirgaonkar: Yes, we can remove `class: System.Net.FtpWebRequest`

image
Since the features are extracted from different instructions, we won't be getting a rule match.

mr-tz: Thanks. And yes, never mind my `instruction` idea :)

PR #603, opened 2022-08-04
Additional rules to support capa-scripts.

adamstorek:
I noticed that the current .NET rules missed a lot of functionality/API calls that the aspx test malware includes. I therefore modified the currently available rules and created new ones to fully leverage the capa-scripts extraction capabilities.

### DIFF
--- a/c2/shell/create-reverse-shell.yml
+++ b/c2/shell/create-reverse-shell.yml
@@ -4,6 +4,7 @@ rule:
     namespace: c2/shell
     authors:
       - moritz.raabe@mandiant.com
+      - adam.storek@mandiant.com
     scope: function
     att&ck:
       - Execution::Command and Scripting Interpreter::Windows Command Shell [T1059.003]
@@ -11,6 +12,8 @@ rule:
       - Impact::Remote Access::Reverse Shell [B0022.001]
     examples:
       - C91887D861D9BD4A5872249B641BC9F9:0x401A77
+      - 2b71dd245520d9eb5f1e4c633fee61c7d83687591d9f64f9390c26dc95057c3c.aspx_
+      - 2e8c7eacd739ca3f3dc4112b41a024157035096b8d0c26ba79d8b893136391bc.aspx_
   features:
     - or:
       - and:
@@ -30,3 +33,8 @@ rule:
           - and:
             - count(api(SetHandleInformation)): 2 or more
             - number: 1 = HANDLE_FLAG_INHERIT
+      - and:
+        - match: create process on Windows
+        - or:
+          - substring: "cmd.exe"
+          - substring: "powershell.exe"
--- a/data-manipulation/encryption/aes/decrypt-data-using-aes.yml
+++ b/data-manipulation/encryption/aes/decrypt-data-using-aes.yml
@@ -1,0 +1,23 @@
+rule:
+  meta:
+    name: decrypt data using AES
+    namespace: data-manipulation/encryption/aes
+    authors:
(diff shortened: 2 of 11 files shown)


mike-hunhoff (changes requested):
great, thank you! I've left comments and suggestions for your review 🚀. I'm adding a "do not merge" tag for now as we won't be able to merge these rules until after the v4.0.0 release.

c2/shell/create-reverse-shell.yml
```diff
@@ -30,3 +33,8 @@ rule:
           - and:
             - count(api(SetHandleInformation)): 2 or more
             - number: 1 = HANDLE_FLAG_INHERIT
+      - and:
+        - match: create process on Windows
+        - or:
+          - substring: "cmd.exe"
+          - substring: "powershell.exe"
```
mr-tz, line 40: is this sufficient to identify a reverse shell?

data-manipulation/encryption/aes/decrypt-data-using-aes.yml
```diff
@@ -0,0 +1,23 @@
+rule:
+  meta:
+    name: decrypt data using AES
```
mr-tz, line 3:
```suggestion
    name: decrypt data using AES via .NET
```

mr-tz: to be consistent with following rule

data-manipulation/hashing/md5/hash-data-with-md5.yml
```diff
@@ -32,6 +32,5 @@ rule:
           - number: 0x8003 = CALG_MD5
           - api: advapi32.CryptCreateHash
       - and:
-        - format: dotnet
```
mr-tz, line 35: need to find a good solution for this, see https://github.com/mandiant/capa/discussions/1106#discussioncomment-3202916 and #601

load-code/powershell/run-powershell-expression.yml
```diff
@@ -4,11 +4,13 @@ rule:
     namespace: load-code/powershell/
     authors:
       - anamaria.martinezgom@mandiant.com
+      - adam.storek@mandiant.com
     scope: function
     att&ck:
       - Execution::Command and Scripting Interpreter::PowerShell [T1059.001]
     examples:
       - 692f7fd6d198e804d6af98eb9e390d61:0x6000004
+      - 4f6fa6a45017397c7e1c9cd5a17235ccb1ff0f5087dfa6b7384552bf507e7fe1.aspx_
```
mr-tz, line 13: so cool, that this just works :)

nursery/encrypt-data-using-rsa.yml
```diff
@@ -4,7 +4,16 @@ rule:
...
       - michael.hunhoff@mandiant.com
+      - adam.storek@mandiant.com
     scope: function
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
+    mbc:
+      - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
+      - Cryptography::Encrypt Data::RSA [C0027.011]
+    examples:
+      - d460cae7d34c51059ef57c5aadb3de099469efbac5fffcf76d0528a511192a28.aspx_
```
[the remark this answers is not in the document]
mr-tz: leads to the question if we expect a test case for binary and script and <other format>...?!

mr-tz (changes requested):
great, please see my comments for further improvements

PR #610, opened 2022-08-15
lint --thorough

mr-tz:
- [x] lint thorough all


mr-tz:
All lints pass except "mutex and exit" which I cannot replicate locally. Seems to be an issue with function ID that does not appear every time.

PR #614, opened 2022-08-31
Rule for accessing Windows directory from KUSER_SHARED_DATA

edeca:
This structure is at a fixed offset in both 32 and 64 bit Windows. This rule looks for a typical function call setup to copy data from this location.

It may be possible to instead look for only the known memory address, which would catch samples that manually access memory rather than calling a function. However, this would be more prone to false positives where the same constants are used for other reasons.

### DIFF
--- a/nursery/get-windows-directory-from-kuser_shared_data.yml
+++ b/nursery/get-windows-directory-from-kuser_shared_data.yml
@@ -11,5 +11,5 @@ rule:
     examples:
       - 7f15b1a47bbe031334e23653879e9661f4b8cde80c307548328fdd3aed87ca46:0x01019080
   features:
-    # Fixed offset of Windows directory path in KUSER_SHARED_DATA structure.
+    # Fixed offset of Windows directory path in KUSER_SHARED_DATA structure; same for 32- and 64-bit Windows.
     - number: 0x7ffe0030 = KUSER_SHARED_DATA.NtSystemRoot
--- a/nursery/get-windows-directory-from-kuser_shared_data.yml
+++ b/nursery/get-windows-directory-from-kuser_shared_data.yml
@@ -1,6 +1,6 @@
 rule:
   meta:
-    name: obtain Windows directory from KUSER_SHARED_DATA
+    name: get Windows directory from KUSER_SHARED_DATA
     namespace: host-interaction/file-system
     authors:
       - david.cannings@pwc.com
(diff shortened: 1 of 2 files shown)


edeca:
Thanks @mr-tz - after reviewing the other rule I think I was obsessing too much over reducing false positives. Simplified.

mr-tz:
Great, thanks. This looks good to me.

mr-tz:
Let's see if either Mike or Willi have feedback and then we can merge.

nursery/get-windows-directory-from-kuser_shared_data.yml
```diff
@@ -0,0 +1,15 @@
...
+      - david.cannings@pwc.com
+    scope: basic block
+    references:
+      - http://www.rohitab.com/discuss/topic/42325-the-kuser-shared-data-structure/
+      - https://www.geoffchappell.com/studies/windows/km/ntoskrnl/inc/api/ntexapi_x/kuser_shared_data/index.htm
+    examples:
+      - 7f15b1a47bbe031334e23653879e9661f4b8cde80c307548328fdd3aed87ca46:0x01019080
+  features:
+    # Fixed offset of Windows directory path in KUSER_SHARED_DATA structure.
+    - number: 0x7ffe0030 = KUSER_SHARED_DATA.NtSystemRoot
```
williballenthin, line 15: is this offset the same for both 32 and 64 bit windows?

williballenthin: > KUSER_SHARED_DATA is a data structure that contains a lot of Windows system information. It is always located in user mode at 0x7ffe0000, even on 64-bit Windows. At 0x7ffe0030, the full path of the Windows directory is stored there. The Windows version number is stored at 0x7ffe026c and 0x7ffe0270. 0x7ffe026c is the major version number, and 0x7ffe0270 is the minor version number.

http://www.rohitab.com/discuss/topic/42325-the-kuser-shared-data-structure/

well, ok then!

given that this is an obvious question, i'd recommend a short comment indicating that we've already checked and confirmed. otherwise, looks good!

williballenthin (approved):
thank you!

nursery/get-windows-directory-from-kuser_shared_data.yml
```diff
@@ -0,0 +1,15 @@
...
+    authors:
+      - david.cannings@pwc.com
+    scope: basic block
+    references:
+      - http://www.rohitab.com/discuss/topic/42325-the-kuser-shared-data-structure/
+      - https://www.geoffchappell.com/studies/windows/km/ntoskrnl/inc/api/ntexapi_x/kuser_shared_data/index.htm
+    examples:
+      - 7f15b1a47bbe031334e23653879e9661f4b8cde80c307548328fdd3aed87ca46:0x01019080
+  features:
+    # Fixed offset of Windows directory path in KUSER_SHARED_DATA structure.
```
mr-tz, line 14:
```suggestion
    # Fixed offset of Windows directory path in KUSER_SHARED_DATA structure; same for 32- and 64-bit Windows.
```

Issue #615, opened 2022-09-01
Improvement to bypass-uac-via-token-manipulation


edeca:
The rule `[bypass UAC via token manipulation](https://github.com/mandiant/capa-rules/blob/master/host-interaction/uac/bypass/bypass-uac-via-token-manipulation.yml)` currently looks for one process typically abused by UAC bypass techniques (`wusa.exe`). It also assumes the process will be run using `ShellExecuteExW`.

Two improvements could be considered:
- Look for APIs associated with process enumeration (e.g. `EnumProcesses`).
- Look for a wider selection of processes (e.g. [this list](https://gist.github.com/dezhub/c0fee68d1e06657a45ec39365362fca7#file-uacbypasstokenmanipulation-py-L365)).

mr-tz:
I think both of these would be good improvements. Feel free to update the open PR as you see fit.

PR #616, opened 2022-09-01
Improve UAC bypass rule based on sample in the wild

edeca:
Update rule based on analysis of a sample from the wild. This sample has not been added to `examples` because it needs to be unpacked manually from memory.

The APIs have been broken into three categories:
- APIs used to run an elevated program.
- APIs for obtaining token information.
- APIs for using the token.

I removed `GetStartupInfoW`, which is not essential to the UAC bypass technique and is typically used to run the elevated process in a hidden window.

Only the string `wusa.exe` is currently included. Opened issue #615 to discuss whether this should be expanded.


mr-tz (approved):
Great improvement and documentation update. Thank you!

edeca:
Thanks @mr-tz - I will try to find some additional samples and update the rule when time allows.

PR #622, opened 2022-09-16
lzo encryption

edeca:
Work in progress rule for an LZO encryption routine found in a malware sample. The sample requires unpacking from memory, therefore searching for an alternate example.

Not ready to be merged yet.


edeca:
PR should not have included all these files - closing.

PR #624, opened 2022-09-20
Add initial UAC bypass with RPC rule

edeca:
Quick rule to detect UAC bypass using RPC.

One query on the inclusion of the byte version of the GUID. This does appear in the reference sample (at 0x140020EC0) and is referenced in the function that we expect test cases to hit. However, the reference is not direct because the code references a `MIDL_STUB_DESC` structure (at 0x140020D60) which then references the bytes in an `RpcInterfaceInformation` structure.

Advice on how to handle this appreciated - would it be worth moving the bytes to `file` scope?

### DIFF
--- a/host-interaction/uac/bypass/bypass-uac-via-rpc.yml
+++ b/host-interaction/uac/bypass/bypass-uac-via-rpc.yml
@@ -1,0 +1,33 @@
+rule:
+  meta:
+    name: bypass UAC via RPC
+    namespace: host-interaction/uac/bypass
+    authors:
+      - david.cannings@pwc.com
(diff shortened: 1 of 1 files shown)


edeca:
Needs mandiant/capa-testfiles#144 merged to pass CI.

mr-tz:
Thanks a lot for the sample, rule, and the details.

Unfortunately, we cannot fully express this with current capa features. The `bytes` feature is not supported at the file scope and the indirect reference is not caught. However, this rule looks good to me here.

mr-tz (approved):

Issue #629, opened 2022-10-31
FNs introduced during revert whiplash


williballenthin:
@cwakil17 in https://github.com/mandiant/capa-rules/discussions/628: 

When I run capa v4.0 in our automated tests, it loses a couple capabilities in the report compared to v3.2.1 due to differences in the rule .yml files. If I edit the .yml files to match, it finds the capabilities. The .yml files I was looking at specifically are hash-data-using-murmur3.yml, check-http-status-code.yml, and calculate-modulo-256-via-x86-assembly.yml.

I see that the original commit that changed this behavior was "remove /x32 and /x64 flavors and use instruction scope" (bc28847dd979298684d6d3917b11245837f73ec8). That commit was later reverted, and then reverted again (restoring the original). Could you elaborate on the reason for this change, since it seems to have removed results?

williballenthin:
If our git commit whiplash caused some matches to now be missed - that's a bug. Thanks for calling this out and lets get these fixed.

All the reverts had to do with us introducing breaking changes to the rule format and updating some rules before the v4 release, but people would pull from master on v3, and then their rules wouldn't work. So, in v4 we clarified what the behavior should be and hopefully fixed that problem. But, in the interim, we made a bunch of commits and reverted and then reverted again. I thought it went through cleanly but seems like not! Thanks again for raising this.

Do you want to propose the PRs to capa-rules to fix up the files? Then you can be on the contributor list ;-) Otherwise, I'm happy to do so.

cwakil17:
Okay, great! I created a PR for it: #631

mr-tz:
See the PR discussion for details. In short, the hits in earlier versions were FPs.

PR #631, opened 2022-10-31
Fix revert whiplash issues that caused capability loss (#629)

cwakil17:
When I run capa v4.0 in our automated tests, it loses a few capabilities in the report compared to v3.2.1 due to differences in the rule .yml files. I edited hash-data-using-murmur3.yml, check-http-status-code.yml, and calculate-modulo-256-via-x86-assembly.yml to match the v3 rules (kept new v4 author format) and the capabilities are restored in the report.

### DIFF
--- a/communication/http/client/check-http-status-code.yml
+++ b/communication/http/client/check-http-status-code.yml
@@ -18,7 +18,7 @@ rule:
       - optional:
         - api: atoi
         - api: wininet.HttpQueryInfo
-      - instruction:
+      - basic block:
         - and:
           - or:
             - mnemonic: cmp
--- a/data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
+++ b/data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
@@ -34,22 +34,20 @@ rule:
         - and:
           - description: 'hash >> 16; hash >> 13; hash >> 16'
           - count(mnemonic(shr)): 3
-          - instruction:
-            - mnemonic: shr
-            - number: 16
-          - instruction:
-            - mnemonic: shr
-            - number: 13
+          - number: 16
+          - number: 13
           - optional:
             - count(characteristic(nzxor)): 3 or more
       - and:
         # Group this two blocks under an `and` as on their own they are not
         # unique enough and would cause false positives
-        - instruction:
-          - description: k ROL r1
-          - mnemonic: rol
-          - number: 15 = r1
-        - instruction:
-          - description: hash ROL r2
-          - mnemonic: rol
-          - number: 13 = r2
+        - basic block:
+          - and:
+            - description: k ROL r1
+            - mnemonic: rol
+            - number: 15 = r1
+        - basic block:
+          - and:
+            - description: hash ROL r2
+            - mnemonic: rol
+            - number: 13 = r2
--- a/lib/calculate-modulo-256-via-x86-assembly.yml
+++ b/lib/calculate-modulo-256-via-x86-assembly.yml
@@ -4,7 +4,7 @@ rule:
     authors:
       - moritz.raabe@mandiant.com
     lib: true
-    scope: instruction
+    scope: basic block
     mbc:
       - Data::Modulo [C0058]
     examples:


communication/http/client/check-http-status-code.yml
```diff
@@ -18,7 +18,7 @@ rule:
       - optional:
         - api: atoi
         - api: wininet.HttpQueryInfo
-      - instruction:
+      - basic block:
```
mr-tz, line 21: Can you share the sample or verify why this regresses?
- Does it use a different instruction (e.g. `and`)?
- Or is this an issue with the rule matching engine?

mr-tz (commented):
Thanks for noticing this and the fixes.
The one comment I've added applies to all cases here. My main concern is to ensure that it's not an issue with the `instruction` scope in the rule engine, but the fact that with this rules are too strict. It would be great if you could provide the samples (or screenshots) but I understand if this is not feasible.

communication/http/client/check-http-status-code.yml
cwakil17: [capa_test.zip](https://github.com/mandiant/capa-rules/files/9984538/capa_test.zip)
Here is the exe file I was running the tests on -- it's just Windows Notepad. I put it in a folder and zipped it because GitHub didn't want me uploading executable files. Let me know if you need anything else!

mr-tz: Great, thanks. Is this the test file for all the changed rules here?

cwakil17: Yes, this is the test file for all 3 rules I changed.

mr-tz: I'll go through these here. This his was a FP before, see e.g.
```asm
.text:0000000140013C44 BA C8 00 00 00    mov     edx, 0C8h       ; Msg
.text:0000000140013C49 48 FF 15 58 41 01+call    cs:SendMessageW
...
.text:0000000140013D0F 85 C0             test    eax, eax
```

for hit
```yaml
check HTTP status code
namespace  communication/http/client
author     @mr-tz
scope      function
mbc        Communication::HTTP Communication::Read Header [C0002.014]
examples   54ac78733552a62d1d05ea4ba3fc604bb49fe000d7fc948da45335b726e64d75:0x10001a20
function @ 0x14001371C
  and:
    os: windows
    subscope:
      and:
        or:
          mnemonic: test @ 0x140013D0F
        or:
          number: 0xC8 = OK @ 0x140013C44
...
```

data-manipulation/hashing/murmur/hash-data-using-murmur3.yml
```diff
@@ -34,22 +34,20 @@ rule:
...
         # unique enough and would cause false positives
-        - instruction:
-          - description: k ROL r1
-          - mnemonic: rol
-          - number: 15 = r1
-        - instruction:
-          - description: hash ROL r2
-          - mnemonic: rol
-          - number: 13 = r2
+        - basic block:
```
mr-tz, line 44: These were false positives before. There's no ROL 15 or ROL 13 instructions in this function

lib/calculate-modulo-256-via-x86-assembly.yml
```diff
@@ -4,7 +4,7 @@ rule:
     authors:
       - moritz.raabe@mandiant.com
     lib: true
-    scope: instruction
+    scope: basic block
```
mr-tz, line 7: Not a FP, but I think the `instruction` scope captures better what we're trying to get at here. In general, the basic block scope is too broad and results in more FPs. So, I'd vote to leave this one as is.

```asm
.text:0000000140010CC5 B8 FF 00 00 00    mov     eax, 0FFh
.text:0000000140010CCA 66 23 C8          and     cx, ax
```

mr-tz (commented):
I've provided comments on each hit inline. To summarize, these hits were false positives (FPs) in v3.2.0 etc. (before we had instruction scope) and the updates now more accurately capture these capabilities. Do you agree or should we investigate/discuss this further?

mr-tz:
Closing per discussion above. Thanks, @cwakil17!

PR #638, opened 2022-12-12
added anti-analysis - anti-vm - vpcext

richardweiss80:
Dear Colleagues,

please have a look at my capa rule. I would be pleased to get additional feedback.

Kind regards,
Richard

### DIFF
--- a/nursery/check-for-vm-using-instruction-vpcext.yml
+++ b/nursery/check-for-vm-using-instruction-vpcext.yml
@@ -1,0 +1,21 @@
+rule:
+  meta:
+    name: check for VM using instruction VPCEXT
+    namespace: anti-analysis/anti-vm/vm-detection
+    authors:
+      - richard.weiss@mandiant.com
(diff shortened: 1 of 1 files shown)


nursery/check-for-vm-using-instruction-vpcext.yml
```diff
@@ -0,0 +1,21 @@
+rule:
+  meta:
+    name: check for VM using instruction VPCEXT
+    namespace: anti-analysis/anti-vm/vm-detection
+    authors:
+      - richard.weiss@mandiant.com
+    description: Detects virtualization using VPCEXT (visual property container extender) instruction. Execution of this instruction will cause an illigal instruction exception outside of a virtual environment otherwise return 0
```
mr-tz, line 7:
```suggestion
    description: Detects virtualization using VPCEXT (visual property container extender) instruction. Execution of this instruction will cause an illegal instruction exception outside of a virtual environment otherwise return 0
```

mr-tz (changes requested):
one typo, otherwise looks good to me

mr-tz:
note that vivisect does not currently support `VPCEXT` so this only works in IDA

mr-tz (approved):

PR #639, opened 2022-12-14
new + updated .NET rules

mike-hunhoff:


host-interaction/console/manipulate-console.yml
```diff
@@ -12,13 +13,18 @@ rule:
...
-        - api: kernel32.SetConsoleCursorPosition
-        - api: kernel32.ReadConsoleOutputCharacter
-        - api: kernel32.WriteConsoleOutputCharacter
-        - api: kernel32.WriteConsoleOutput
-        - api: kernel32.WriteConsoleInput
-      - optional:
-        - api: kernel32.GetStdHandle
-        - number: 0xfffffff5 = STD_OUTPUT_HANDLE
+        - api: System.Console::SetWindowSize
+        - api: System.Console::SetWindowPosition
```
mr-tz, line 29: Do these capture the goal of this rule (I haven't checked the docs, but just based on their name)?

nursery/get-system-information-via-wmi-in-dotnet.yml
```diff
@@ -0,0 +1,20 @@
+rule:
+  meta:
+    name: get system information via WMI in .NET
```
mr-tz, line 3: WMI can do other things than just get system information. Adjust the name here?

nursery/write-to-console.yml
```diff
@@ -0,0 +1,13 @@
+rule:
+  meta:
+    name: write to console
```
mr-tz, line 3: Merge with the console rule above? The APIs seem to line up.

host-interaction/console/manipulate-console.yml
mike-hunhoff: `manipulate console`  is a fairly broad rule name that I'd argue includes the referenced APIs based on their MSDN descriptions:

```
Sets the height and width of the console window to the specified values.
Sets the position of the console window relative to the screen buffer.
```

Now if we tightened the rule name to something like `manipulate console output` the referenced APIs would no longer fit and we could instead split them to a new rule called something like `manipulate console window`.

I see the following options:
1. Leave as-is
2. Rename `manipulate console` to `manipulate console output` and create a second rule called `manipulate console window` that includes the referenced APIs

I favor option 2. Thoughts @mr-tz ?

nursery/get-system-information-via-wmi-in-dotnet.yml
mike-hunhoff: `access WMI data in .NET`?

nursery/write-to-console.yml
mike-hunhoff: Good point; I'll make this change pending response to my comments on the referenced `manipulate console` rule.

nursery/get-system-information-via-wmi-in-dotnet.yml
mr-tz: sounds good

host-interaction/console/manipulate-console.yml
mr-tz: 2. sounds good, for the first name I suggest `manipulate console buffer` (output and input).

mr-tz (approved):
lgtm, thanks!

Issue #642, opened 2022-12-21
Create rules releases


mr-tz:
Add releases of `capa-rules` to ease usability
- [x] update documentation (doc/rules.md in flare-capa), see https://github.com/mandiant/capa/pull/1245
  - [x] no more git, simply download ZIP
  - [x] capa code master and rules master guaranteed to work together, not necessarily true for other combinations
- [x] create historic rule releases
- [x] automatically generated via GitHub Actions, see #643 
- [x] update code (error outputs, etc.), see https://github.com/mandiant/capa/pull/1247
- [x] remove vX branches
- [x] ~~append to / link to releases in flare-capa~~ (tabling for now)

mr-tz:
@mike-hunhoff and @williballenthin do you agree that we can now remove the vX branches and tags, v1, v2, etc.?

williballenthin:
yup I agreeOn Jan 2, 2023, at 6:38 PM, Moritz ***@***.***> wrote:﻿
@mike-hunhoff and @williballenthin do you agree that we can now remove the vX branches and tags, v1, v2, etc.?

mr-tz:
Branches and tags have been removed.

PR #643, opened 2022-12-21
add release workflow

mr-tz:
supports #642 

In https://github.com/mandiant/capa/blob/master/.github/workflows/tag.yml we create a rules tag for every new code release.

This workflow then creates a respective release.

See this in Action in my fork:
- https://github.com/mr-tz/capa-rules/actions/runs/3749612283/jobs/6368283950
- result: https://github.com/mr-tz/capa-rules/releases/tag/v4.1.0

firefox_NISVtqqpLg
I find it pretty handy that we can just use the default ZIP and TGZ archives in the Assets section.

### DIFF
--- a/.github/scripts/create_releases.py
+++ b/.github/scripts/create_releases.py
@@ -1,0 +1,228 @@
+#!/usr/bin/env python3
+"""
+examples: %(prog)s v5.0.0 | %(prog)s -p v1.0.0 v2.0.0 | %(prog)s -c v3.0.0 -r test/capa-rules
+"""
+
+import os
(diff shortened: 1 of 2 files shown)


mr-tz:
We can discuss what the release content (body) should look like.

mr-tz:
See new test automatic test release based on test tag `v4.1.9`: https://github.com/mr-tz/capa-rules/releases/tag/v4.1.9
Action: https://github.com/mr-tz/capa-rules/actions/runs/3767077973/jobs/6404251738

PR #647, opened 2023-01-03
add "execute via .NET startup hook"

williballenthin:
closes #646

example file:
https://github.com/mandiant/capa-testfiles/pull/149

image

image

PR #649, opened 2023-01-03
tests: better detect rule name

williballenthin:
from https://github.com/mandiant/capa-rules/pull/647

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,7 +69,7 @@ jobs:
         cd rules/
         for changed_file in ${{ steps.files.outputs.added_modified }} ${{ steps.files.outputs.renamed }}; do
           if [[ ! $changed_file =~ .git|.md ]]; then
-            tag=$(grep 'name:' $changed_file | sed 's/^.*: //')
+            tag=$(grep '\sname:' $changed_file | sed 's/^.*: //')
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         cd rules/
         for changed_file in ${{ steps.files.outputs.added_modified }} ${{ steps.files.outputs.renamed }}; do
           if [[ ! $changed_file =~ .git|.md ]]; then
-            tag=$(grep 'name:' $changed_file | sed 's/^.*: //')
+            tag=$(grep '\sname:' $changed_file | sed 's/^.*: //')
```
williballenthin, line 72: im not sure if this needs to be escaped further. im inclined to test in prod, but what does the reviewer think?

williballenthin:
```suggestion
            tag=$(grep '\\sname:' $changed_file | sed 's/^.*: //')
```

not sure

mr-tz (approved):
good catch

.github/workflows/tests.yml
williballenthin: i dont think further quoting is needed here

https://yaml-multiline.info/

PR #654, opened 2023-01-05
Adjusted rules for the ELF BEACON sample shared by Intezer

mr-tz:
Originally proposed by @ramen0x3f in #463.

WindowsTerminal_3tuSnG54QI


mr-tz:
```
rules with FAIL:
  - get system time on Linux
```

well, this is a fun one... sometimes it works, sometimes it doesn't... (sometimes vivisect emits the api, sometimes it doesn't) 😕  

removing it for now, as the rule's value is small, but opening an issue

williballenthin (approved):

PR #655, opened 2023-01-06
improve rules that often match

mr-tz:
I've analyzed the top 50 or so rule matches over 5000 recent random samples.
In these rules I noticed room for improvement. Happy to discuss as needed.


executable/resource/extract-resource-via-kernel32-functions.yml
```diff
@@ -28,4 +28,3 @@ rule:
             - api: LdrFindResourceEx_U
           - api: kernel32.SizeofResource
           - api: kernel32.FreeResource
-      - api: user32.LoadString
```
mr-tz, line 31: I don't think that's really what we are after normally. Anyone thing it's worth creating a new rule for it?

host-interaction/process/terminate/terminate-process-via-fastfail.yml
```diff
@@ -1,17 +0,0 @@
-rule:
```
mr-tz, line 1: ref: #143, I've only seen this hit in library code (which Python flirt often doesn't recognize)

host-interaction/registry/create-or-open-registry-key.yml
```diff
@@ -1,10 +1,10 @@
 rule:
   meta:
     name: create or open registry key
```
mr-tz, line 3: analogous to `create or open file`, we have a lot more specific rules that use this (now library) rule, e.g. `set registry value` or `delete registry value`

executable/resource/extract-resource-via-kernel32-functions.yml
williballenthin: agree that this should be removed

lib/calculate-modulo-256-via-x86-assembly.yml
```diff
@@ -15,5 +15,5 @@ rule:
     - and:
       - mnemonic: and
       - or:
-        - number: 0x800000FF
-        - number: 0xFF
+        - operand[1].number: 0x800000FF
+        - operand[1].number: 0xFF
```
williballenthin, line 19: can you explain this change?

williballenthin (approved):

lib/calculate-modulo-256-via-x86-assembly.yml
mr-tz: trying to be more explicit here, but yeah, there can only be an immediate in the second operand, so I guess this is not useful here

mr-tz: any preference? be explicit here or undo change and stick to `number`?

mr-tz, line 19:
```suggestion
        - operand: 0x800000FF
        - operand: 0xFF
```

lib/calculate-modulo-256-via-x86-assembly.yml
```diff
@@ -15,5 +15,5 @@ rule:
     - and:
       - mnemonic: and
       - or:
-        - number: 0x800000FF
-        - number: 0xFF
+        - operand: 0x800000FF
+        - operand: 0xFF
```
mr-tz, line 19:
```suggestion
        - number: 0x800000FF
        - number: 0xFF
```

host-interaction/file-system/exists/check-if-file-exists.yml
```diff
@@ -17,12 +17,12 @@ rule:
           - api: kernel32.GetFileAttributes
           - instruction:
             - mnemonic: cmp
-            - number: 0xFFFFFFFF = INVALID_FILE_ATTRIBUTES
+            - operand[1].number: 0xFFFFFFFF = INVALID_FILE_ATTRIBUTES
```
mr-tz, line 20:
```suggestion
            - number: 0xFFFFFFFF = INVALID_FILE_ATTRIBUTES
```

williballenthin (approved):

Issue #632, opened 2022-11-01
Findcrypt constants


mr-tz:
See https://github.com/JusticeRage/Manalyze/blob/3ec7aed7f0f10108fb2f92612400bdf21f94e8bb/bin/yara_rules/findcrypt.yara

williballenthin:
- [ ] CRC32
 - [ ]  MD%
 - [ ]  SHA1
 - [ ]  SHA256
 - [ ]  SHA512
 - [ ]  Whirlpool
 - [ ]  Tiger
 - [ ]  AES
 - [ ]  Blowfish
 - [ ]  DES
 - [ ] RC5/6
 - [ ]  Camellia
 - [ ]  Twofish
 - [ ]  TEA
 - [ ]  Base58
 - [ ]  Known Diffie-Hellman primes

williballenthin:
interesting that some of the "constants" are actually these:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/156560/211332126-eb8d88a2-3c41-489e-ac3d-27cbb598740c.png">

<img width="423" alt="image" src="https://user-images.githubusercontent.com/156560/211332028-2b0123e2-c5c3-4a37-bcbb-73ef67889f53.png">

williballenthin:
the remaining rules are:
 - DES: we have an existing DES rule and im not confident in my understanding of either to update it.
 - Known Diffie-Hellman primes: i dont know where these values come from so would discourage including them verbatim

otherwise, we're good!

mr-tz:
awesome, thank you!

PR #657, opened 2023-01-09
more SHA1 constants

williballenthin:
ref #632


williballenthin:
notably this now identifies SHA1 in the new example, whereas it didnt match before, so this is net new coverage

mr-tz (approved):

PR #659, opened 2023-01-11
Update encrypt-data-using-curve25519 rule

mwilliams31:
- Added description that documents targeted code
- Added description for each instruction
- Added an additional number often used to clear only bit 7 of the last byte of the secret key


mr-tz (approved):
neat, thank you!!

PR #661, opened 2023-01-13
add upper limit for basic block count

### LINKED ISSUE #435: FPs encrypt-data-using-rc4-prga.yml
As discussed here https://github.com/fireeye/capa/pull/679#issuecomment-884306100
Followed by @williballenthin's comment https://github.com/fireeye/capa/pull/679#issuecomment-884332188

mr-tz:
closes #435

Library ID fails and RC4 PRGA matches on `___strgtold12_l`. An upped BB count prevents this.

Are there concerns on the upper limit value here?

### DIFF
--- a/data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
+++ b/data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
@@ -25,7 +25,7 @@ rule:
         - count(mnemonic(movzx)): 4 or more
       # should not call (many) functions
       - count(characteristic(calls from)): (0, 4)
-      # should not be too simple or too complex
+      # should not be too simple or too complex (50 is picked by intuition)
       - count(basic blocks): (4, 50)
       - match: contain loop
       - optional:
--- a/data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
+++ b/data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
@@ -25,8 +25,8 @@ rule:
         - count(mnemonic(movzx)): 4 or more
       # should not call (many) functions
       - count(characteristic(calls from)): (0, 4)
-      # should not be too simple
-      - count(basic blocks): 4 or more
+      # should not be too simple or too complex
+      - count(basic blocks): (4, 50)
       - match: contain loop
       - optional:
         - or:


data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
```diff
@@ -25,8 +25,8 @@ rule:
         - count(mnemonic(movzx)): 4 or more
       # should not call (many) functions
       - count(characteristic(calls from)): (0, 4)
-      # should not be too simple
-      - count(basic blocks): 4 or more
+      # should not be too simple or too complex
+      - count(basic blocks): (4, 50)
```
williballenthin, line 29: how did we pick 50? i agree we should use an upper limit, but im leaning towards keeping it as low as possible, to keep the rule tight. thoughts?

williballenthin: yeah thats a good point. i recall that our RC4 rules sometimes FP, so i wonder about the tradeoff of catching the inlined routines versus matching too broadly. do you have a sense for this?

anyways, not a blocker here, but a discussion point.

data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
```diff
@@ -25,8 +25,8 @@ rule:
         - count(mnemonic(movzx)): 4 or more
       # should not call (many) functions
       - count(characteristic(calls from)): (0, 4)
-      # should not be too simple
-      - count(basic blocks): 4 or more
+      # should not be too simple or too complex
```
williballenthin, line 28:
```suggestion
      # should not be too simple or too complex (50 is picked by intuition)
```

data-manipulation/encryption/rc4/encrypt-data-using-rc4-prga.yml
mr-tz: Not really, the only rough data I have is that out of 5000 random samples `encrypt data using RC4 PRGA` hit on 834 files (14%) which seems a lot. Manually inspecting 3 samples shows 1 valid hit and 2 library function hits (in IDA) that can be filtered out using the upper BB limit.

PR #663, opened 2023-01-17
52 Weeks of capa Rules - 2

mike-hunhoff:
adding/updating rules for `.NET`.


communication/tcp/serve/start-tcp-server.yml
```diff
@@ -4,15 +4,24 @@ rule:
...
+        - or:
+          - api: accept
+          - api: WSAAccept
+      - api: System.Net.Sockets.TcpListener::Start
+      - api: System.Net.Sockets.TcpListener::AcceptTcpClient
+      - api: System.Net.Sockets.TcpListener::BeginAcceptTcpClient
+      - api: System.Net.Sockets.TcpListener::AcceptTcpClientAsync
+      - api: System.Net.Sockets.TcpListener::AcceptSocket
+      - api: System.Net.Sockets.TcpListener::BeginAcceptSocket
+      - api: System.Net.Sockets.TcpListener::AcceptSocketAsync
```
mr-tz, line 27:
```suggestion
      - api: System.Net.Sockets.TcpListener::AcceptSocketAsync

```

nursery/decode-data-using-url-encoding.yml
```diff
@@ -0,0 +1,15 @@
+rule:
+  meta:
+    name: decode data using URL encoding
```
mr-tz, line 3: `decode URL-encoded data`?

nursery/manipulate-user-privileges.yml
```diff
@@ -0,0 +1,10 @@
+rule:
+  meta:
+    name: manipulate user privileges
```
mr-tz, line 3: add this to `modify access privileges`?

nursery/decode-data-using-url-encoding.yml
mike-hunhoff: This was my initial choice but it does not adhere to the naming scheme we have been using e.g.
- `decode data using Base64 in .NET`
- `decrypt data using RSA`
- `decrypt data via SSPI`

Happy to make the suggested change pending additional thoughts

nursery/manipulate-user-privileges.yml
mike-hunhoff: `modify access privileges` appears specific to a process whereas `LsaAddAccountRights` is specific to a user which is why I kept them separate.  Thoughts?

nursery/decode-data-using-url-encoding.yml
mr-tz: Hm, right, I'm fine leaving as is then.

nursery/manipulate-user-privileges.yml
mr-tz: ah, right, never mind

nursery/manipulate-user-privileges.yml
```diff
@@ -0,0 +1,10 @@
...
+    namespace: communication/authentication
```
mr-tz, line 4:
```suggestion
    namespace: host-interaction/user
```

PR #665, opened 2023-01-19
add whirlpool hash

williballenthin:
ref #632 

viv doesn't identify the example function, unfortunately, due to it being referenced via a chain of two pointers.

### DIFF
--- a/nursery/hash-data-using-whirlpool.yml
+++ b/nursery/hash-data-using-whirlpool.yml
@@ -1,0 +1,22 @@
+rule:
+  meta:
+    name: hash data using Whirlpool
+    namespace: data-manipulation/hashing/whirlpool
+    authors:
+      - william.ballenthin@mandiant.com
+    scope: function
+    references:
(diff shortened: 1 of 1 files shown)


mr-tz (approved):

nursery/hash-data-using-whirlpool.yml
```diff
@@ -0,0 +1,26 @@
...
+  meta:
+    name: hash data using Whirlpool
+    namespace: data-manipulation/hashing/whirlpool
+    authors:
+      - william.ballenthin@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
+    mbc:
+      - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
```
mr-tz, line 11: Is `Encryption-Standard Algorithm` correct here?

williballenthin: good catch

mr-tz (approved):

PR #667, opened 2023-01-19
add TEA encryption

williballenthin:
ref #632


nursery/encrypt-data-using-tea.yml
```diff
@@ -0,0 +1,17 @@
...
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
+    mbc:
+      - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
+    references:
+      - https://en.wikipedia.org/wiki/Tiny_Encryption_Algorithm
+  features:
+    - and:
+      - number: 0x9E3779B9 = delta
+      - number: 0xC6EF3720 = sum
```
mr-tz, line 17: I see the delta in the reference, but not the sum value. Will the sum number always be present?

mr-tz (approved):

PR #623, opened 2022-09-16
Rule for lzo compression

edeca:
Work in progress rule for LZO encryption. Found in a sample which requires manual unpacking, therefore looking for alternate samples.

Further work needed before merging.

PR #677, opened 2023-01-30
LZO compression / decompression rules continued

mr-tz:
addressing feedback added to #623

supersedes #623

### DIFF
--- a/data-manipulation/compression/compress-data-using-lzo.yml
+++ b/data-manipulation/compression/compress-data-using-lzo.yml
@@ -14,12 +14,12 @@ rule:
     examples:
       - ee3b869b668abec332d07c66d1a39f6dbf3a598cc1325b57a0504f8d24ac2e28.dll_:0x10026370
   features:
-    - or:
-      - and:
-        - description: LZO compress
-        - number: 0x4000
-        - number: 0x3FFF
+    - and:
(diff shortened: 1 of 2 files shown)


data-manipulation/compression/compress-data-using-lzo.yml
```diff
@@ -0,0 +1,25 @@
...
+  features:
+    - and:
+      - number: 0x4000
+      - number: 0x3FFF
+      - or:
+        - number: 0xBFFE
+        - number: 0xBFFF
+      - number: 0x800
+      - number: 0x7FF
+      - number: 0x201F
```
williballenthin, line 25: if this FPs then perhaps we can restrict the bb count to something reasonable.

williballenthin: i assume there's also a loop, but im not sure if thats important to add.

williballenthin (approved):

data-manipulation/compression/compress-data-using-lzo.yml
mr-tz: Correct, my hope's that the combination of these numbers don't show up too often, but it's hard to test without [#test](https://github.com/mandiant/capa/issues/424) or a similar mechanism.

PR #680, opened 2023-02-01
add create tcp socket via raw AFD driver

### LINKED ISSUE #537: communicate directly with afd.sys to create TCP sockets
Reference/example: https://twitter.com/x86matthew/status/1498751350632497152?s=21

williballenthin:
closes #537

https://github.com/mandiant/capa-testfiles/pull/153


williballenthin:
tbh i dont expect to match that often/ever, but its an interesting technique and a good exercise to translate into a capa rule.

mr-tz (approved):
neat rule indeed!

Issue #679, opened 2023-02-01
address rule tweaks in CallSignaturesPlugin


williballenthin:
master's thesis https://www.ru.nl/publish/pages/769526/joren_vrancken.pdf by @joren485
page 72 has some suggestions, like:
<img width="372" alt="image" src="https://user-images.githubusercontent.com/156560/216013795-26b769b7-a1db-4d53-b2a4-bc58012b35a5.png">

code here: https://github.com/joren485/CallSignaturesPlugin
signatures: https://github.com/joren485/CallSignaturesPlugin/tree/main/CallSignaturesPlugin/signatures

joren485:
First of all, thanks for your interest in my thesis!

I created three PRs with some potential improvements.

Some other points that are harder to address:
* Detecting indirect calls: You will most likely need mandiant/capa#771 for this.
* False positives with common constants (for example, the `SHGetFolderPath` call in 9.2.4 of my thesis): Again, you will need something like mandiant/capa#771 to detect the actual position of the argument. 
* Capa does not catch some samples that use command line programs (e.g. `reg.exe` and `sc.exe`), because `host-interaction/process/create` is not always detected (e.g. because `ShellExecute` is called indirectly). An easy solution is to remove the requirement of `host-interaction/process/create`, however this will also increase the number of false positives.

joren485:
@williballenthin I believe I created PRs for all rule tweaks. Can this be closed or do you want/see other improvements?

PR #685, opened 2023-02-05
Add T1053.002 to schedule task via command line rule

joren485:
Part of #679

This PR adds detection of creating scheduled tasks via the [at.exe command](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/at).


williballenthin (approved):
awesome, thanks @joren485!

persistence/scheduled-tasks/schedule-task-via-command-line.yml
```diff
@@ -6,14 +6,22 @@ rule:
       - 0x534a@mailbox.org
     scope: function
     att&ck:
+      - Persistence::Scheduled Task/Job::At [T1053.002]
       - Persistence::Scheduled Task/Job::Scheduled Task [T1053.005]
```
williballenthin, line 10: i wonder if we should split these rules out so that the techniques don't get mixed (e.g. don't want to tag "AT job" if at wasn't used).

mr-tz:
Hey @joren485, do you have a moment to split up the rule or do you want me to help/propose a change?

joren485:
@mr-tz Yeah, no worries. I will create a separate rule when I find some time this week.

persistence/scheduled-tasks/schedule-task-via-schtasks.yml
```diff
@@ -1,6 +1,6 @@
 rule:
   meta:
-    name: schedule task via command line
+    name: schedule task via schtasks
```
mr-tz, line 3: should we leave this as `command line` as we also include `Register-ScheduledTask`?

williballenthin: im not sure, since then "at.exe" would fall back into this rule. i think this is probably fine as is, unless we run into problems or confusion?

mr-tz: yeah, so we're good to merge?

mr-tz:
I think this is good to go, but let's wait till after the v5 release with merging.

mr-tz:
Thanks a lot @joren485!

PR #686, opened 2023-02-05
Split the 'reg.exe add' check into multiple strings

joren485:
Part of #679 

This PR improves the detection of the `reg.exe add` command. Some samples split the command up into multiple parts. For example, a sample might pass `reg.exe` as one argument and `ADD HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\ ...` as another.


mr-tz (approved):
looks good, thank you!

mike-hunhoff (approved):

PR #687, opened 2023-02-05
Add checks for creating a service by writing to the registry

joren485:
Part of #679 

This PR adds checks for samples that create a service by directly editing the registry.


williballenthin (approved):
nice, thanks @joren485!

PR #691, opened 2023-02-06
Rule for reference file/folder stream path on Windows

### LINKED ISSUE #531: reference file/folder stream path on Windows
Can we develop a sane regex to detect file/folder stream paths e.g. `c:\windows\ime:myfile.txt`. Best case we can simply look for `:` in the file/folder name of a given path (on Windows).

MS reference: https://docs.microsoft.com/en-us/windows/win32/fileio/file-streams

bkojusner:
This rule passes the thorough linting test and closes #531


host-interaction/file-system/reference-absolute-stream-path-windows.yml
```diff
@@ -0,0 +1,28 @@
+rule:
+  meta:
+    name: reference absolute stream path Windows
```
williballenthin, line 3:
```suggestion
    name: reference absolute stream path on Windows
```

host-interaction/file-system/reference-absolute-stream-path-windows.yml
```diff
@@ -0,0 +1,28 @@
...
+    namespace: host-interaction/file-system
+    authors:
+      - blas.kojusner@mandiant.com
+    scope: basic block
```
williballenthin, line 7:
```suggestion
    scope: function
```

im not quite sure if this should be file, function, or basic block. i dont think we have guidance. im inclined to suggest "function" since analysts operate at this level more often.

thoughts @mr-tz or @mike-hunhoff ?

williballenthin (changes requested):
recommend changing the rule name, then happy to merge!

host-interaction/file-system/reference-absolute-stream-path-windows.yml
mr-tz: I think original guidance was to keep it to the tighter scope, but BB or function here are fine with me.

williballenthin: keeping this as BB works for me, thanks @mr-tz

PR #699, opened 2023-02-13
adding "packaged as single-file .NET application"

### LINKED ISSUE #636: detect .NET single bundle applications
> Bundling all application-dependent files into a single binary provides an application developer with the attractive option to deploy and distribute the application as a single file.

https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli

mike-hunhoff:
closes #636

### DIFF
--- a/nursery/packaged-as-single-file-dotnet-application.yml
+++ b/nursery/packaged-as-single-file-dotnet-application.yml
@@ -1,0 +1,14 @@
+rule:
+  meta:
+    name: packaged as single-file .NET application
+    namespace: executable/installer/dotnet
+    authors:
+      - michael.hunhoff@mandiant.com
+    scope: file
+    description: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli
+  features:
+    - and:
+      - string: "Failure processing application bundle; possible file corruption."
+      - string: "Failure processing application bundle."
+      - string: "Bundle header version compatibility check failed"
+      - string: "A fatal error was encountered. Could not extract contents of the bundle"
--- a/nursery/packaged-as-single-file-dotnet-application.yml
+++ b/nursery/packaged-as-single-file-dotnet-application.yml
@@ -4,8 +4,9 @@ rule:
     namespace: executable/installer/dotnet
     authors:
       - michael.hunhoff@mandiant.com
-    description: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli
+    description: Single binary containing target .NET application and all application-dependent files
     scope: file
+    reference: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli
   features:
     - and:
       # the installer portion of a single-file application is native code (in the samples I have seen) so format(dotnet) does not apply here.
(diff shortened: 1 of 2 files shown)


mr-tz (approved):
Neat, would `format: dotnet` or `match: compiled to the .NET platform` make sense here also or are these not dotnet files?

nursery/packaged-as-single-file-dotnet-application.yml
```diff
@@ -0,0 +1,14 @@
+rule:
+  meta:
+    name: packaged as single-file .NET application
+    namespace: executable/installer/dotnet
+    authors:
+      - michael.hunhoff@mandiant.com
+    description: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli
+    scope: file
+  features:
+    - and:
```
williballenthin, line 10:
```suggestion
    - and:
      # the installer portion of a single-file application is native code (in the samples I have seen) so format(dotnet) does not apply here.
```

nursery/packaged-as-single-file-dotnet-application.yml
```diff
@@ -0,0 +1,14 @@
...
+      - string: "Failure processing application bundle; possible file corruption."
+      - string: "Failure processing application bundle."
+      - string: "Bundle header version compatibility check failed"
+      - string: "A fatal error was encountered. Could not extract contents of the bundle"
```
williballenthin, line 14: did you have a chance to look at VT for these strings? if not, i can do it tomorrow and try to find a good example.

williballenthin (approved):

nursery/packaged-as-single-file-dotnet-application.yml
```diff
@@ -0,0 +1,14 @@
+rule:
+  meta:
+    name: packaged as single-file .NET application
+    namespace: executable/installer/dotnet
```
williballenthin, line 4: thoughts on putting this in `compiler/` next to the other packagers like py2exe, etc?

https://github.com/mandiant/capa-rules/tree/2586016cf2fb92298fb6ac462d88c12c91b76e73/compiler

williballenthin: maybe "compiler" isn't the best term there, but thats a different discussion.

mike-hunhoff: Hmmm I think it fits better under installer because the .NET code isn't, itself, compiled. It's wrapped by a native bootstrap program that writes all of the necessary files to disk before executing the target assembly. But open to further discussion.

nursery/packaged-as-single-file-dotnet-application.yml
```diff
@@ -0,0 +1,14 @@
...
+    authors:
+      - michael.hunhoff@mandiant.com
+    description: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli
```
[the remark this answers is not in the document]
mike-hunhoff: good catch; I've moved the URL to a reference and added a separate description

nursery/packaged-as-single-file-dotnet-application.yml
mike-hunhoff: see https://github.com/mandiant/capa-testfiles/pull/164

mike-hunhoff:
need to merge https://github.com/mandiant/capa-testfiles/pull/164

Issue #676, opened 2023-01-30
Coverage Magic Numbers in Cryptography


mr-tz:
Magic Numbers in Cryptography
see https://rst.im/magic/

PR #700, opened 2023-02-14
Adding rules for TEA, XTEA, XXTEA and RIPEMD hashing 

raymondlleong:
Adding rules for XTEA, XXTEA, RIPEMD-128, 160, 256, 320
Adding to current rules for TEA encrypt & decrypt

Relates to issue #676


nursery/hash-data-using-ripemd128.yml
```diff
@@ -0,0 +1,26 @@
...
+          - number: 0x98BADCFE = also used in MD5, RIPEMD-128, RIPEMD-256, and RIPEMD-320
+          - number: 0x10325476 = also used in MD5, RIPEMD-128, RIPEMD-256, and RIPEMD-320
+        - bytes: 01 23 45 67 89 AB CD EF FE DC BA 98 76 54 32 10
+
+      - not:
+        - or:
+          - number: 0x76543210 = RIPEMD-256 and RIPEMD-320 constant
+          - number: 0xFEDCBA98 = RIPEMD-256 and RIPEMD-320 constant
+          - number: 0xC3D2E1F0 = RIPEMD-160 and RIPEMD-320 constant
+          - number: 0x3C2D1E0F = RIPEMD-320 constant
```
williballenthin, line 26: is there a way to exclude MD5 here?

nursery/decrypt-data-using-tea.yml
```diff
@@ -4,14 +4,43 @@ rule:
...
     att&ck:
       - Defense Evasion::Obfuscated Files or Information [T1027]
     mbc:
       - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
     references:
       - https://en.wikipedia.org/wiki/Tiny_Encryption_Algorithm
+    examples:
+      - 5B99FA01C72CEBC53A76CC72E9581189:0x180001027
+      - 7FD2E2E3C88675D877190ABAA3002B55:0x010004EC0
+      - 775B26D6ACCB5A4AE868983F8F5006A3:0x010018890
```
williballenthin, line 18: now that we have examples, let's move this rule out of the nursery.

williballenthin: in fact, all rules in this PR with examples we can move out of the nursery, awesome!

nursery/decrypt-data-using-tea.yml
```diff
@@ -4,14 +4,43 @@ rule:
...
+            - operand[1].number: 0x5
+          - instruction:
+            - mnemonic: shl
+            - operand[1].number: 0x4
+          - count(mnemonic(shr)): 0x2
+          - characteristic: nzxor
+      - or:
+        - operand[1].number: 0x9E3779B9 = key schedule constant
+        - operand[1].number: 0x61C88647 = key schedule constant two's complement
+      - operand[1].number: 0xC6EF3720 = sum
```
williballenthin, line 34: neat logic!

nursery/hash-data-using-ripemd128.yml
williballenthin: i don't totally understand all the cases here and if you've covered them, but i trust you to have done that work :-)

nursery/hash-data-using-ripemd256.yml
```diff
@@ -0,0 +1,27 @@
+rule:
+  meta:
+    name: hash data using RIPEMD256
+    namespace: data-manipulation/hashing/ripemd256
+    authors:
+      - raymond.leong@mandiant.com
+    scope: function
+    references:
+      - https://en.wikipedia.org/wiki/RIPEMD-256
+  features:
```
williballenthin, line 10: if no examples, please keep in the nursery

williballenthin (changes requested):
this is great! thank you for taking the time to write these rules and find so many examples.

would you please remove the `.DS_Store` files? also, it looks like some file permissions changed. we don't really use those, so i'm not sure if it matters? not a problem if its not obvious how to fix that.

with regards to the testfiles PR, as long as we have a good example for each rule, i think there are diminishing returns for adding more, especially if the logic is basically the same. so for that large file, if the logic is well covered by the other example, would you please remove it?

thanks!

raymondlleong:
Thank you for the review! A follow up commit addresses these comments above:
- rules with examples are moved out of nursery 
- unneeded examples removed from rules and their example files removed from the PR for capa-testfiles
- DS_store files deleted

nursery/hash-data-using-ripemd128.yml
raymondlleong: Good point and added an additional function scope for an array constant not present in MD5. Would have liked to keep it a function scoped rule but the magic constants are sometimes in a function that solely moves their values into a structure with nothing else to signature off:
Screen Shot 2023-02-14 at 11 47 18 PM

raymondlleong: The RIPEMD160 rule is having a similar issue with SHA1. The magic constants listed in https://rst.im/magic/RIPEMD160 are also present in SHA1, including the K constants even though they arent listed in the corresponding SHA1 page. The routines have a lot of similarities and more research is needed to write a rule that better deconflicts the two. I think keeping the RIPEMD160 rule out of the repo may be better for now to avoid false positives. 

Of note, the logic in the SHA1 rule below will also hit on RIPEMD initialization but with SHA1 being much more common than the other two may be less of an issue:
```
- or:
      - and:
        - description: Magic initialization constants used in SHA1
        - number: 0x67452301 = A, also used by MD5
        - number: 0xEFCDAB89 = B, also used by MD5
        - number: 0x98BADCFE = C, also used by MD5
        - number: 0x10325476 = D, also used by MD5
        - number: 0xC3D2E1F0 = specific to SHA1, not MD4 nor MD5
```

data-manipulation/hashing/ripemd160/hash-data-using-ripemd160.yml
```diff
@@ -0,0 +1,27 @@
...
+    references:
+      - https://en.wikipedia.org/wiki/RIPEMD-160
+    examples:
+      - 35F9CFE5110471A82E330D904C97466A:0x10009340
+  features:
+    - and:
+      - or:
+        - and:
+          - description: Magic initialization constants used in RIPEMD-160
+          - number: 0x67452301 = also used in MD5, SHA1, RIPEMD-128, RIPEMD-256, and RIPEMD-320
```
mr-tz, line 17: I like that these inline descriptions can point analysts to the alternative algorithms!

nursery/hash-data-using-ripemd128.yml
williballenthin: awesome, thanks for the explanation. it seems a little tricky to get perfectly right, but i tend to agree with you that we should focus on the mostly likely scenarios and being helpful to most analysts. maybe disabling one of the rules and adding extra documentation the other is best?

raymondlleong: Thanks for the suggestion. Updated the documentation in the MD5 and SHA1 rules to highlight the shared constants with the RIPEMD routines and removed the RIPEMD160 rule that would be prone to false positives with SHA1 (didnt see a way to disable it).

mr-tz (approved):
LGTM, thanks!

mr-tz:
If @williballenthin doesn't object in the next few days, I'll merge this. Thanks, everyone!

williballenthin:
send it!

mr-tz:
Thanks, Willi and Ray!

PR #714, opened 2023-02-26
add get http query uri

williballenthin:
ref #3

please LINT_THOROUGH

### DIFF
--- a/communication/http/get-http-query-uri.yml
+++ b/communication/http/get-http-query-uri.yml
@@ -12,4 +12,4 @@ rule:
   features:
     - and:
       - api: wininet.HttpQueryInfo
-      - number: 0x13 = HTTP_QUERY_URI
+      - number: 0xD = HTTP_QUERY_URI
--- a/communication/http/get-http-query-uri.yml
+++ b/communication/http/get-http-query-uri.yml
@@ -1,0 +1,15 @@
+rule:
+  meta:
+    name: get HTTP query URI
+    namespace: communication/http
+    authors:
+      - william.ballenthin@mandiant.com
+    scope: basic block
+    mbc:
+      - Communication::HTTP Communication [C0002]
+    examples:
+      - Practical Malware Analysis Lab 03-02.dll_:0x10004452
+  features:
+    - and:
+      - api: wininet.HttpQueryInfo
+      - number: 0x13 = HTTP_QUERY_URI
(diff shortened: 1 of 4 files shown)


communication/http/get-http-query-uri.yml
```diff
@@ -0,0 +1,15 @@
...
+      - william.ballenthin@mandiant.com
+    scope: basic block
+    mbc:
+      - Communication::HTTP Communication [C0002]
+    examples:
+      - Practical Malware Analysis Lab 03-02.dll_:0x10004452
+  features:
+    - and:
+      - api: wininet.HttpQueryInfo
+      - number: 0x13 = HTTP_QUERY_URI
```
mr-tz, line 15: Should this be 0xD (decimal 13) via https://learn.microsoft.com/en-us/windows/win32/wininet/query-info-flags?

communication/http/get-http-query-uri.yml
```diff
@@ -0,0 +1,15 @@
+rule:
+  meta:
+    name: get HTTP query URI
```
mr-tz, line 3: I'm not sure about "query"... How about request or header?

communication/http/get-http-query-uri.yml
williballenthin: yes, definitely.

1) 🤦🏼
2) should have run with `lint --thorough` locally, but forgot
3) why didn't CI catch this with `--thorough`?

communication/http/get-http-query-uri.yml
williballenthin: tbh, i misunderstood what this `HttpQueryInfo` option fetched. i thought it was the query parameters portion of the URL, but i see now its the whole URL. so i'd propose:

`get HTTP request URL`

communication/http/get-http-query-uri.yml
williballenthin:
```suggestion
      - number: 0xD = HTTP_QUERY_URI
```

williballenthin: ah i need `LINT_THOROUGH` in the PR text.

mr-tz:
Hm, the linter should have run thoroughly before. Maybe there's a 13 in the block as well?

mr-tz (approved):
😃 cool

PR #728, opened 2023-03-20
updated getusername rule to include getcomputername for malware that checks hostname

emtuls:


williballenthin:
i like the idea of adding this detection. would you please add it as a new rule, though?

williballenthin (changes requested):
please submit logic as a new rule: `check for sandbox hostname`

anti-analysis/anti-vm/vm-detection/check-for-sandbox-username-hostname.yml
```diff
@@ -15,7 +16,9 @@ rule:
       - ccbf7cba35bab56563c0fbe4237fdc41:0x402B90
   features:
     - and:
-      - api: GetUserName
+      - or:
+        - api: GetUserName
+        - match: get hostname
```
williballenthin, line 21: add comment here explaining that username often overlaps with hostname in many of these checks

williballenthin (approved):

anti-analysis/anti-vm/vm-detection/check-for-sandbox-username-hostname.yml
```diff
@@ -15,7 +16,9 @@ rule:
       - ccbf7cba35bab56563c0fbe4237fdc41:0x402B90
   features:
     - and:
-      - api: GetUserName
+      - or:
+        - api: GetUserName
```
mike-hunhoff, line 20: substitute match for https://github.com/mandiant/capa-rules/blob/master/host-interaction/session/get-session-user-name.yml?

emtuls:
Updated as per both suggestions.

williballenthin (approved):

mr-tz:
please rename:
```
FAIL: filename doesn't match the rule name: Rename rule file to match the rule name, expected: "check-for-sandbox-username-or-hostname.yml", found: "check-for-sandbox-username-hostname.yml"
```

anti-analysis/anti-vm/vm-detection/check-for-sandbox-username-hostname.yml
```diff
@@ -37,33 +41,37 @@ rule:
...
           description: Trickbot Downloader Username Check
-        - string: /WIN7\-TRAPS/i
+        - string: /WIN7-TRAPS/i
           description: Trickbot Downloader Username Check
         - string: /WDAGUtilityAccount/i
           description: Windows Defender Application Guard
+        - string: /HAL9TH/i
+          description: Windows Defender Emulator
+        - string: /REM/i
+          description: SANS REM VM
```
mr-tz, line 77:
```suggestion
          description: SANS REM VM

```

mr-tz, line 77: strings with less than 4 characters won't be extracted, can you add a character or remove this entry?

emtuls: Removed.

mr-tz (approved):
thanks!

PR #730, opened 2023-03-21
adding new method for windows sandbox detection used by QBOT

emtuls:


williballenthin (approved):
logic looks good; lets see how CI/linter goes

anti-analysis/anti-vm/vm-detection/check-for-windows-sandbox-via-subdirectory.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "echernofsky@google.com"
+    scope: basic block
+    att&ck:
+      - Defense Evasion::Virtualization/Sandbox Evasion::System Checks [T1497.001]
+    mbc:
+      - Anti-Behavioral Analysis::Virtual Machine Detection [B0009]
+    references:
+      - https://www.elastic.co/security-labs/qbot-malware-analysis
+    examples:
+      - ffaf9cd085434fd5230511ab895ba494:0x10015A6A
```
mr-tz, line 15: if this is the according testfile, let's change to use sha256 sum here
https://github.com/mandiant/capa-testfiles/pull/166

mr-tz:
Lint on the sample seems to fail: `FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example`. Does it work locally?

mr-tz (approved):
thanks!

PR #733, opened 2023-03-24
updated "check for debugger via API" with more techniques

emtuls:


emtuls:
I noticed that a similar rule exists for one of the numbers in the API that I added to this (ProcessDebugPort): https://github.com/mandiant/capa-rules/blob/master/anti-analysis/anti-debugging/debugger-detection/check-processdebugport.yml

With that being said, should I add the other two numbers/methods as their own rules as well? Not sure what level of granularity we should be going to.

emtuls:
Ah. I see that these are already in the nursery. My mistake.

PR #751, opened 2023-04-26
Update reference-absolute-stream-path-on-windows.yml

bkojusner:
Resolves FP cases where `C:\\windows\\system32\\kerne132.dll` and `\\\\\\\\\\\\\\server\\share\\file:stream` are detected. Updated comments and regex to reflect this. Made additional modifications to regex for performance enhancement.


williballenthin (approved):
thank you!

PR #761, opened 2023-05-19
remove very common .rsrc section rule

mr-tz:
related to #7


executable/resource/extract-resource-via-kernel32-functions.yml
```diff
@@ -18,7 +18,7 @@ rule:
           - api: kernel32.LockResource
           - api: LdrAccessResource
         - optional:
-          - match: contain a resource (.rsrc) section
+          - section: .rsrc
```
mr-tz, line 21: upps, we cannot do that here

mr-tz, line 21:
```suggestion
```

williballenthin (approved):

PR #762, opened 2023-05-23
Create covertly decode and write data to Windows directory using indirect calls

dkelly2e:


data-manipulation/encoding/xor/covertly decode and write data to Windows directory using indirect calls.txt
mr-tz, line 1: the filename should use `-` instead of space and end with `.yml`

dkelly2e:
Updated! Thanks for the clarification

williballenthin (approved):

mr-tz:
rules with FAIL:
  - covertly decode and write data to Windows directory using indirect calls
  - detect VM via disk hardware WMI queries
  - detect VM via motherboard hardware WMI queries

PR #763, opened 2023-05-26
Improve clipboard rules

### LINKED ISSUE #293: Improvements for clipboard rules
I ran into a FP for `read-clipboard-data` that sparked some ideas for improvement. The FP was in a sample that contained a bunch of API calls whose results weren't used. I don't have time to write/test new rules so I'm documenting my research efforts here in case someone wants to run with it.

Here's some documentation that could be inserted as references for multiple rules:
* https://docs.microsoft.com/en-us/windows/win32/dataxchg/using-the-clipboard

In the documentation you'll notice the following APIs are frequently used:
```
kernel32.GlobalAlloc
kernel32.GlobalLock
kernel32.GlobalUnlock
```
These could be incorporated as `optional` features for some clipboard rules. They are present in the existing rules' examples. If they show up often enough, maybe they get moved from `optional` to required to reduce FPs.

Another thought involved targeting specific [data formats](https://docs.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats) for the `GetClipboardData` rule. Malware commonly targets text:
```
0x1 = CF_TEXT
0xD = CF_UNICODETEXT
```
Both formats are represented in the existing rule examples. Using these formats, the rule could become a basic block component of a larger function rule. The new rule could also incorporate the `loop` characteristic to target the read operation. 

Lastly, we might consider consolidating the `replace-clipboard-data` and `write-clipboard-data` rules. The `replace` rule includes `user32.EmptyClipboard` but I wonder if `user32.SetClipboardData` would overwrite data existing data that wasn't emptied. The `EmptyClipboard` API could be `optional` in the `write-clipboard-data` rule.

mr-tz:
closes #293

> Lastly, we might consider consolidating the replace-clipboard-data and write-clipboard-data rules. The replace rule includes user32.EmptyClipboard but I wonder if user32.SetClipboardData would overwrite data existing data that wasn't emptied. The EmptyClipboard API could be optional in the write-clipboard-data rule.


williballenthin (approved):

mwilliams31:
@mr-tz any merit to the following aspect of #293? I'm curious what your take on the suggestion is.

> Both formats are represented in the existing rule examples. Using these formats, the rule could become a basic block component of a larger function rule. The new rule could also incorporate the loop characteristic to target the read operation.

mr-tz:
The read clipboard rule includes, we could add a basic block scope to tie them closer together. And add `match: contain loop`.
```yml
        - and:
          - api: user32.GetClipboardData
          - optional:
            - number: 0x1 = CF_TEXT
            - number: 0x7 = CF_OEMTEXT
            - number: 0xD = CF_UNICODETEXT
```

For the write I think we're fine without the specific types.

How does that sound?

mwilliams31:
Ahh, I didn't realize those formats were added in a previous commit. I'll defer to you on the basic block and loop modifications. Thanks for taking a closer look at these rules!

mr-tz:
Thanks @mwilliams31, all updated per you suggestions.

PR #764, opened 2023-05-26
Create read-from-mailslot.yml

crowface28:


williballenthin (approved):
logic looks good. lets see what CI says

crowface28:
Looks like it failed for no sample provided.  Because reasons, I can't supply that.  Possible to skip?

williballenthin:
could we use VT to find an alternative sample? Otherwise let's move the rule to the nursery until we can find a test case.

williballenthin:
(for future reference, no need to provide references to non-public samples. feel free to update the PR to redact that if you'd like.)

crowface28:
Removed example and moved to nursery.  Will work on compiling a sample we can use for linting

williballenthin (approved):
thank you!

PR #767, opened 2023-05-31
Add rule hash-data-using-sha512managed-in-dotnet.yml

jlmndt:
Adds rule for detecting hashing using .NET's Sha512Managed class


williballenthin (approved):

williballenthin:
thank you!

PR #775, opened 2023-06-20
Added rule to detect minifilter driver enumeration

jaxxpnd:
Added CAPA rules to detect minifilter driver enumeration, which can assist in detecting installed AV engines on a compromised system.

### DIFF
--- a/anti-analysis/anti-av/enumerate-minifilter-drivers.yml
+++ b/anti-analysis/anti-av/enumerate-minifilter-drivers.yml
@@ -1,0 +1,18 @@
+rule:
+  meta:
+    name: enumerate minifilter drivers
+    namespace: anti-analysis/anti-av
+    authors:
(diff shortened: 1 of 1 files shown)


williballenthin (changes requested):
logic looks good, though im hesitant to label *any* instance of `FltEnumerateFilters` as anti-AV. could we place this under `host-interaction` somewhere? 

is there more logic we could add to tighten this up to definitely be "anti-av"?

mr-tz:
The referenced sample is also not yet in our test files corpus. So we need to add it there or remove the example section.

jaxxpnd:
I moved this rule and placed it under host-interaction/filter and added the sample to the samples repository.

mr-tz:
Can you add the testfile to https://github.com/mandiant/capa-testfiles, please?

jaxxpnd:
Added the file now

PR #785, opened 2023-07-04
Add create-vmci-socket.yml

jtothej:
Add create-vmci-socket.yml

### DIFF
--- a/communication/socket/create-vmci-socket.yml
+++ b/communication/socket/create-vmci-socket.yml
@@ -1,0 +1,27 @@
+rule:
+  meta:
+    name: create VMCI socket
+    namespace: communication/socket
+    authors:
+      - jakub.jozwiak@mandiant.com
+    scope: basic block
+    mbc:
+      - Communication::Socket Communication::Create Socket [C0001.003]
+    references:
+      - https://www.vmware.com/products/beta/ws/VMCIsockets.pdf
+    examples:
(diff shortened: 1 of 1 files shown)


communication/socket/create-vmci-socket.yml
```diff
@@ -0,0 +1,27 @@
+rule:
+  meta:
+    name: create VMCI socket
+    namespace: communication/socket
```
williballenthin, line 4: do you think this is most related to "socket" functionality or VM interaction?

williballenthin: im not too familar with this tech, but i wonder if the namespace should be more related to vm interaction, maybe like `hsot-interaction/virtual-machine/vmci` or something

williballenthin (approved):

communication/socket/create-vmci-socket.yml
jtothej: It's a good point and I think that VMCI sockets could fit in both categories/namespaces but ultimately they are just another socket type so `communication/socket` (or `communication/socket/vmci`) seems to be more adequate.

williballenthin: cool, thanks for confirming!

mr-tz (approved):
nice, thanks!

PR #786, opened 2023-07-04
Add switch-active-desktop.yml

jtothej:
Add switch-active-desktop.yml

### DIFF
--- a/anti-analysis/anti-debugging/debugger-evasion/switch-active-desktop.yml
+++ b/anti-analysis/anti-debugging/debugger-evasion/switch-active-desktop.yml
@@ -1,0 +1,20 @@
+rule:
+  meta:
+    name: switch active desktop
+    namespace: anti-analysis/anti-debugging/debugger-evasion
+    authors:
+      - jakub.jozwiak@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Debugger Evasion [T1622]
+    mbc:
+      - Anti-Behavioral Analysis::Debugger Evasion [B0002]
+    references:
+      - https://anti-debug.checkpoint.com/techniques/interactive.html#switchdesktop
+    examples:
+      - 26beba7352a32b803aa19e0782011a383a1df19549910e7b2f2f244e49678524:0x10001670
+  features:
+    - and:
+      - api: user32.CreateDesktop
+      - api: user32.SwitchDesktop
+      - number: 0x182 = DESKTOP_CREATEWINDOW | DESKTOP_WRITEOBJECTS | DESKTOP_SWITCHDESKTOP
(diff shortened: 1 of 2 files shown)


williballenthin (approved):

anti-analysis/anti-debugging/debugger-evasion/switch-active-desktop.yml
```diff
@@ -0,0 +1,20 @@
...
+      - Anti-Behavioral Analysis::Debugger Evasion [B0002]
+    references:
+      - https://anti-debug.checkpoint.com/techniques/interactive.html#switchdesktop
+    examples:
+      - 26beba7352a32b803aa19e0782011a383a1df19549910e7b2f2f244e49678524:0x10001670
+  features:
+    - and:
+      - api: user32.CreateDesktop
+      - api: user32.SwitchDesktop
+      - number: 0x182 = DESKTOP_CREATEWINDOW | DESKTOP_WRITEOBJECTS | DESKTOP_SWITCHDESKTOP
```
mr-tz, line 20: should we also track other combinations of these constants or do they all have to be present?

jtothej: I was thinking about it too - probably most combinations of access rights would work here as long as they include `DESKTOP_SWITCHDESKTOP`. Not sure if it wouldn't be better to just drop them instead of listing all possible combinations?

mr-tz: If the APIs are reliant enough by themselves, let's mark the number (or numbers) as optional

jtothej: Hard to say how common it is though I did not find that many examples of code calling `CreateDesktop()` and `SwitchDesktop()` in the same function. So maybe let's move it to `host-interaction/gui` while keeping references to potential use for debugger evasion?

mr-tz: Good idea. That sounds great.

mr-tz: Thanks!

> while keeping references to potential use for debugger evasion

What about that, should we add a `description` for this?

jtothej: I was thinking about listed `att&ck` and `mbc` techniques and Check Point reference describing use of `CreateDesktop()` and `SwitchDesktop()` - would that be enough?

mr-tz: sure, thanks!

PR #787, opened 2023-07-05
Update and promote hide-thread-from-debugger.yml

jtothej:
Update and promote hide-thread-from-debugger.yml


anti-analysis/anti-debugging/debugger-evasion/hide-thread-from-debugger.yml
```diff
@@ -0,0 +1,33 @@
...
+          - or:
+            - api: NtSetInformationThread
+            - api: ZwSetInformationThread
+          - number: 0x11 = ThreadHideFromDebugger
+      - and:
+        - or:
+          - string: "NtSetInformationThread"
+          - string: "ZwSetInformationThread"
+        - api: GetProcAddress
+        - api: GetCurrentThread
```
mr-tz, line 32: How about `match: link function at runtime on Windows` here? Would be cool if we would extract these as API features using some heuristics/analysis and don't require extra rule logic like this.

jtothej: That works too! Updated in https://github.com/mandiant/capa-rules/pull/787/commits/299ba4258b2bb19ac60558ca4681c0ab89b4f5e2

mr-tz: thanks!

PR #790, opened 2023-07-06
Add patch-event-tracing-for-windows-function.yml

jtothej:
Add patch-event-tracing-for-windows-function.yml


mr-tz (commented):
should this include the string (api name) features also as api features?

jtothej:
> should this include the string (api name) features also as api features?

I did not include them as all implementations I've seen just get address of ETW related functions via `GetProcAddress()`, if you think it makes sense to add api entries as well I'll update the rule.

PR #792, opened 2023-07-10
Update hash-data-using-fnv.yml

sara-rn:


data-manipulation/hashing/fnv/hash-data-using-fnv.yml
```diff
@@ -33,3 +40,4 @@ rule:
           - or:
             - mnemonic: imul
             - mnemonic: mul
+            - api: _allmul
```
mr-tz, line 43: should this be a `function-name` feature?

linter fails:   FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example

williballenthin: `function-name` is for matching the name of the current function. i think `api` is correct here.

mr-tz (approved):

mr-tz:
Thanks!

PR #798, opened 2023-07-13
Add patch-antimalware-scan-interface-function.yml and updated patch-e…

jtothej:
Add patch-antimalware-scan-interface-function.yml and updated patch-event-tracing-for-windows-function.yml

### DIFF
--- a/anti-analysis/anti-av/patch-antimalware-scan-interface-function.yml
+++ b/anti-analysis/anti-av/patch-antimalware-scan-interface-function.yml
@@ -1,0 +1,33 @@
+rule:
+  meta:
+    name: patch Antimalware Scan Interface function
+    namespace: anti-analysis/anti-av
+    authors:
+      - jakub.jozwiak@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Impair Defenses::Disable or Modify Tools [T1562.001]
+    mbc:
+      - Defense Evasion::Disable or Evade Security Tools [F0004]
+    references:
+      - https://fluidattacks.com/blog/amsi-bypass/
+    examples:
+      - edb92795c06a2bde47e652639327253a1148ee675ba2f0d1d9ac8690ef1820b1:0x14001126C
+  features:
+    - and:
+      - match: link function at runtime on Windows
+      - or:
+        - api: kernel32.VirtualProtect
+        - api: kernel32.VirtualProtectEx
+        - api: ntdll.NtProtectVirtualMemory
+        - api: ZwProtectVirtualMemory
+        - string: "VirtualProtect"
+        - string: "VirtualProtectEx"
+        - string: "NtProtectVirtualMemory"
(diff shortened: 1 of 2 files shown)


anti-analysis/anti-av/patch-antimalware-scan-interface-function.yml
```diff
@@ -0,0 +1,33 @@
...
+      - match: link function at runtime on Windows
+      - or:
+        - api: kernel32.VirtualProtect
+        - api: kernel32.VirtualProtectEx
+        - api: ntdll.NtProtectVirtualMemory
+        - api: ZwProtectVirtualMemory
+        - string: "VirtualProtect"
+        - string: "VirtualProtectEx"
+        - string: "NtProtectVirtualMemory"
+        - string: "ZwProtectVirtualMemory"
```
williballenthin, line 27: i think we should factor this logic out into a library rule, like `patch read-only memory at runtime` or something

jtothej: I was initially thinking about using `allocate RWX memory` (which in turn depends on `allocate memory`) but `allocate memory` matches on both memory allocation functions and functions that change of protection of already allocated memory.
So maybe we could split `allocate memory` to actual `allocate memory` and something like `change memory protection` (including `VirtualProtect`, `VirtualProtectEx`, `NtProtectVirtualMemory`, `ZwProtectVirtualMemory`). The we could have `change memory protection to RWX` and use it in rules in this PR. The problem with this approach would be figuring out which rules currently using `allocate memory` rule depend on memory allocation functions, which depend on change of memory functions and which on both.

williballenthin: yup, that sounds like a great refactor!

mr-tz: refactoring this in #836

mr-tz, line 27:
```suggestion
      - match: change memory protection
```

anti-analysis/anti-av/patch-event-tracing-for-windows-function.yml
```diff
@@ -19,9 +19,11 @@ rule:
       - match: link function at runtime on Windows
       - or:
         - api: kernel32.VirtualProtect
+        - api: kernel32.VirtualProtectEx
         - api: ntdll.NtProtectVirtualMemory  # exported by only ntdll, not ntoskrnl
         - api: ZwProtectVirtualMemory  # exported by both ntdll and ntoskrnl
         - string: "VirtualProtect"
+        - string: "VirtualProtectEx"
         - string: "NtProtectVirtualMemory"
         - string: "ZwProtectVirtualMemory"
```
mr-tz, line 28:
```suggestion
      - match: change memory protection
```

PR #805, opened 2023-07-22
Update get-os-version.yml - Get OS version via PEB

jtothej:
Update get-os-version.yml - Get OS version via PEB


williballenthin (approved):
nice!

mr-tz (approved):

PR #807, opened 2023-08-01
Update metadata and promote create-shortcut-via-ishelllink.yml

jtothej:
Update metadata and promote create-shortcut-via-ishelllink.yml


williballenthin (approved):
thanks @jtothej!

PR #809, opened 2023-08-02
Add get-uefi-variable.yml and set-uefi-variable.yml

jtothej:
Add get-uefi-variable.yml and set-uefi-variable.yml


williballenthin (approved):
🤔  im really interested to see where this matches globally

PR #811, opened 2023-08-08
Add check-for-av-emulation-using-virtualallocexnuma.yml

jtothej:
Add check-for-av-emulation-using-virtualallocexnuma.yml

### DIFF
--- a/anti-analysis/anti-av/check-for-av-emulation-using-virtualallocexnuma.yml
+++ b/anti-analysis/anti-av/check-for-av-emulation-using-virtualallocexnuma.yml
@@ -10,7 +10,7 @@ rule:
     references:
       - https://www.purpl3f0xsecur1ty.tech/2021/03/30/av_evasion.html
     examples:
-      - 23604a06b0720a430f8d6f6b14b589d850e4cfd291a47f22f199324f21169c1a:0x10084520
+      - 17b31898c3613d6e86d06260c7e6f7738d3afdf0f0a4010e2f5259c502d38b09:0x10001A1D
   features:
     - and:
       - api: kernel32.VirtualAllocExNuma
--- a/anti-analysis/anti-av/check-for-av-emulation-using-virtualallocexnuma.yml
+++ b/anti-analysis/anti-av/check-for-av-emulation-using-virtualallocexnuma.yml
@@ -1,0 +1,17 @@
+rule:
+  meta:
+    name: check for AV emulation using VirtualAllocExNuma
+    namespace: anti-analysis/anti-av
+    authors:
+      - jakub.jozwiak@mandiant.com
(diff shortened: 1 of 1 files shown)


anti-analysis/anti-av/check-for-av-emulation-using-virtualallocexnuma.yml
```diff
@@ -0,0 +1,17 @@
...
+    mbc:
+      - Anti-Behavioral Analysis::Emulator Evasion [B0005]
+    references:
+      - https://www.purpl3f0xsecur1ty.tech/2021/03/30/av_evasion.html
+    examples:
+      - 17b31898c3613d6e86d06260c7e6f7738d3afdf0f0a4010e2f5259c502d38b09:0x10001A1D
+  features:
+    - and:
+      - api: kernel32.VirtualAllocExNuma
+      - api: kernel32.GetCurrentProcess
```
mr-tz, line 17: I don't think just these two calls indicate any anti-AV. Can you elaborate?

jtothej: I have not tested it by myself but multiple sources claim that `VirtualAllocExNuma` is not/was not handled well by AV emulation engines and thus might be used for emulation/sandbox detection. In all instances (sample source code and malware) I found this is typically done by calling `VirtualAllocExNuma` with current process handle and checking if return value is NULL.

mr-tz: Benign code could work the same though, see ,e.g., https://learn.microsoft.com/en-us/windows/win32/memory/allocating-memory-from-a-numa-node. So, I don't think that's a check we could rely on here. Thoughts?

jtothej: Absolutely agree it is completely benign - at the same time it seemed to be quite unique/rare combination (VT: ~35k samples importing both, ~1.5k samples with 0 detections). I can now definitely see samples using both functions in intended way so let's reject this PR. Thank you for the review!

williballenthin: could we perhaps rename the rule to simply describe the behavior for what it is, and then let some downstream system use it for detection, if desired?

the distribution on VT is interesting so i think there's value to having this rule. a capa rule doesn't have to describe only malicious things.

"allocate memory on this NUMA node" or similar?

mr-tz: Good point, we currently (only) include `VirtualAllocExNuma` in the `allocate memory` rule.

PR #814, opened 2023-08-17
Add dynamic rules

yelhamer:
This is a draft PR for adding the mixed-scope rules to capa.

The rules were generated using this script: https://github.com/mandiant/capa/pull/1696

Once we've handled the main parts of merging the capa dynamic branch, I'll go back to this PR and add the missing comments (either manually or by updating the migration script)


anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
```diff
@@ -3,8 +3,9 @@ rule:
     name: check for sandbox and av modules
     namespace: anti-analysis/anti-av
     authors:
-      - "@_re_fox"
-    scope: basic block
+      - '@_re_fox'
```
mr-tz, line 6: prefer to stick to `"`

anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
```diff
@@ -3,8 +3,9 @@ rule:
...
+    scopes:
+      static: basic block
```
mr-tz, line 8: this rule could be run on `dynamic: thread`?

anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
```diff
@@ -34,9 +35,9 @@ rule:
           description: Virtual PC
         - string: /wpespy\.dll/i
           description: WPE Pro
-        - string: /cmdvrt(64|32).dll/i
+        - string: "/cmdvrt(64|32).dll/i"
```
mr-tz, line 38: there shouldn't be quotes around regex strings

anti-analysis/anti-av/patch-event-tracing-for-windows-function.yml
```diff
@@ -19,8 +21,8 @@ rule:
       - match: link function at runtime on Windows
       - or:
         - api: kernel32.VirtualProtect
-        - api: ntdll.NtProtectVirtualMemory  # exported by only ntdll, not ntoskrnl
```
mr-tz, line 22: comments missing

anti-analysis/anti-debugging/debugger-evasion/hide-thread-from-debugger.yml
```diff
@@ -18,12 +20,19 @@ rule:
...
+            - or:
+              - api: NtSetInformationThread
+              - api: ZwSetInformationThread
+            - number: 0x11 = ThreadHideFromDebugger
+        - thread:
+          - and:
+            - or:
+              - api: NtSetInformationThread
+              - api: ZwSetInformationThread
+            - number: 0x11 = ThreadHideFromDebugger
```
mr-tz, line 35: For future improvements, we may think about unifying this to make it easier to read, like `- basic block or thread` maybe?

The real fix is to support call scope, of course

anti-analysis/anti-vm/vm-detection/check-for-windows-sandbox-via-device.yml
```diff
@@ -16,4 +18,4 @@ rule:
   features:
     - and:
       - match: create or open file
-      - string: "\\\\.\\GLOBALROOT\\device\\vmsmb"
+      - string: \\.\GLOBALROOT\device\vmsmb
```
mr-tz, line 21: quotes and escape characters

anti-analysis/anti-vm/vm-detection/check-for-windows-sandbox-via-genuine-state.yml
```diff
@@ -15,10 +17,19 @@ rule:
...
+            - api: SLIsGenuineLocal
+      - or:
+        - basic block:
+          - and:
+            - api: UuidFromString
+            - string: "55c92734-d682-4d71-983e-d6ec3f16059f"
+        - thread:
+          - and:
+            - api: UuidFromString
+            - string: "55c92734-d682-4d71-983e-d6ec3f16059f"
```
mr-tz, line 35: here some syntactic sugar would help again, like
```yml
- basic block or thread:
  - and:
    - ...
```

anti-analysis/packer/aspack/packed-with-aspack.yml
```diff
@@ -16,9 +18,9 @@ rule:
       - 2055994ff75b4309eee3a49c5749d306
   features:
     - or:
+      - string: "The procedure entry point %s could not be located in the dynamic link library %s"
+      - string: "The ordinal %u could not be located in the dynamic link library %s"
       - section: .aspack
```
mr-tz, line 23: Not a problem, but maybe noteworthy that the features get reordered here.

anti-analysis/packer/huan/packed-with-huan.yml
```diff
@@ -4,7 +4,9 @@ rule:
     namespace: anti-analysis/packer/huan
     authors:
       - jakub.jozwiak@mandiant.com
-    scope: file
+    scopes:
+      static: file
+      dynamic: file
```
mr-tz, line 9: good shortcut here may be to only update the scopes and don't even worry about the features for `file` scope rules

collection/file-managers/gather-blazeftp-information.yml
mr-tz, line 1: made it to here in my first review session

mr-tz (commented):
Great progress, I've highlighted issues I've seen inline.

anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
yelhamer: hmm, perhaps it should be `dynamic: process` instead?

mr-tz: or that, yes

williballenthin: i think we should use the most specific scope as possible. so we should prefer thread rather than process.

im guessing that we'll want to use thread in almost all cases, unless we're looking for combinations of behaviors across an entire program, for which we only have a few rules today.

thoughts?

anti-analysis/anti-av/patch-event-tracing-for-windows-function.yml
```diff
@@ -4,7 +4,9 @@ rule:
     namespace: anti-analysis/anti-av
     authors:
       - jakub.jozwiak@mandiant.com
-    scope: function
+    scopes:
+      static: function
+      dynamic: process
```
williballenthin, line 9: ditto here: i think this should be thread, not process.

anti-analysis/anti-debugging/debugger-detection/check-for-trap-flag-exception.yml
```diff
@@ -4,7 +4,8 @@ rule:
     namespace: anti-analysis/anti-debugging/debugger-detection
     authors:
       - michael.hunhoff@mandiant.com
-    scope: basic block
+    scopes:
+      static: basic block
```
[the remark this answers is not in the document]
williballenthin: @mr-tz @yelhamer

anti-analysis/anti-debugging/debugger-evasion/hide-thread-from-debugger.yml
williballenthin: yeah, this repetition is really tedious.

it makes the logic hard to follow, and also makes future updates more difficult.

i think we can pretty quickly add some special cases to the rules parser to enable a combination of hard-coded scope combinations. we should remove this once the call scope is implemented though.

anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
williballenthin: furthermore, in this specific example, i think this should be `call`, because we're trying to match `GetModuleHandle("avghook.dll" or "snxhk.dll" or...)`.

im not sure if automated script can help us here. maybe basic block should map to call if feasible?

anti-analysis/anti-av/protect-spawned-processes-with-mitigation-policies.yml
```diff
@@ -4,7 +4,9 @@ rule:
     namespace: anti-analysis/anti-av
     authors:
       - jakub.jozwiak@mandiant.com
-    scope: basic block
+    scopes:
+      static: basic block
+      dynamic: thread
```
williballenthin, line 9: this should be call scope.

anti-analysis/anti-debugging/debugger-detection/check-for-debugger-via-api.yml
```diff
@@ -5,7 +5,9 @@ rule:
     authors:
       - michael.hunhoff@mandiant.com
       - anushka.virgaonkar@mandiant.com
-    scope: function
+    scopes:
+      static: function
+      dynamic: process
```
williballenthin, line 10: another call scope

anti-analysis/anti-debugging/debugger-detection/check-for-peb-ntglobalflag-flag.yml
```diff
@@ -4,7 +4,9 @@ rule:
     namespace: anti-analysis/anti-debugging/debugger-detection
     authors:
       - moritz.raabe@mandiant.com
-    scope: function
+    scopes:
+      static: function
+      dynamic: process
```
williballenthin, line 9: this rule can't match in dynamic analysis because it relies on an offset feature being present.

williballenthin: it also relies on the basic block subscope matching, which also doesn't work in dynamic analysis. so, i think the rule should be marked as unsupported in dynamic scope.

anti-analysis/anti-debugging/debugger-detection/check-for-time-delay-via-gettickcount.yml
```diff
@@ -4,7 +4,9 @@ rule:
     namespace: anti-analysis/anti-debugging/debugger-detection
     authors:
       - michael.hunhoff@mandiant.com
-    scope: function
+    scopes:
+      static: function
+      dynamic: process
```
williballenthin, line 9: this rule also cannot match in dynamic scope due to it relying on a basic block match

anti-analysis/anti-debugging/debugger-detection/check-for-trap-flag-exception.yml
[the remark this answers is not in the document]
williballenthin: an additional benefit is that we can potentially merge only a subset of migrated rules, those with explicit `scopes.dynamic` set, and consider any rules without this to be needing a migration. then we can perform the migration in phases, perhaps easy rules first and harder rules manually.

anti-analysis/anti-av/check-for-sandbox-and-av-modules.yml
mr-tz: or at least flag for manual review

anti-analysis/anti-debugging/debugger-detection/check-for-trap-flag-exception.yml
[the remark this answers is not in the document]
mr-tz: meaning all rules need explicit `static` and `dynamic` scope definitions?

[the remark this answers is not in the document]
williballenthin: all rules should have them, or the linter should complain, and indicate that the migration for that rule hasn't been completed yet.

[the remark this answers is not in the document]
yelhamer: I am not opposed to this syntax, but I prefer the current syntax (missing dynamic == invalid for dynamic) since it feels more straight-forward. But as I said I am open t both if you wish to switch to that instead.

As for picking only a subset of rules that have the `scopes.dynamic` set, I think that it wouldn't be very beneficial since we currently don't have any custom-made dynamic rules for the new syntax, but rather only migrated rules. If we want to proceed with the rules migration gradually then I suggest we pick rules that have only a static scope, and add the other ones later on.

mr-tz:
I've started to merge some open PRs and we should take another look there. We then may want to open a new PR for the final upgrade process.

yelhamer:
I see. I'll open a final PR once all the mistakes from this grt addressed

anti-analysis/anti-av/block-operations-on-executable-memory-pages-using-arbitrary-code-guard.yml
```diff
@@ -3,7 +3,7 @@ rule:
     name: block operations on executable memory pages using Arbitrary Code Guard
     namespace: anti-analysis/anti-av
     authors:
-      - jakub.jozwiak@mandiant.com
+      - "jakub.jozwiak@mandiant.com"
```
mr-tz: can we do this only when necessary, i.e., when they list github usernames (starting with a special character)?

data-manipulation/encryption/aes/encrypt-data-using-aes-mixcolumns-step.yml
```diff
@@ -38,6 +35,6 @@ rule:
       - optional:
         - instruction:
           - and:
-            - description: "byte assignment, e.g., in: b[c] = r[c] << 1; b[c] ^= h * 0x1B;"
+            - description: 'byte assignment, e.g., in: b[c] = r[c] << 1; b[c] ^= h * 0x1B;'
             - mnemonic: mov
-            - operand[0].offset: 0x0
+            - operand[0].offset: '0x0'
```
mr-tz, line 40: there shouldn't be quotes here

data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
```diff
@@ -15,6 +17,40 @@ rule:
       - 112f9f0e8d349858a80dd8c14190e620:0x4CE3B2
   features:
     - or:
+      - or:
+        - and:
+          - number: 0x3bcc908b = CAMELLIA_SIGMA1R
+          - number: 0xa09e667f = CAMELLIA_SIGMA1L
+          - number: 0x4caa73b2 = CAMELLIA_SIGMA2R
+          - number: 0xb67ae858 = CAMELLIA_SIGMA2L
```
mr-tz, line 25: what happened here, some rule part duplication?

anti-analysis/anti-av/patch-event-tracing-for-windows-function.yml
mr-tz: if these persist to be an annoying issue, maybe we don't upgrade them for now and leave for manual review?

data-manipulation/encryption/camellia/encrypt-data-using-camellia.yml
yelhamer: I believe so yeah...
original rule:
image

yelhamer:
I've done a manual pass over the rules via a diffing tool to fix up any missing things.

There's still the issue of the dynamic scopes not being restrictive enough (thread instead of call, etc.). and I am not sure yet how to automate that (especially since the number of features supported by dynamic scopes might increase in the near future).

It might be a good idea to merge this to the dynamic-syntax branch and reactivate the capa ruleset and main tests for now.

williballenthin:
> It might be a good idea to merge this to the dynamic-syntax branch and reactivate the capa ruleset and main tests for now.

I agree that we should do this so we can get our test cases running again. I'll do this shortly.

williballenthin:
> There's still the issue of the dynamic scopes not being restrictive enough (thread instead of call, etc.). and I am not sure yet how to automate that (especially since the number of features supported by dynamic scopes might increase in the near future).

I'm starting to think we should take a phased approach: we should make our high confidence changes first, review and merge, then turn our attention to the next batch of rules to migrates, and so forth. by the end, hopefully there's only a few dozen things to do by hand.

if we have a way to update the syntax of the rules but also mark some rules as "not yet migrated" then we don't have to migrate all the rules in a single atomic operation (which had been our goal, because it avoids the potential for merge conflicts). 

im a little worried about rushing some of these changes in without fully being confident in them. so, incremental migration can help with this, i think.

williballenthin:
im going to merge this now into the `dynamic-syntax` branch so that we can re-enable our tests in capa core

PR #816, opened 2023-08-22
Add dynamic capa rules

yelhamer:
This PR adds rules that work for the new release of capa which adds the extraction of dynamic capabilities.

The rules here allow it for the new version of capa to work as usual for static analysis (since there were some changes to how capa handles rules, i.e. `scopes` keyword, dynamic/static scoping, etc.), but the rules however will not work for dynamic feature extractors (the CAPE feature extractor for now). This is because we are gradually migrating the current rules into the dynamic format, and that progress can be seen on the dynamic-syntax branch here (issue: https://github.com/mandiant/capa/issues/1747)

If you'd like to test out capa's new dynamic capability extraction, use the rules in that branch instead, since the main blocker for merging that is the scopes there aren't restrictive enough (process instead of thread, thread instead of api-call, etc.), as well as the fact that we anticipate the possible addition of some more dynamic features and support for more static characteristic features (possibly)

This PR should be merged when the dynamic capa branch is merged into mandiant/capa:master


mr-tz:
superseded by #839

mr-tz:
superseded by #839

PR #818, opened 2023-08-27
add `send SMS on Android`

mr-tz:
First rule for Android (native code called via JNI), here from a private 64-bit .so file. Very similar code is also referenced in https://stackoverflow.com/questions/30175062/jni-callvoidmethod-leads-to-fatal-signal-6-and-invalid-indirect-reference.

Question: do we want to mix mobile with desktop for namespaces and ATT&CK?

### DIFF
--- a/nursery/send-sms-on-android.yml
+++ b/nursery/send-sms-on-android.yml
@@ -1,0 +1,24 @@
+rule:
+  meta:
+    name: send SMS on Android
+    namespace: communication/sms
+    authors:
+      - "@mr-tz"
+    scope: function
+    # att&ck:
+    #   - Mobile::SMS Control [T1582]
+  features:
+    - and:
+      - os: android
+      # ... = (*env)->FindClass(env, "android/telephony/SmsManager");
+      - string: "android/telephony/SmsManager"
+      - optional:
+        - or:
+          - and:
+            - arch: i386
+            - offset: 0x30 = (*env)->FindClass
+          - and:
+            - arch: amd64
+            - offset: 0x1C = (*env)->FindClass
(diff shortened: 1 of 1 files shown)


nursery/send-sms-on-android.yml
```diff
@@ -0,0 +1,24 @@
+rule:
+  meta:
+    name: send SMS on Android
+    namespace: communication/sms
+    authors:
+      - "@mr-tz"
+    scope: function
+    # att&ck:
+    #   - Mobile::SMS Control [T1582]
```
mr-tz, line 9: we currently only lint for Enterprise techniques, not https://attack.mitre.org/techniques/mobile/

nursery/send-sms-on-android.yml
```diff
@@ -0,0 +1,24 @@
...
+  features:
+    - and:
+      - os: android
+      # ... = (*env)->FindClass(env, "android/telephony/SmsManager");
+      - string: "android/telephony/SmsManager"
+      - optional:
```
williballenthin, line 15: eventually we'll probably want to factor this out, but let's wait until it's used in a few places first.

nursery/send-sms-on-android.yml
williballenthin: seems reasonable to add, assuming it's low effort and maintenance. we could also wait until someone wants to use this and/or we have more substantial coverage.

williballenthin (approved):

nursery/send-sms-on-android.yml
mr-tz: yeah, probably a good idea to write a few dozen rules and talk to potential users first

PR #820, opened 2023-08-29
Add `dynamic: unspecified` to static-only rules

yelhamer:
We have enforced that all rules must specify both the static and dynamic scopes. This PR updates the initial version of the rules (the one generated automatically) to have both scopes specified. Merging this PR should fix the failing the tests at https://github.com/mandiant/capa/pull/1697


yelhamer:
Hmm. Apparently we have 167 static-only rules in our ruleset (~20% of our total number of rules)...

williballenthin (approved):
thank you!

PR #821, opened 2023-09-02
adding new rules based on private Linux sample(s)

mike-hunhoff:


nursery/create-detached-thread-on-linux.yml
```diff
@@ -0,0 +1,11 @@
+rule:
+  meta:
+    name: create detached thread on Linux
```
mr-tz, line 3:
```suggestion
    name: detach thread on Linux
```

mr-tz: or `mark thread detached`

nursery/get-user-database-entry-on-linux.yml
```diff
@@ -0,0 +1,13 @@
...
+    namespace: host-interaction/session
+    authors:
+      - michael.hunhoff@mandiant.com
+    scope: basic block
+  features:
+    - and:
+      - os: linux
+      - or:
+        - api: getpwuid
+        - api: getpwuid_r
```
mr-tz, line 13: also `getpwnam` and `getpwnam_r`?

nursery/get-user-database-entry-on-linux.yml
```diff
@@ -0,0 +1,13 @@
+rule:
+  meta:
+    name: get user database entry on Linux
```
mr-tz, line 3:
```suggestion
    name: get password database entry on Linux
```

nursery/communicate-with-kernel-module-via-netlink-socket-on-linux.yml
```diff
@@ -0,0 +1,13 @@
+rule:
+  meta:
+    name: communicate with kernel module via Netlink socket on Linux
+    namespace: host-interaction/kernel
```
mr-tz, line 4: not sure the namespace is great, but cannot think of anything else right now

mr-tz (commented):
nice, work, a few suggestions inline

nursery/communicate-with-kernel-module-via-netlink-socket-on-linux.yml
mike-hunhoff: Agreed -  I had the same thoughts when creating the rule. I'll leave it as-is for now.

nursery/get-user-database-entry-on-linux.yml
mike-hunhoff: Nice call

nursery/get-user-database-entry-on-linux.yml
mike-hunhoff: Nice catch

mr-tz (approved):
great, thank you!

Issue #706, opened 2023-02-22
Xamarin.Android (Build fully native Android apps using C#)


mike-hunhoff:
Consider writing capa rules for `Xamarin.Android` applications:

> Xamarin.Android exposes the complete Android SDK for .NET developers. Build fully native Android apps using C# or F# in Visual Studio.

([credit](https://learn.microsoft.com/en-us/xamarin/android/))

Developers can write cross-platform C# that is compiled to platform-specific C# (Android, iOS) . Here, we should aim to target Android-specific C# implemented via [Mono.Android](https://learn.microsoft.com/en-us/dotnet/api/android?view=xamarin-android-sdk-13):

<img width="701" alt="Screen Shot 2023-02-22 at 10 42 23 AM" src="https://user-images.githubusercontent.com/42192796/220716864-dedbd640-5e06-4ab4-b206-56843c0f21bf.png">

([credit](https://visualstudio.microsoft.com/xamarin/))

`Xamarin.Android` framework has been leveraged by malware authors:
- https://blog.cyble.com/2021/10/22/fake-voicemail-app-built-through-xamarin-platform-spreads-spyware/
- https://maldr0id.blogspot.com/2015/03/android-malware-goes-mono-net-and-lua.html

Quick hunt on VT finds:
- [f22974d0d7736ed5cf04102d42d9db5f1592463c137921c02595f382411fb595](https://www.virustotal.com/gui/file/f22974d0d7736ed5cf04102d42d9db5f1592463c137921c02595f382411fb595)

We already have some coverage on this sample:

<img width="947" alt="Screen Shot 2023-02-22 at 11 33 12 AM" src="https://user-images.githubusercontent.com/42192796/220724773-02bacbe0-9fcf-4c96-9449-89845a9be844.png">

Deliverables:
- understand `Mono.Android` SDK and its uses by malware authors
- develop capa rules based on this understanding

komen205:
Hello,

I'm not sure if this will be helpful in any way, but maybe identifying that the file is actually a Xamarin application could be quite useful. Since now it seems that it only identifies that the file is a dotnet one.

We can do this by targeting the magic "XALZ", for reference, https://github.com/xamarin/xamarin-android/pull/4686

PR #824, opened 2023-09-04
init add of Xamarin rules

mike-hunhoff:
Adding new rules based on #706 . This PR may be impacted by the discussions in #818 regarding mobile ATT&CK techniques. I wanted to open a PR for discussion before diving too far into creating new rules.

Questions:

- Should Android example files be placed in separate tests folder?

### DIFF
--- a/nursery/access-camera-in-dotnet-on-android.yml
+++ b/nursery/access-camera-in-dotnet-on-android.yml
@@ -1,0 +1,10 @@
+rule:
+  meta:
+    name: access camera in .NET on Android
+    namespace: host-interaction/hardware/camera
+    authors:
+      - michael.hunhoff@mandiant.com
+    scope: function
+  features:
+    - or:
+      - api: Android.Hardware.Camera::Open
--- a/nursery/get-system-information-in-dotnet-on-android.yml
+++ b/nursery/get-system-information-in-dotnet-on-android.yml
@@ -1,0 +1,10 @@
+rule:
+  meta:
+    name: get system information in .NET on Android
+    namespace: host-interaction/os/info
+    authors:
+      - michael.hunhoff@mandiant.com
+    scope: function
+  features:
+    - and:
+      - class: Android.OS.Build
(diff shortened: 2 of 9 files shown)


nursery/access-camera-in-dotnet-on-android.yml
```diff
@@ -0,0 +1,10 @@
+rule:
+  meta:
+    name: access camera in .NET on Android
+    namespace: host-interaction/hardware/camera
+    authors:
+      - michael.hunhoff@mandiant.com
+    scope: function
+  features:
+    - or:
+      - api: Android.Hardware.Camera::Open
```
williballenthin, line 10: restrict to `os: android`? or maybe the name and API namespace are sufficient?

we'll have to come up with a consistent style on things like this. i don't know today.

nursery/get-system-information-in-dotnet-on-android.yml
```diff
@@ -0,0 +1,10 @@
+rule:
+  meta:
+    name: get system information in .NET on Android
```
williballenthin, line 3: seems rather generic. could we make this name a little more specific? is it the OS version or similar?

williballenthin (approved):
so cool to see

nursery/access-camera-in-dotnet-on-android.yml
mr-tz: os and/or format restrictions make sense to add

mr-tz (commented):
I think the testfiles should be separate.

Similarly, do we want to keep the rules files separate? Some rules contain various formats/OS so that's a little messy.
There's pro and cons, so maybe we better open up a separate discussion?

nursery/access-camera-in-dotnet-on-android.yml
mike-hunhoff: Discussed offline w/ @mr-tz . We'll leave this for now and continue discussion in the issue that I just created to track this.

PR #825, opened 2023-09-05
encrypt data using RC4 via SystemFunction032

richardweiss80:
Added Advapi32.SystemFunction032 capa rule.

Kind regards,
Richard

### DIFF
--- a/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
+++ b/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
@@ -1,0 +1,23 @@
+rule:
+  meta:
+    name: encrypt data using RC4 via WinAPI SystemFunctionXXX Procedures
+    namespace: data-manipulation/encryption/rc4
+    authors:
+      - richard.weiss@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
+    mbc:
+      - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
+      - Cryptography::Encrypt Data::RC4 [C0027.009]
+    references:
+      - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
+      - https://blog.gentilkiwi.com/tag/systemfunction032
+  features:
+      - or:
+        - api: SystemFunction032
+        - optional:
+          - and:
+            - string: Advapi32.dll
+            - string: SystemFunction032
+              description: Procedure Address resolved at runtime
--- a/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
+++ b/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
@@ -13,10 +13,14 @@ rule:
     references:
       - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
       - https://blog.gentilkiwi.com/tag/systemfunction032
+    examples:
+      - 1F31917FFF1E43B0BBC72576906B0746A88F978084FADA49312900B5241A9FC8:0x14009E058 # SystemFunction032
   features:
     - or:
       - api: SystemFunction032
       - and:
-        - string: "Advapi32.dll"
+        - or:
+          - substring: "Advapi32"
+          - substring: "cryptsp"
         - string: "SystemFunction032"
           description: Procedure Address resolved at runtime
--- a/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
+++ b/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
@@ -14,13 +14,12 @@ rule:
       - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
       - https://blog.gentilkiwi.com/tag/systemfunction032
     examples:
-      - 3BBDF04C25FCD9876733EAA9163B3ED64D81396E7414619758D9376EDF4C103E:0x1000976C # SystemFunction032
+      - 3BBDF04C25FCD9876733EAA9163B3ED64D81396E7414619758D9376EDF4C103E:0x1000976C # api match
   features:
     - or:
       - api: SystemFunction032
       - and:
         - or:
-          - substring: "Advapi32"
-          - substring: "cryptsp"
+          - string: /advapi32/i
+          - string: /cryptsp/i
         - string: "SystemFunction032"
-          description: Procedure Address resolved at runtime
--- a/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
+++ b/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
@@ -14,7 +14,7 @@ rule:
       - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
       - https://blog.gentilkiwi.com/tag/systemfunction032
     examples:
-      - 1F31917FFF1E43B0BBC72576906B0746A88F978084FADA49312900B5241A9FC8:0x14009E058 # SystemFunction032
+      - 3BBDF04C25FCD9876733EAA9163B3ED64D81396E7414619758D9376EDF4C103E:0x1000976C # SystemFunction032
   features:
     - or:
       - api: SystemFunction032
--- a/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
+++ b/nursery/encrypt-data-using-rc4-via-systemfunction032.yml
@@ -1,6 +1,6 @@
 rule:
   meta:
-    name: encrypt data using RC4 via WinAPI SystemFunctionXXX Procedures
+    name: encrypt data using RC4 via SystemFunction032
     namespace: data-manipulation/encryption/rc4
     authors:
       - richard.weiss@mandiant.com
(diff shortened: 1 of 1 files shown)


nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,22 @@
...
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
+    mbc:
+      - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
+      - Cryptography::Encrypt Data::RC4 [C0027.009]
+    references:
+      - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
+      - https://blog.gentilkiwi.com/tag/systemfunction032
+  features:
+    - or:
```
williballenthin, line 17:
```suggestion
    - and:
```

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,22 @@
...
+      - richard.weiss@mandiant.com
+    scope: function
+    att&ck:
+      - Defense Evasion::Obfuscated Files or Information [T1027]
+    mbc:
+      - Defense Evasion::Obfuscated Files or Information::Encryption-Standard Algorithm [E1027.m05]
+      - Cryptography::Encrypt Data::RC4 [C0027.009]
+    references:
+      - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
+      - https://blog.gentilkiwi.com/tag/systemfunction032
```
williballenthin, line 15: can we find any public examples of this technique that we can reference as an `example`?

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: I decided not to use AND and optional because there will be no capa api finding, if GetProcAddess is used.

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: Yes, I will add. Thank you for the hint. Will have a look this evening for at least one sample.

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
williballenthin: `optional` can also be written as `0 or more` which means it always evaluates to `true` so it shouldn't be used under an `or`. I'd recommend that it only be used under `and` to provide more context to the primary logic. 

I suspect that using `or` with `optional` would match in every possible scope, which is probably not wanted.

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: added example and also created a PR for the testfile

richardweiss80: testfile uploaded and reference changed.

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,26 @@
...
+    examples:
+      - 3BBDF04C25FCD9876733EAA9163B3ED64D81396E7414619758D9376EDF4C103E:0x1000976C # SystemFunction032
```
mr-tz, line 17: can you clarify what the comment indicates?

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,26 @@
...
+  features:
+    - or:
+      - api: SystemFunction032
+      - and:
+        - or:
+          - substring: "Advapi32"
```
mr-tz, line 23: could this be lower case as well?

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,26 @@
...
+          - substring: "cryptsp"
+        - string: "SystemFunction032"
+          description: Procedure Address resolved at runtime
```
mr-tz, line 26: can you make the description more descriptive or remove? :)

mr-tz (commented):
Awesome, thanks! just a few minor comments.

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
mr-tz: the comment may be helpful if there's multiple examples to indicate which way it's used, can you clarify here if it's imported or run-time resolved?

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
mr-tz: `string: /advapi32/i`

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
mr-tz: should be clear from the rule context, so remove?!

mr-tz (approved):
great, thanks!

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,25 @@
...
+    examples:
+      - 3BBDF04C25FCD9876733EAA9163B3ED64D81396E7414619758D9376EDF4C103E:0x1000976C # api match
+  features:
+    - or:
+      - api: SystemFunction032
+      - and:
+        - or:
+          - string: /advapi32/i
+          - string: /cryptsp/i
+        - string: "SystemFunction032"
```
mr-tz, line 25: oh, one other idea: how about an optional: match: `link function at runtime on Windows`?

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: thank you, changed

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: Awesome idea, will add and test it

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,26 @@
...
+    references:
+      - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
+      - https://blog.gentilkiwi.com/tag/systemfunction032
+    examples:
+      - 3BBDF04C25FCD9876733EAA9163B3ED64D81396E7414619758D9376EDF4C103E:0x1000976C # api match
+  features:
+    - or:
+      - api: SystemFunction032
+      - and:
+        - match: link function at runtime on Windows
```
mr-tz, line 22:
```suggestion
        - optional:
          - match: link function at runtime on Windows
```

mr-tz: final suggestion to be a bit more flexible

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,26 @@
...
+        - string: "SystemFunction032"
+        - or:
```
mr-tz, line 24:
```suggestion
        - optional:
```

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
```diff
@@ -0,0 +1,26 @@
...
+          - string: /advapi32/i
+          - string: /cryptsp/i
```
mr-tz, line 26:
```suggestion
          - string: /cryptsp/i
          - characteristic: indirect call
```

mr-tz (commented):
thanks, also see https://github.com/mandiant/capa-rules/pull/828/files
both rules should be consistent now

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: charfateristic: indirect call is optional part of link function at runtime, so I would not reuse in this rule again.

nursery/encrypt-data-using-rc4-via-systemfunction032.yml
richardweiss80: or was intended to use is as 1 or mroe (at least 1) to avoid that this is a string only match before I added "matching of link function at runtime". The decision of at least 1 and not 0 or more was inteded to have a more robust rule. With match we could think about chaning it to:
```
- and:
- string: "systemfunction032"
- or:
- -string: /..../i  
- -string: /.../i
- -match: /../i
```
opportunities:
- optional (0 or more) + AND (your recommendation )-> at least 1 -> or (rule change above) perhaps with performance relevance
- optional (0 or more) + AND for better reading (your recommendation )
- AND + OR for stronger robustness of the rule to avoid using "SystemFunction032" as sting as decoy and a indirect/resolved calling of another function

My recommendation:
Would stay with the last option and loking from time to time in VT the differences between content match of only Systemfunction032 and combined with the module names.

mr-tz: can you elaborate on the robustness of the last option? I think with the string and match present we can be pretty sure (of course not absolutely) what's happening... do the additional DLL strings add more value vs. potential misses where they are missing

mr-tz: friendly ping @richardweiss80

mr-tz:
Thanks!

PR #828, opened 2023-09-17
Update generate-random-numbers-via-rtlgenrandom.yml

### LINKED ISSUE #827: Idea of rule modification: generate-random-numbers-via-rtlgenrandom
https://github.com/mandiant/capa-rules/blame/b9c2bc120e21154fd7e3e1d8b7150f8de92b1a50/data-manipulation/prng/generate-random-numbers-via-rtlgenrandom.yml

Dear Willi,

fantastic rule; thank you for insipiration. What do you think about these changes?

1. Adding a reference. e.g.: 
- https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
- https://blog.gentilkiwi.com/tag/systemfunction036
2. Adding comment to the examples if detected by api or string
3. changing of:
- `string: "advapi32.dll"` to `string: /advapi32/i`: `.dll` is not needed and could be appended by the OS, and the string as argument is used in a case-insensitive way
- deletion of line 21: it is covered by the regex string

It is only an idea and I hope you like it. :)

Kind regards,
Richard

mr-tz:
closes #827

### DIFF
--- a/data-manipulation/prng/generate-random-numbers-via-rtlgenrandom.yml
+++ b/data-manipulation/prng/generate-random-numbers-via-rtlgenrandom.yml
@@ -4,12 +4,16 @@ rule:
     namespace: data-manipulation/prng
     authors:
       - william.ballenthin@mandiant.com
+      - richard.weiss@mandiant.com
     scope: function
     mbc:
       - Cryptography::Generate Pseudo-random Sequence::Use API [C0021.003]
+    references:
+      - https://doxygen.reactos.org/df/d13/sysfunc_8c_source.html
+      - https://blog.gentilkiwi.com/tag/systemfunction036
     examples:
-      - b7841b9d5dc1f511a93cc7576672ec0c:0x10002B80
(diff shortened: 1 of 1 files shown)


williballenthin (approved):

data-manipulation/prng/generate-random-numbers-via-rtlgenrandom.yml
```diff
@@ -4,19 +4,23 @@ rule:
...
       - api: SystemFunction036
       - and:
         - match: link function at runtime on Windows
         - string: "SystemFunction036"
         - optional:
-          - string: "advapi32.dll"
-          - string: "Advapi32.dll"
+          - string: /advapi32/i
+          - string: /cryptsp/i
           - characteristic: indirect call
```
mr-tz, line 26:
```suggestion
```

mr-tz: already part of `link function at runtime on Windows`

mike-hunhoff (approved):

PR #829, opened 2023-09-18
set state tcp connection

johnk3r:
Rule to detect the use of the "SetTcpEntry" API, commonly used by TA to disable EDR.


host-interaction/network/connectivity/set-state-tcp-connection.yml
```diff
@@ -0,0 +1,18 @@
+rule:
+  meta:
+    name: set state tcp connection
```
williballenthin, line 3:
```suggestion
    name: set TCP connection state
```

williballenthin (changes requested):
logic looks great, recommend tweaking the rule name

mr-tz:
please rename to `set-tcp-connection-state.yml` then good to merge

host-interaction/network/connectivity/set-tcp-connection-state.yml
```diff
@@ -0,0 +1,18 @@
+rule:
+  meta:
+    name: set TCP connection state
+    namespace: host-interaction/network/connectivity
+    authors:
+      - "@johnk3r"
+    description: The SetTcpEntry function sets the state of a TCP connection.
+    scope: function
+    att&ck:
+      - Defense Evasion::Impair Defenses [T1562]
```
mr-tz, line 10: Can we be sure this is always the technique here?

johnk3r: MSFT describes the following:

1 - **The SetTcpEntry function sets the state of a TCP connection.**
2 - **Currently, the only state to which a TCP connection can be set is MIB_TCP_STATE_DELETE_TCB.**

So in my understanding yes. Do you have any other experience with this?

mr-tz: I see, should we then add `number: 12 = MIB_TCP_STATE_DELETE_TCB` to the logic?

johnk3r: Nice ideia!


image

What do you think?

mr-tz: looks good, please add it to the rule and then we're g2g

mr-tz (approved):
thank you!

PR #831, opened 2023-09-23
Create add-value-to-global-atom-table.yml

mr-tz:
part of #88

### DIFF
--- a/nursery/add-value-to-global-atom-table.yml
+++ b/nursery/add-value-to-global-atom-table.yml
@@ -1,0 +1,13 @@
+rule:
+  meta:
+    name: add value to global atom table
+    namespace: host-interaction/process/inject
+    authors:
+      - "@mr-tz"
+    scope: function
+    references:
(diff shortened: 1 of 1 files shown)


nursery/add-value-to-global-atom-table.yml
```diff
@@ -0,0 +1,13 @@
...
+    namespace: host-interaction/process/inject
+    authors:
+      - "@mr-tz"
+    scope: function
+    references:
+      - https://www.fortinet.com/blog/threat-research/atombombing-brand-new-code-injection-technique-for-windows
+      - https://github.com/BreakingMalwareResearch/atom-bombing
+  features:
+    - and:
+      - api: GlobalAddAtom
```
mike-hunhoff, line 13:
```suggestion
      - api: AddAtom
      - api: GlobalAddAtom
      - api: GlobalAddAtomEx
```

mike-hunhoff (changes requested):
Great @mr-tz  - I've suggested two additional APIs that I believe we should match here as well. Let me know what you think!

mr-tz:
Great, thanks!

PR #833, opened 2023-10-01
Create capture-process-snapshot.yml

### LINKED ISSUE #527: process snapshot
reference: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/proc_snap/process-snapshotting-functions

example: https://github.com/ORCA666/KCTHIJACK/blob/main/KCTHijack/main.c

mr-tz:
closes #527

### DIFF
--- a/nursery/capture-process-snapshot.yml
+++ b/nursery/capture-process-snapshot.yml
@@ -1,0 +1,10 @@
+rule:
+  meta:
+    name: capture process snapshot
+    namespace: host-interaction/process/dump
+    authors:
+      - "@mr-tz"
+    scope: function
+  features:
+    - or:
+      - api: PssCaptureSnapshot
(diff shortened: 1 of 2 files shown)


nursery/capture-process-snapshot.yml
```diff
@@ -0,0 +1,10 @@
+rule:
+  meta:
+    name: capture process snapshot
```
williballenthin, line 3: can you think of any way to make this name a little more descriptive? like what kind of snapshot?

perhaps "capture process memory snapshot"?

williballenthin (approved):

nursery/capture-process-snapshot.yml
mr-tz: there's multiple information that can be queried, see https://learn.microsoft.com/en-us/windows/win32/api/processsnapshot/ne-processsnapshot-pss_capture_flags, so I suggest data?!

PR #837, opened 2023-10-13
Upgrade rules for static and dynamic scopes

mr-tz:
Similar to #816 based initially on a script.

Plan to fix most rules manually in here moving forward after inline discussions below.

PR #839, opened 2023-10-25
upgrade rules using updated script

mr-tz:
#837 and all the comments broke the GitHub web UI...

Repeating the work now here using a script and fixing things locally - for the most part.


mr-tz:
See https://github.com/mandiant/capa-rules/pull/839/commits/409da0b775bd467d70826e5149974f3b2fffdbc3 for some first updates to rules with subscopes. Is that what you had in mind @williballenthin @yelhamer?

anti-analysis/anti-vm/vm-detection/check-for-windows-sandbox-via-genuine-state.yml
```diff
@@ -15,10 +17,19 @@ rule:
...
           - api: UuidFromString
           - string: "55c92734-d682-4d71-983e-d6ec3f16059f"
+      # dynamic
+      - call:
+        - and:
+          - api: SLIsGenuineLocal
+      - call:
+        - and:
+          - api: UuidFromString
+          - string: "55c92734-d682-4d71-983e-d6ec3f16059f"
```
williballenthin, line 35: this duplication is obviously unfortunate, but i understand why

mr-tz:
If you've looked at a dozen or so changes from each commit (as they roughly group changes) I'm fairly confident in these updates and don't think we need a detailed review of each rule.

williballenthin:
ah, that's a great way to group changes. i'll do this review tomorrow. thanks!

anti-analysis/anti-vm/vm-detection/check-for-windows-sandbox-via-genuine-state.yml
mr-tz: potentially there's just a lot of duplication for `basic block` -> `call`

could we support `basic block / call` syntax to simplify this?

williballenthin (approved):
(the GH PR review interface is basically unusable for this set of commits 🤦🏼 )

I reviewed commit-by-commit, limiting myself to about 20 changed files per commit. I found no blockers.

Things I took away:

  1. good: comment why a flavor is unsupported
  2. good: comments indicating a branch is for static/dynamic flavor
  3. good: nest static/dynamic branches under a single OR to group logic
  4. maybe we can rely on the default logical statement for a subscope to be "AND", which simplifies some rules. consider:

```yaml
- call:
  - and:
    - api: foo
    - number: 1
```

versus:

```yaml
- call:
  - api: foo
  - number: 1
```

things to check:

  a. VMCI socket: should this be scope=call not scope=thread?
  b. protected exception handling: check call counting works as desired
  c. check if file exists: use scope=call?

---

sidebar: here's how i did the review commit by commit, mostly for my personal future reference

```
git switch --detach b05c54734b56dfc77c542b9df723ad88f5799678
git reset $(git merge-base HEAD HEAD~)
```

this checks out the commit i want to inspect, and then creates a workspace as if only the changes in that commit are active an uncommited. so, then i can do `git status` to see all the modified fields and `git diff` to see the diffs.

notably, i can use tools like tig, gitui, and vs code to review the diffs and stage/unstage to mark things i agree with.

williballenthin:
- [ ] ensure there are no more TODOs associated with scopes

williballenthin:
@mr-tz please see [dbefb55](https://github.com/mandiant/capa-rules/pull/839/commits/dbefb55f7be802757e925935ae8e454acf60e211)

mr-tz:
Is `VMCI_SOCKETS_GET_AF_VALUE` a valid argument to `socket`?

The other changes look good, thank you!

williballenthin:
via here:  https://github.com/mandiant/capa/blob/210a13d94ea40114e01b40927c77c1c74047780f/capa/rules/__init__.py#L544

an `instruction` subscope acts implicitly like an `and:` so we can save a line/indentation by not adding those. same with `call`. maybe we can do a cleanup of all the rules in another PR.

williballenthin:
> Is VMCI_SOCKETS_GET_AF_VALUE a valid argument to socket?

ah, you're right, this is only passed to `DeviceIoControl`/`ioctl`. maybe the logic should look like:




```yaml
      - and:
        - os: windows
        - api: DeviceIoControl
        - number: 0x81032068 = VMCI_SOCKETS_GET_AF_VALUE
        - optional:
          - api: socket
      - and:
        - os: linux
        - api: ioctl
        - number: 0x7B8 = VMCI_SOCKETS_GET_AF_VALUE
        - optional:
          - api: socket
```

williballenthin (approved):
i think this PR is ready for merge (once dynamic analysis is merged)!

mr-tz:
For `call` scope the optional `socket` would never match and the scopes should still be thread (and function) IMO.

mr-tz:
> via here: https://github.com/mandiant/capa/blob/210a13d94ea40114e01b40927c77c1c74047780f/capa/rules/__init__.py#L544
> 
> an `instruction` subscope acts implicitly like an `and:` so we can save a line/indentation by not adding those. same with `call`. maybe we can do a cleanup of all the rules in another PR.

I don't see that implemented for call, but think we should add and simplify the rules.

mr-tz:
rebased to master and:
- changed scopes again for VMCI socket: 5e2dae155ff67edc7489dc779c6f8159285aecb9
- updated scopes for new rules: 54308899027bf3c43e41067b0f986f1be55449ee
- graduated rule: 305adfd16bab08b2526596e56b50b638d8f4bf43

williballenthin (approved):

PR #841, opened 2023-11-09
identify .NET single file bundles

sara-rn:
.NET single file deployments allow all the dependencies to be included in a single binary.  Other capa rules might not be relevant for this single file bundles. Perhaps a warning could be displayed, same as for MSI. Additional information can be found in https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # capa rules
 
 [![Rule linter status](https://github.com/mandiant/capa-rules/workflows/CI/badge.svg)](https://github.com/mandiant/capa-rules/actions?query=workflow%3A%22CI%22)
-[![Number of rules](https://img.shields.io/badge/rules-847-blue.svg)](rules)
+[![Number of rules](https://img.shields.io/badge/rules-848-blue.svg)](rules)
 [![License](https://img.shields.io/badge/license-Apache--2.0-green.svg)](LICENSE.txt)
 
 This is the standard collection of rules for [capa](https://github.com/mandiant/capa) - the tool to automatically identify capabilities of programs.
--- a/bundled-with-dotnet-single-file-deployment.yml
+++ b/bundled-with-dotnet-single-file-deployment.yml
@@ -1,0 +1,19 @@
+rule:
+  meta:
+    name: bundled with .NET single-file deployment
+    namespace: runtime/dotnet
+    authors:
+      - sara.rincon@mandiant.com
+    scope: file
+    references:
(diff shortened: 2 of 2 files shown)


bundled-with-dotnet-single-file-deployment.yml
```diff
@@ -0,0 +1,19 @@
...
+    references:
+      - https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli
+      - https://github.com/dotnet/runtime/blob/84de9b678613675e0444b265905c82d33dae33a8/src/installer/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
+    examples:
+      - a83339b07cf2bf1aeda192de42760c625d4b2f106260f58c902d02f4766848d5
+  features:
+    - or:
+      - and:
+        - match: contains PDB path
+        - string: "singlefilehost.pdb"
```
mr-tz, line 17: can you add a description why this may be a good signal?

sara-rn: sure.  this is the symbol file for the native executable when publish single file is set to True

bundled-with-dotnet-single-file-deployment.yml
```diff
@@ -15,5 +15,6 @@ rule:
       - and:
         - match: contains PDB path
         - string: "singlefilehost.pdb"
+          description = symbol file for the native executable when publish single file is true
```
mr-tz, line 18:
```suggestion
          description = symbol file for the native executable when PublishSingleFile is true
```

README.md
mr-tz, line 1: please remove this file from the PR

bundled-with-dotnet-single-file-deployment.yml
```diff
@@ -0,0 +1,20 @@
+rule:
+  meta:
+    name: bundled with .NET single-file deployment
+    namespace: runtime/dotnet
```
mr-tz, line 4: should this go into the `internal\limitation\file` directory/namespace to avoid further processing?

README.md
mr-tz, line 1: sorry, I meant from the changes, not to delete the file :)

sara-rn: yep, understood, didn't know how to I guess, can you reject this specific commit?

mr-tz:
Cool, can you address the parsing/linter fail and then this should be good to go.

mr-tz:
There's also `packaged as single-file .NET application` which we should consider as part of this update.

sara-rn:
will create a new PR for this

PR #843, opened 2023-11-13
PLUGX: make more restrictive to fix FP

### LINKED ISSUE #842: PLUGX rule is too loose
## Summary

https://github.com/mandiant/capa-rules/blob/master/malware-family/plugx/match-known-plugx-module.yml

can match on a function with the numbers:  0x88888888  and 0x1001. the former seems likely to be seen in optimized arithmetic implementations, and 0x1001 seems like a reasonable enum or error constant.

## Examples

this is definitely not PLUGX: f9b962d6668a3813695b5c8dab533d6ecbcb2cef5c5fad6726637d735f6dd33b

## Possible improvements

make the command ID `or` block something like `four or more:`

williballenthin:
closes #842

### DIFF
--- a/malware-family/plugx/match-known-plugx-module.yml
+++ b/malware-family/plugx/match-known-plugx-module.yml
@@ -23,6 +23,7 @@ rule:
           - operand[1].number: 0x20120225
           - operand[1].number: 0x20120324
           - operand[1].number: 0x20121107
+          - operand[1].number: 0x20140613
           - operand[1].number: 0x20190301
           - operand[1].number: 0x20190520
           - operand[1].number: 0x20200208
--- a/malware-family/plugx/match-known-plugx-module.yml
+++ b/malware-family/plugx/match-known-plugx-module.yml
@@ -6,73 +6,80 @@ rule:
     authors:
       - still@teamt5.org
     description: the sample references known PlugX watermarks (hexified YYYYMMDD + command opcode)
-    scope: function
+    scope: basic block
     references:
       - https://circl.lu/assets/files/tr-12/tr-12-circl-plugx-analysis-v1.pdf
       - https://www.fireeye.com/blog/threat-research/2014/07/pacific-ring-of-fire-plugx-kaba.html
       - https://www.avira.com/en/blog/new-wave-of-plugx-targets-hong-kong
     examples:
-      - 64E9F62840DB2F65FC717CFAF99081F9:0x10024950
+      - 64E9F62840DB2F65FC717CFAF99081F9:0x10024BCB
   features:
     - and:
-      - or:
+      - instruction:
         - description: module timestamp
-        - number: 0x20120225
-        - number: 0x20120324
-        - number: 0x20121107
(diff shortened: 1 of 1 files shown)


williballenthin:
@Still34 FYI

williballenthin:
looks like i need to review the linter failure before this is merged. please review the intent of the logic change in the meantime.

Still34:
Shoot, I'll take a look in a bit

mr-tz (approved):
logic LGTM

Still34:
Looks perfectly fine to me as well

williballenthin:
when I proposed the changes here I imagined (without looking, sorry) that there'd be a large command ID dispatcher that had a bunch of the command IDs referenced in the same function, using a case/switch statement. turns out this is not the case for plugx. instead, we have individual functions that handle each command, parsing or creating the protocol, like:

image


note that we see:
```
mov [base+0], TIMESTAMP
mov [base+4], COMMAND
```


note that along the way i did find a FP in the same sample: see how 0x4000 is used as a bitmask in a related function, not as the command ID:

image


so, i'll propose updating the rule to:

```yaml
- and:
  - instruction:
    - mnemonic: mov
    - operand[0].offset: 0
    - or:
      - operand[1].number: TIMESTAMP1
      - operand[1].number: TIMESTAMP2
  - instruction:
    - mnemonic: mov
    - operand[0].offset: 4
    - or:
      - operand[1].number: COMMAND1
      - operand[1].number: COMMAND2
```

i think its appropriate to set the scope to basic block, too.

williballenthin:
image

mr-tz (approved):

Still34:
Throw this timestamp in there as well `0x20140613`, also I could not get the match to pass with sample edfff708808ef3a5453dc7713d306b20 @ 0x1000A258 (and yes the timestamp is added to the local rule).

mr-tz:
with the added value it matches for me:
image

malware-family/plugx/match-known-plugx-module.yml
```diff
@@ -6,73 +6,80 @@ rule:
...
-        - number: 0xD000 = get TCP table
-        - number: 0xD001 = get UDP table
-        - number: 0xD002 = set TCP entry
-        - number: 0xE000 = start keylogger thread
+        - mnemonic: mov
+        - operand[0].offset: 0
+        - or:
+          - operand[1].number: 0x20120225
+          - operand[1].number: 0x20120324
+          - operand[1].number: 0x20121107
```
mr-tz, line 25:
```suggestion
          - operand[1].number: 0x20121107
          - operand[1].number: 0x20140613
```

williballenthin:
@Still34 
> also I could not get the match to pass 

would you double check using -vv?

it looks like today we're filtering out malware family rules from the default display. im not quite sure why we're doing that. do you remember @mr-tz?

mr-tz:
I'm not sure why. Maybe we wanted to render it separately...

PR #847, opened 2023-11-20
updated .NET bundled rule and file limitation

sara-rn:
added 2 new rules, one in internal/limitation/file and another one in executable/dotnet-singlefile instead of installer/dotnet. Existing rule in installer/dotnet has been also moved to executable/dotnet-singlefile.
This way there is a specific file limitation rule for .net single-file application applications and the description is more accurate.


executable/dotnet-singlefile/bundled-with-dotnet-single-file-deployment.yml
```diff
@@ -0,0 +1,19 @@
+rule:
+  meta:
+    name: bundled with .NET single-file deployment
+    namespace: executable/dotnet-singlefile
```
mr-tz, line 4: how about `executable/dotnet/singlefile` or `executable/dotnet/single-file`?

PR #852, opened 2023-11-28
suggest to run on dynamic trace for packed samples

mr-tz:


internal/limitation/file/internal-packer-file-limitation.yml
```diff
@@ -8,8 +8,9 @@ rule:
       This sample appears to be packed.
 
       Packed samples have often been obfuscated to hide their logic.
-      capa cannot handle obfuscation well. This means the results may be misleading or incomplete.
+      capa cannot handle obfuscation well using static analysis. This means the results may be misleading or incomplete.
       If possible, you should try to unpack this input file before analyzing it with capa.
+      Alternatively, run the sample in a supported sandbox and run capa on the analysis report to obtain dynamic analysis results.
```
williballenthin, line 13:
```suggestion
      Alternatively, run the sample in a supported sandbox and invoke capa against the report to obtain dynamic analysis results.
```

williballenthin (approved):

PR #859, opened 2023-12-09
Update README to showcase dynamic rules

yelhamer:
This PR replaces an old example rule in the documentation that still used the `scope` field, with a more appropriate one that showcases dynamic capa rules and is updated to the correct format.


mr-tz (approved):

mr-tz:
Thanks!

PR #860, opened 2023-12-09
Fix the dynamic flavor scope for allocate-or-change-rw-memory.yml

yelhamer:
As mentioned by @mike-hunhoff , the rule in its current format can match if the `number: 0x4 = PAGE_READWRITE` feature is present in another call in the same thread. Setting the rule's dynamic scope to `call` helps prevent this.

### DIFF
--- a/lib/allocate-memory.yml
+++ b/lib/allocate-memory.yml
@@ -7,7 +7,7 @@ rule:
     lib: true
     scopes:
       static: basic block
-      dynamic: thread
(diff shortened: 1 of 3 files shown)


yelhamer:
I've fixed two additional rules (allocate-memory.yml and change-memory-protection.yml) that were previously matching at thread level.

mike-hunhoff (approved):
great, thanks!

PR #861, opened 2023-12-13
Update reference-analysis-tools-strings.yml

ruppde:
fix 2 problems:

1. fix false positive by not hitting, if there was a word character before `ida`:
```
regex: /ida[gqtuw]?(\.exe)?$/i
- "@.didat" @ file+0x2F7

405d4c2ef7419bf265edef0fe86c8ba1ed634b10dccaaa0a6c6b953645598619
```

2. regex didn't match ida64.exe because it required one of the characters in the brackets.

### DIFF
--- a/anti-analysis/reference-analysis-tools-strings.yml
+++ b/anti-analysis/reference-analysis-tools-strings.yml
@@ -24,8 +24,8 @@ rule:
       - string: /procmon(\.exe)?/i
       - string: /regmon(\.exe)?/i
       - string: /procexp(\.exe)?/i
-      - string: /ida[gqtuw]?(\.exe)?$/i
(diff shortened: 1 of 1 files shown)


williballenthin:
1. `- "@.didat" @ file+0x2F7` is this actually a reference to the IDA filename, or a FP?

williballenthin:
more generally, the four character pattern "ida[t...]" seems like it might very easily show up in random strings. should we restrict it further, like ensuring its not found within a substring, but only a whole string?

ruppde:
yes, fp.

I think the $ in the end prevents it from hitting any random substring.

mr-tz (approved):
LGTM as is

mr-tz:
thank you, @ruppde!

PR #862, opened 2023-12-15
Update Mappings for MBC (Q4 2023)

ryantxu1:
Note: We have introduced edits to move Archive Collected Data [E1560] to the Collection Objective in MBC

PR #874, opened 2024-01-17
Revert "Update Mappings for MBC (Q4 2023)"

mr-tz:
Reverts mandiant/capa-rules#862


mr-tz:
@ryantxu1 doesn't seem to be codified in the

PR #878, opened 2024-02-05
add delete drive layout via ioctl

### LINKED ISSUE #877: delete drive layout via IOCTL
> execute the Windows API function DeviceIoControl with the control code IOCTL_DISK_DELETE_DRIVE_LAYOUT, the effect of which would be to remove the boot signature from the master boot record, so that the disk will be formatted from sector zero to the end of the disk. 

also references here:
  - https://www.crowdstrike.com/blog/the-anatomy-of-wiper-malware-part-3/
  - https://tyleo.github.io/sharedlib/doc/winapi/winioctl/constant.IOCTL_DISK_DELETE_DRIVE_LAYOUT.html

> [IOCTL_DISK_DELETE_DRIVE_LAYOUT](http://msdn.microsoft.com/en-us/library/aa365158(v=vs.85).aspx)	0x7c100		Removes the boot signature from the master boot record, so that the disk will be formatted from sector zero to the end of the disk. Partition information is no longer stored in sector zero.

williballenthin:
closes #877

sample in https://github.com/mandiant/capa-testfiles/pull/226

### DIFF
--- a/impact/wipe-disk/delete-drive-layout-via-ioctl.yml
+++ b/impact/wipe-disk/delete-drive-layout-via-ioctl.yml
@@ -1,0 +1,25 @@
+rule:
+  meta:
+    name: delete drive layout via IOCTL
+    namespace: impact/wipe-disk
+    authors:
+      - william.ballenthin@mandiant.com
+    scopes:
+      static: basic block
+      dynamic: thread
+    att&ck:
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
LGTM once tests pass

williballenthin:
image

PR #879, opened 2024-02-05
add rules for volume interaction via IOCTLs

williballenthin:
from https://www.crowdstrike.com/blog/the-anatomy-of-wiper-malware-part-3/


host-interaction/driver/interact-with-driver-via-control-codes.yml
mr-tz, line 1: should we standardize this to also use `via ioctl`?

williballenthin:
yeah let me refactor a few things here. i'll make `interact with driver via ioctl` the main place that we reference `DeviceIoControl` and move the driver load/unload to other rules.

nursery/set-system-properties.yml
```diff
@@ -0,0 +1,17 @@
+rule:
+  meta:
+    name: set system properties
```
mr-tz, line 3: this loads a file path / driver, see https://is.muni.cz/th/y8opj/Bc.pdf
rename the rule and file?

mr-tz (approved):
LGTM besides the name of one rule

nursery/set-system-properties.yml
williballenthin: ah, ok, i'll move that over to `install driver`

PR #882, opened 2024-02-23
[CI] Use badge in gist for rules number in README

Ana06:
This does not rely on `img.shields.io`, simplifies the workflow code, and remove the need for the noisy commits in capa-rules. We still need the commits in capa to sync the submodule. I'll send a PR in capa to use the new badge in a moment.

This requires giving the token gist persmissions, which I have already done. So it should work when we merge it.

We are using gist badges in [VM-Packages](https://github.com/mandiant/VM-Packages) since a few days, in case you want to see it working. We update the number of packages when adding a new package and the the daily package failures once a day.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -8,21 +8,19 @@ jobs:
     steps:
     - name: Checkout capa-rules
       uses: actions/checkout@v3
-    - name: Update rules number badge in README
+    - name: Count rules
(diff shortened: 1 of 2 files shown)


williballenthin (approved):
nice simplification!

mr-tz (approved):
LGTM besides the commit hash pinning

PR #891, opened 2024-04-19
reduce FPs by adjusting `go1.` substring feature

mr-tz:
`go1.` would for example match on `logo1.jpg`

### DIFF
--- a/compiler/go/compiled-with-go.yml
+++ b/compiler/go/compiled-with-go.yml
@@ -14,7 +14,7 @@ rule:
       - string: /^Go build ID:/
       - substring: "go.buildid"
       - string: /^Go buildinf:/
-      - substring: "go1."
+      - string: /go1.\d+./
       - substring: "runtime.main"
       - substring: "main.main"
       - substring: "runtime.gcWork"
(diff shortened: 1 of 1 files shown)


compiler/go/compiled-with-go.yml
```diff
@@ -14,7 +14,7 @@ rule:
       - string: /^Go build ID:/
       - substring: "go.buildid"
       - string: /^Go buildinf:/
-      - substring: "go1."
+      - string: /go1.\d+./
```
mr-tz, line 17:
```suggestion
      - string: /go1.\d+/
```

compiler/go/compiled-with-go.yml
```diff
@@ -14,7 +14,7 @@ rule:
       - string: /^Go build ID:/
       - substring: "go.buildid"
       - string: /^Go buildinf:/
-      - substring: "go1."
+      - string: /go1.\d+/
```
mr-tz, line 17: or maybe better: `\bgo1.`?

`\b` == assert position at a word boundary

PR #892, opened 2024-04-23
add android OS where applicable

mr-tz:
I first started to create (copy/paste) new rules `... on Android` but that duplicates a lot and Android is based on the Linux kernel so I think the names are fine (but happy to discuss).


williballenthin (approved):
very reasonable, i agree

PR #893, opened 2024-04-23
add API features for ws2_32 ordinals

williballenthin:
Programs can import networking routines from ws2_32 by ordinal, and this is occasionally seen in the while, such as in PMA 01-01.dll.

This file shows an example of mapping ordinals to names for ws2_32: https://github.com/phracker/HopperScripts/blob/9468cdadb2c139d474662ae82716a5098e7350e4/WS2_32.dll%20Ordinals%20to%20Names.py#L8

Today, we rely on vivisect (or the binary analysis backend) to resolve ordinal numbers to human-readable names; however, not all analysis backends may have this mapping. So, we can extend our rules to also match these ordinals.

For example, with these changes using a backend that doesn't support ordinal mapping:

image

### DIFF
--- a/communication/dns/resolve-dns.yml
+++ b/communication/dns/resolve-dns.yml
@@ -16,16 +16,19 @@ rule:
       - 17264e3126a97c319a6a0c61e6da951e:0x5FDC25D0
   features:
     - or:
+      - api: gethostbyname
       - api: ws2_32.gethostbyname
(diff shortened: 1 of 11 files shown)


mr-tz (approved):

mr-tz:
Good idea. Alternatively, we'd have to add special logic to translate names to ordinals and vice versa.

PR #896, opened 2024-05-03
Add additional shellcode execution callback functions

Still34:
## Summary

This PR adds more APIs related to shellcode execution via callbacks for the existing rule.


mr-tz (approved):
awesome, thank you!

PR #902, opened 2024-05-31
Add two new CAPA rules: act-as-share-provider-dll.yml and act-as-windbg-extension.yml

jtothej:
Add two new CAPA rules: act-as-share-provider-dll.yml and act-as-windbg-extension.yml


williballenthin (approved):
nice!

mr-tz (approved):
thanks!

PR #908, opened 2024-06-06
fix the scope of some rules with dependencies

williballenthin:
ref: https://github.com/mandiant/capa/issues/2124

### DIFF
--- a/linking/runtime-linking/link-function-at-runtime-on-windows.yml
+++ b/linking/runtime-linking/link-function-at-runtime-on-windows.yml
@@ -6,8 +6,8 @@ rule:
       - moritz.raabe@mandiant.com
       - michael.hunhoff@mandiant.com
     scopes:
-      static: function
-      dynamic: unsupported  # requires characteristic features
+      static: instruction
+      dynamic: call
     att&ck:
       - Execution::Shared Modules [T1129]
     examples:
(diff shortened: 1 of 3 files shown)


linking/runtime-linking/link-function-at-runtime-on-windows.yml
```diff
@@ -6,8 +6,8 @@ rule:
       - moritz.raabe@mandiant.com
       - michael.hunhoff@mandiant.com
     scopes:
-      static: function
-      dynamic: unsupported  # requires characteristic features
+      static: instruction
+      dynamic: call
```
mr-tz, line 10: with the optional part we're targeting multiple API calls which won't work this way

williballenthin: ah, right...

I'm inclined to remove the `optional` and let this be used as a dependency more widely. Thoughts?

williballenthin: or relax scope to `basic block`

mike-hunhoff (commented):
Looks good pending @mr-tz comments, thanks!

linking/runtime-linking/link-function-at-runtime-on-windows.yml
mr-tz: removed in 0c7d1bdfacac57cc6480b1e9f2e60019c1a130c7

PR #912, opened 2024-07-17
Add capa rules create-thread-bypass-freeze.yml and check-thread-suspend-count-exceeded.yml to nursery.

ejfocampo:
Adding two capa rules - create-thread-bypass-freeze.yml and check-thread-suspend-count-exceeded.yml to the nursery. I did not find any public malware sample exhibiting the technique but I did test it with a basic POC code I created based on the referenced articles.

### DIFF
--- a/nursery/check-thread-suspend-count-exceeded.yml
+++ b/nursery/check-thread-suspend-count-exceeded.yml
@@ -1,0 +1,27 @@
+rule:
+  meta:
+    name: check thread suspend count exceeded
+    authors:
+      - ervinocampo@google.com
+    scopes:
+      static: file
+      dynamic: unsupported #requires mnemonic feature
+    att&ck:
+      - Defense Evasion::Debugger Evasion [T1622]
+    mbc:
+      - Anti-Behavioral Analysis::Debugger Detection [B0001]
+    references:
+      - https://secret.club/2021/01/04/thread-stuff.html
+      - https://research.checkpoint.com/2023/raspberry-robin-anti-evasion-how-to-exploit-analysis/
+  features:
+    - and:
+      - match: create thread bypass process freeze
+      - function:
+        - and:
+          - or:
+            - api: ntdll.NtSuspendThread
+            - string: "NtSuspendThread"
+          - basic block:
+            - and:
+              - number: 0xc000004a = STATUS_SUSPEND_COUNT_EXCEEDED
+              - mnemonic: cmp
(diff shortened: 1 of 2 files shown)


nursery/check-thread-suspend-count-exceeded.yml
```diff
@@ -0,0 +1,27 @@
+rule:
+  meta:
+    name: check thread suspend count exceeded
+    authors:
+      - ervinocampo@google.com
+    scopes:
+      static: file
```
williballenthin, line 7: could the scope be `function`?

ejfocampo: Thanks for reviewing the PR Willi. I opted for a file scope since the instructions being matched for "create thread bypass process freeze" will typically be on a separate function as the rest of the instructions being matched. For the simplest implementation of this, the call to NtCreateThreadEx parameter will be on the main function, while the call to NtSuspendThread will be on the function passed to NtCreateThreadEx.

nursery/create-thread-bypass-freeze.yml
```diff
@@ -0,0 +1,23 @@
+rule:
+  meta:
+    name: create thread bypass process freeze
```
williballenthin, line 3:
```suggestion
    name: create thread bypassing process freeze
```

nursery/check-thread-suspend-count-exceeded.yml
```diff
@@ -0,0 +1,27 @@
...
+    att&ck:
+      - Defense Evasion::Debugger Evasion [T1622]
+    mbc:
+      - Anti-Behavioral Analysis::Debugger Detection [B0001]
+    references:
+      - https://secret.club/2021/01/04/thread-stuff.html
+      - https://research.checkpoint.com/2023/raspberry-robin-anti-evasion-how-to-exploit-analysis/
+  features:
+    - and:
+      - match: create thread bypass process freeze
```
williballenthin, line 18:
```suggestion
      - match: create thread bypassing process freeze
```

williballenthin (approved):
see two inline suggestions, then lets merge!

nursery/create-thread-bypass-freeze.yml
williballenthin: oh, sorry, this will require a file rename, too.

ejfocampo:
Inline suggestions applied and file renamed to create-thread-bypassing-process-freeze.yml

williballenthin (approved):

mr-tz (approved):
thanks!

PR #913, opened 2024-07-18
Create upload-file-to-onedrive.yml

jaredscottwilson:
Please ensure that:
1. each rule passes thorough linting: It does
2. you've uploaded each referenced example binary: https://github.com/mandiant/capa-testfiles/pull/243


williballenthin (approved):
👏

PR #919, opened 2024-08-11
Update execute-shellcode-via-windows-callback-function.yml

jtothej:
Update execute-shellcode-via-windows-callback-function.yml - Add functions that can be used to execute shellcode


mr-tz (approved):

mr-tz:
Thank you!

PR #920, opened 2024-08-11
Add new rule linked-against-minhook.yml

jtothej:
Add new rule linked-against-minhook.yml

### DIFF
--- a/linking/static/minhook/linked-against-minhook.yml
+++ b/linking/static/minhook/linked-against-minhook.yml
@@ -1,0 +1,41 @@
+rule:
+  meta:
+    name: linked against MinHook
+    namespace: linking/static/minhook
+    authors:
+      - jakub.jozwiak@mandiant.com
+    scopes:
+      static: file
+      dynamic: file
+    att&ck:
+      - Defense Evasion::Hijack Execution Flow [T1574]
+    references:
+      - https://www.codeproject.com/Articles/44326/MinHook-The-Minimalistic-x-x-API-Hooking-Libra
+      - https://github.com/TsudaKageyu/minhook
+    examples:
+      - 3d5a99114b24e4373b77077072e4ed52ffce9e2607ec6fe4ed2f08d3b1c59026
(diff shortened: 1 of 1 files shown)


linking/static/minhook/linked-against-minhook.yml
```diff
@@ -0,0 +1,41 @@
...
+      - string: "MH_ERROR_NOT_CREATED"
+      - string: "MH_ERROR_ENABLED"
+      - string: "MH_ERROR_DISABLED"
+      - string: "MH_ERROR_NOT_EXECUTABLE"
+      - string: "MH_ERROR_UNSUPPORTED_FUNCTION"
+      - string: "MH_ERROR_MEMORY_ALLOC"
+      - string: "MH_ERROR_MEMORY_PROTECT"
+      - string: "MH_ERROR_MODULE_NOT_FOUND"
+      - string: "MH_ERROR_FUNCTION_NOT_FOUND"
+      - string: "MH_Initialize"
```
williballenthin, line 31: are any of these exports?

williballenthin (approved):

linking/static/minhook/linked-against-minhook.yml
jtothej: When MinHook is compiled as a standalone DLL some of the symbols are exported, e.g. https://github.com/TsudaKageyu/minhook/blob/c1a7c3843bd1a5fe3eb779b64c0d823bca3dc339/dll_resources/MinHook.def

mr-tz: do you think it's worth adding `export` features for these?

jtothej: Added `export` features in https://github.com/mandiant/capa-rules/pull/920/commits/d2d77916d71f11d59d51d4d761b9f8849cafec0b. I can imagine situations where these symbols are exported in executables compiled with MinHook.

PR #927, opened 2024-08-16
update linux/android rules and reduce fps for intel-specific rules

mike-hunhoff:


lib/calculate-modulo-256-via-x86-assembly.yml
```diff
@@ -15,6 +15,9 @@ rule:
     #  and ecx, 800000FFh
     #  and ecx, 0FFh
     - and:
+      - or:
+        - arch: i386
+        - arch: amd64
```
mike-hunhoff, line 20: `and` + `0xFF` appear very frequently in ARM resulting in FPs for dependent rules, so we limit this rule to Intel archs

mr-tz (approved):

PR #928, opened 2024-09-05
Move from nursery and update get-system-firmware-table.yml

jtothej:
Move from nursery and update `get-system-firmware-table.yml`


williballenthin (approved):
great!

PR #932, opened 2024-09-24
Add access-firewall-policy-via-inetfwpolicy2.yml and access-firewall-…

jtothej:
…rule-properties-via-inetfwrule.yml


mr-tz (approved):
thank you!

PR #935, opened 2024-09-25
Add rule for detecting hiding shutdown actions

Still34:
## Summary

This PR adds a rule that detects registry values related to hiding shutdown actions (i.e., logout, shutdown, reboot, etc.), specifically ones using policy registry values. This is often abused by ransomware to prevent mid-encryption-shutdowns.

### DIFF
--- a/host-interaction/os/hide-shutdown-actions-via-policy.yml
+++ b/host-interaction/os/hide-shutdown-actions-via-policy.yml
@@ -10,7 +10,9 @@ rule:
     att&ck:
       - Defense Evasion::Modify Registry [T1112]
     examples:
-      - a6594d9550d56ddeaac8b3140821e698eefb7163ba29f0119c2ef19beb6040b0
+      - a6594d9550d56ddeaac8b3140821e698eefb7163ba29f0119c2ef19beb6040b0:0x14000b47f
+    references:
+      - https://securelist.com/mallox-ransomware/113529/
   features:
     - and:
       - optional:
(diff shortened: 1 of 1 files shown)


Still34:
Let me know if this looks okay. I've opted to use case insensitive regex instead of substrings. If all looks good, I'll open a PR on the test data repo.

williballenthin (approved):
rule logic looks good 👍 

i'm not familiar with this technique though i can kinda guess about it based on the key names. are you able to find a reference or two that i could use to learn more (and we can include in the meta block)?

Still34:
image
Roughly what it may look like in some ransomware; this specific one was Mallox/TargetCompany.

mr-tz:
I agree on the reference, this appears to be one: https://securelist.com/mallox-ransomware/113529/

host-interaction/os/hide-shutdown-actions-via-policy.yml
```diff
@@ -0,0 +1,31 @@
...
+      - still@teamt5.org
+    scopes:
+      static: function
+      dynamic: call
+    att&ck:
+      - Defense Evasion::Modify Registry [T1112]
+    examples:
+      - a6594d9550d56ddeaac8b3140821e698eefb7163ba29f0119c2ef19beb6040b0:0x14000b47f
+    references:
+      - https://securelist.com/mallox-ransomware/113529/
```
mr-tz, line 15:
```suggestion
    references:
      - https://securelist.com/mallox-ransomware/113529/
    examples:
      - a6594d9550d56ddeaac8b3140821e698eefb7163ba29f0119c2ef19beb6040b0:0x14000b47f
```

mr-tz:
thank you!

PR #948, opened 2024-10-24
Add .NET AoT & TouchSocket linking detection 

Still34:
## Summary

This PR adds a new rule for detecting .NET binaries that are compiled Ahead-of-Time (AoT), as well as another rule for detecting strings related to the TouchSocket library, a Chinese-developed networking library that has been adopted by at least one Chinese APT group. 

## Related Sample

Additional discussion is required surrounding the example linked in the rule, as both of the rules feature a sample that is a known malicious sample that exceeds 15MB in size and can cause a significant amount of time to process, and may cause timeout for the CI pipeline. I can replace this with a self-written sample Hello World that would meet the requirements for both of these rules and be significantly smaller (~1.6MB) and faster, if other maintainers believe this would be a better approach.


mr-tz:
Great, once we have added that to the testfiles and tests pass, this is good to be merged!

PR #949, opened 2024-10-24
added detections for reading/writing shadow file, password database

jlmndt:
added detections for reading/writing shadow file, password database

### DIFF
--- a/nursery/get-password-database-entry-on-linux.yml
+++ b/nursery/get-password-database-entry-on-linux.yml
@@ -4,6 +4,7 @@ rule:
     namespace: host-interaction/session
     authors:
       - michael.hunhoff@mandiant.com
+      - jonathanlepore@google.com
     scopes:
       static: basic block
       dynamic: call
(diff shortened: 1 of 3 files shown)


nursery/get-password-database-entry-on-linux.yml
```diff
@@ -4,6 +4,7 @@ rule:
     namespace: host-interaction/session
     authors:
       - michael.hunhoff@mandiant.com
+      - jonathanlepore@google.com
     scopes:
       static: basic block
```
williballenthin, line 9: let's make this `instruction` and same with the other rules, since they each only look for a single call instruction.

nursery/get-shadow-password-file-entry-on-linux.yml
```diff
@@ -0,0 +1,21 @@
+rule:
+  meta:
+    name: get shadow password file entry on linux
```
williballenthin, line 3: please capitalize "Linux" in the rule metadata

williballenthin (changes requested):
thank you!

williballenthin (approved):

mr-tz (approved):

PR #952, opened 2024-10-31
Change dynamic scope to call for atomic operations

jorik-utwente:
To reduce false positives, we change the dynamic scope of a couple of operations to `call`.
E.g., writing to the registry is always done in one Windows API call, so we change the scope from `thread` to `call`.

Original discussion: https://github.com/mandiant/capa-rules/discussions/951

### DIFF
--- a/host-interaction/file-system/copy/copy-file.yml
+++ b/host-interaction/file-system/copy/copy-file.yml
@@ -7,7 +7,7 @@ rule:
       - michael.hunhoff@mandiant.com
     scopes:
       static: function
-      dynamic: thread
+      dynamic: call
     mbc:
       - File System::Copy File [C0045]
     examples:
--- a/host-interaction/file-system/write/write-file-on-windows.yml
+++ b/host-interaction/file-system/write/write-file-on-windows.yml
@@ -7,7 +7,7 @@ rule:
       - anushka.virgaonkar@mandiant.com
     scopes:
       static: function
-      dynamic: thread
+      dynamic: call
     mbc:
       - File System::Writes File [C0052]
     examples:
--- a/host-interaction/file-system/move/move-file.yml
+++ b/host-interaction/file-system/move/move-file.yml
@@ -7,7 +7,7 @@ rule:
       - michael.hunhoff@mandiant.com
     scopes:
       static: function
-      dynamic: thread
+      dynamic: call
     mbc:
       - File System::Move File [C0063]
     examples:
--- a/host-interaction/registry/create/set-registry-value.yml
+++ b/host-interaction/registry/create/set-registry-value.yml
@@ -7,7 +7,7 @@ rule:
       - michael.hunhoff@mandiant.com
     scopes:
       static: function
-      dynamic: thread
+      dynamic: call
     mbc:
       - Operating System::Registry::Set Registry Key [C0036.001]
     examples:


host-interaction/file-system/copy/copy-file.yml
```diff
@@ -26,8 +26,7 @@ rule:
...
             - api: kernel32.SHFileOperation
-      - call:
-        - and:
-          - number: 2 = FO_COPY
-          - or:
-            - api: kernel32.SHFileOperation
+      - and:
+        - number: 2 = FO_COPY
+        - or:
+          - api: kernel32.SHFileOperation
```
williballenthin, line 32: this changes the logic for static analysis: the number 2 and API SHFileOperation must be found anywhere within the function, rather than in a basic block (as the prior branch encodes). 

while the `call:` branch here would duplicate the scope declared at the top of the file, i think it still serves a purpose here. so i *think* these lines should be reverted.

williballenthin: i think this comment applies to the other files, too.

host-interaction/file-system/write/write-file-on-windows.yml
```diff
@@ -24,11 +24,10 @@ rule:
...
               - match: create or open file
-          - call:
-            - or:
-              - number: 0x40000000 = GENERIC_WRITE
-              - number: 0x2 = FILE_WRITE_DATA
-              - match: create or open file
+          - or:
+            - number: 0x40000000 = GENERIC_WRITE
+            - number: 0x2 = FILE_WRITE_DATA
+            - match: create or open file
```
williballenthin, line 30: here's where things can get tricky: the scope of `create or open file` is (fortunately) also `call`, but in some other rules is a larger scope (like `thread`), which is impossible to match.

so, i think this change is fine.

but we should be aware that in other scenarios, there's some tricky things we might have to fix.

host-interaction/file-system/copy/copy-file.yml
jorik-utwente: You are right, I will fix this.

jorik-utwente: This should be fixed now

host-interaction/file-system/copy/copy-file.yml
```diff
@@ -7,7 +7,7 @@ rule:
       - michael.hunhoff@mandiant.com
     scopes:
       static: function
-      dynamic: thread
+      dynamic: call
```
mr-tz, line 10: our rule parser doesn't like a call subscope in a file-wide call scope: `call subscope supported only for the process and thread scopes` - that's why the rule was as is, I assume

mr-tz: this works here:
```yml
rule:
  meta:
    name: copy file
    namespace: host-interaction/file-system/copy
    authors:
      - moritz.raabe@mandiant.com
    scopes:
      static: function
      dynamic: call
    mbc:
      - File System::Copy File [C0045]
    examples:
      - Practical Malware Analysis Lab 01-01.exe_:0x401440
  features:
    - or:
      - api: kernel32.CopyFile
      - api: kernel32.CopyFileEx
      - api: CopyFile2
      - api: CopyFileTransacted
      - api: LZCopy
      - api: System.IO.FileInfo::CopyTo
      - api: System.IO.File::Copy
      # static
      - basic block:
        - and:
          - number: 2 = FO_COPY
          - or:
            - api: kernel32.SHFileOperation
      # dynamic
      - and:
        - number: 2 = FO_COPY
        - or:
          - api: kernel32.SHFileOperation

```

mr-tz:
Thanks for all your suggestions, @jorik-utwente! I'm back this week to work on getting the improvements merged.

host-interaction/file-system/copy/copy-file.yml
mr-tz: Oh, yes, that's not what we want.
I think we'd have to change the rule parser to properly handle this. @williballenthin, what do you think?

williballenthin: This makes sense to me. +1

mr-tz: https://github.com/mandiant/capa/pull/2490

mr-tz: https://github.com/mandiant/capa/pull/2490 merged, so we can create rules with call/call scopes now.

mr-tz (approved):
thank you!!

PR #954, opened 2024-11-05
Add 38 new registry-based persistence techniques

jorik-utwente:
Add 38 new registry-based persistence techniques.
All new rules have references to online resources describing the technique.

This PR requires https://github.com/mandiant/capa-rules/pull/952 to be merged.

I am planning to add examples later. Would it be preferred to have POC implementations, or real-world malware that uses the technique?


mr-tz:
> Would it be preferred to have POC implementations, or real-world malware that uses the technique?

We currently have both, most test files are actual malware samples though.

nursery/persist-via-cor_profiler_path-registry-value.yml
```diff
@@ -0,0 +1,20 @@
...
+    att&ck:
+      - Persistence::Hijack Execution Flow::COR_PROFILER [T1574.012]
+    references:
+      - https://redcanary.com/blog/threat-detection/cor_profiler-for-persistence/
+  features:
+    - and:
+      - match: set registry value
+      - or:
+        - string: /Environment/i
+        - string: /SYSTEM\\(CurrentControlSet|ControlSet001)\\Control\\Session Manager\\Environment/i
```
mr-tz, line 19: simplify here to only use `/Environment/i` or remove it?

nursery/persist-via-default-file-association-registry-key.yml
```diff
@@ -0,0 +1,20 @@
...
+    namespace: persistence/registry
+    authors:
+      - j.j.vannielen@utwente.nl
+    scopes:
+      static: function
+      dynamic: call
+    att&ck:
+      - Persistence::Event Triggered Execution::Change Default File Association [T1546.001]
+    references:
+      - https://woshub.com/managing-default-file-associations-in-windows-10/
```
mr-tz, line 13: is this the right reference? it seems to talk about other registry paths/options

nursery/persist-via-default-file-association-registry-key.yml
```diff
@@ -0,0 +1,20 @@
...
+  features:
+    - and:
+      - match: set registry value
+      - or:
+        - string: /file\\shell\\open\\command/i
+        - string: /file\\shell\\print\\command/i
+        - string: /file\\shell\\printto\\command/i
```
mr-tz, line 20: can you share an example entry here? maybe we remove the leading `file` (i.e. it could be `*`)?

nursery/persist-via-disk-cleanup-handler-registry-key.yml
```diff
@@ -0,0 +1,20 @@
...
+      - Persistence::Event Triggered Execution [T1546]
+    references:
+      - https://www.hexacorn.com/blog/2018/09/02/beyond-good-ol-run-key-part-86/
+      - https://learn.microsoft.com/en-us/windows/win32/lwef/disk-cleanup
+  features:
+    - and:
+      - match: set registry value
+      - string: /Microsoft\\Windows\\CurrentVersion\\Explorer\\VolumeCaches\\/i
+      - optional:
+        - match: persist via COM hijack
```
mr-tz, line 20: is this relevant here or can be removed?

nursery/persist-via-dotnet_startup_hooks-registry-key.yml
```diff
@@ -0,0 +1,23 @@
...
+  features:
+    - or:
+      - and:
+        - match: set registry value
+        - string: /SYSTEM\\(CurrentControlSet|ControlSet001)\\Control\\Session Manager\\Environment/i
+        - string: /DOTNET_STARTUP_HOOKS/i
+      - and:
+        - match: set registry value
+        - string: /Environment/i
+        - string: /DOTNET_STARTUP_HOOKS/i
```
mr-tz, line 23: simplify here, as the second and block would always hit if the first one does?

nursery/persist-via-filter-handlers-registry-key.yml
```diff
@@ -0,0 +1,19 @@
...
+    att&ck:
+      - Persistence::Event Triggered Execution [T1546]
+    references:
+      - https://learn.microsoft.com/en-us/windows/win32/search/-search-ifilter-about
+  features:
+    - and:
+      - match: set registry value
+      - string: /\\\.(.*?)\\Handler/i
+      - optional:
+        - match: persist via COM hijack
```
mr-tz, line 19: would this match in the same call? also is it related or can be removed?

nursery/persist-via-path-registry-key.yml
```diff
@@ -0,0 +1,20 @@
...
+    att&ck:
+      - Persistence::Hijack Execution Flow::Path Interception by PATH Environment Variable [T1574.007]
+    references:
+      - https://attack.mitre.org/techniques/T1574/007/
+  features:
+    - and:
+      - match: set registry value
+      - or:
+        - string: /SYSTEM\\(CurrentControlSet|ControlSet001)\\Control\\Session Manager\\Environment/i
+        - string: /Environment/i
```
mr-tz, line 19: simplify like mentioned above?

nursery/persist-via-task-scheduler.yml
```diff
@@ -0,0 +1,26 @@
...
+      - and:
+        - match: set registry value
+        - string: /Microsoft\\Windows NT\\CurrentVersion\\Schedule\\TaskCache\\/i
+        - string: /^Actions$/i
+      - and:
+        - match: host-interaction/process/create
+        - string: /schtasks(|\.exe) /i
+        - or:
+          - string: /\/change/i
+          - string: /\/create/i
```
mr-tz, line 26: see rule `schedule task via schtasks`, let's adjust there?

nursery/persist-via-universal-app-uri-registry-key.yml
```diff
@@ -0,0 +1,17 @@
...
+      static: function
+      dynamic: call
+    att&ck:
+      - Persistence::Event Triggered Execution [T1546]
+    references:
+      - https://giuliocomi.blogspot.com/2019/10/abusing-windows-10-narrators-feedback.html
+  features:
+    - and:
+      - match: set registry value
+      - string: /Classes\\App.*\\Shell\\open\\command/i
```
mr-tz, line 17: combine this with the above rule and just use the `shell\\open\\command` parts?

nursery/persist-via-userinitmprlogonscript-registry-value.yml
```diff
@@ -0,0 +1,20 @@
...
+    att&ck:
+      - Persistence::Boot or Logon Initialization Scripts::Logon Script (Windows) [T1037.001]
+    references:
+      - https://attack.mitre.org/techniques/T1037/001/
+  features:
+    - and:
+      - match: set registry value
+      - or:
+        - string: /SYSTEM\\(CurrentControlSet|ControlSet001)\\Control\\Session Manager\\Environment/i
+        - string: /Environment/i
```
mr-tz, line 19: see comments above

mr-tz (commented):
Wow, great enumeration of persistence mechanisms here. I've not seen most of them used in malware so I'm very curious to their prevalence in the wild.

As before comments and suggestions inline.

I'd also like to get another reviewer on all of these because these are quite big updates.

Thank you!

mr-tz:
Updates look great to me, thank you very much!

nursery/persist-via-print-monitors-registry-key.yml
```diff
@@ -0,0 +1,19 @@
...
+  meta:
+    name: persist via Print Monitors registry key
+    namespace: persistence/registry
+    authors:
+      - j.j.vannielen@utwente.nl
+    scopes:
+      static: function
+      dynamic: call
+    att&ck:
+      - Persistence::Boot or Logon Autostart Execution::Port Monitors [T1547.010]
```
mike-hunhoff, line 11: Should the rule name use _Port Monitors_ instead of _Print Monitors_?

mike-hunhoff (commented):
Wow, awesome work @jorik-utwente ! These updates look good but I did notice that the `dynamic` scope should be `thread` instead of `call` for rules that match against both the registry subkey and value, as the subkey and value strings are likely going to be referenced in separate function calls.

That raises another issue with capa's existing rule _set registry value_. The `dynamic` scope for this rule should also be `thread` instead of `call` otherwise the [`optional` match for _create or open registry key_](https://github.com/mandiant/capa-rules/blob/ed816a8e53446cd21a9d634b7d0531df664ba1c1/host-interaction/registry/create/set-registry-value.yml#L22) can never match (it's `optional` so it's not show stopper). Thoughts @mr-tz ?

nursery/persist-via-print-monitors-registry-key.yml
jorik-utwente: Thanks for your input, @mike-hunhoff !

To my understanding, the cape report always contains the registry path and value name within one registry write function call.

E.g.
```
         {
            ...
            "category": "registry",
            "api": "RegSetValueExW",
            "status": true,
            "return": "0x00000000",
            "arguments": [
              {
                "name": "Handle",
                "value": "0x00000228"
              },
              {
                "name": "ValueName",
                "value": "DefaultNotificationsSetting"
              },
              {
                "name": "Type",
                "value": "4",
                "pretty_value": "REG_DWORD"
              },
              {
                "name": "Buffer",
                "value": "1"
              },
              {
                "name": "BufferLength",
                "value": "4"
              },
              {
                "name": "FullName",
                "value": "HKEY_LOCAL_MACHINE\\SOFTWARE\\Policies\\Google\\Chrome\\DefaultNotificationsSetting"
              }
            ],
            ...
          },
```

In what case do the registry path and value name spread out over multiple Windows API calls?

mike-hunhoff: I'm not familiar with the underlying representation used by CAPE for API calls, but it appears that it may be adding some syntactic sugar e.g. _FullName_ is a field specific to CAPE.

Other sandboxes may not do this, e.g. VMRay:
```xml
<fncall ts="126662" fncall_id="28904" process_id="1" thread_id="1" name="RegOpenKeyExW" addr="0x7ff8e7439670" from="0x7ff8a1570edf">
	<in>
		<param name="hKey" type="void_ptr" value="0xffffffff80000001"/>
		<param name="lpSubKey" type="ptr" value="0x2c7d58c">
			<deref type="str" value="Software\\Microsoft\\Windows\\CurrentVersion\\Run"/>
		</param>
		<param name="ulOptions" type="unsigned_32bit" value="0x0"/>
		<param name="samDesired" type="unknown" value="0x2001f"/>
		<param name="phkResult" type="void_ptr" value="0xbbdb78"/>
	</in>
	<out>
		<param name="phkResult" type="ptr" value="0xbbdb78">
			<deref type="unsigned_64bit" value="0x75c"/>
		</param>
		<param name="ret_val" type="unknown" value="0x0"/>
	</out>
</fncall>
[...]
<fncall ts="126664" fncall_id="28906" process_id="1" thread_id="1" name="RegSetValueExW" addr="0x7ff8e743b080" from="0x7ff8a1574d2c">
	<in>
		<param name="hKey" type="void_ptr" value="0x75c"/>
		<param name="lpValueName" type="ptr" value="0x2c7b5b4">
			<deref type="str" value="Windows Defender Updater"/>
		</param>
		<param name="Reserved" type="unsigned_32bit" value="0x0"/>
		<param name="dwType" type="unsigned_32bit" value="0x1"/>
		<param name="lpData" type="ptr" value="0x2c7d33c">
[...]
		</param>
		<param name="cbData" type="unsigned_32bit" value="0x7a"/>
	</in>
[...]
</fncall>
```

You can see the subkey `Software\Microsoft\Windows\CurrentVersion\Run` and value `Windows Defender Updater` are referenced in separate API calls `RegOpenKeyExW` and `RegSetValueExW`, respectively (`0x75c` is the handle to the opened registry subkey).

Although this does not appear to be the case for CAPE, we assume most sandboxes will log API calls, including parameter and return names and values, according to their documentation, e.g. https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regsetvalueexw.

@mr-tz do you know if the CAPE extractor is extracting the _FullName_ field?

mr-tz: I think CAPE is doing some enhancements to the API trace by enhancing values. We cannot assume this for all sandbox backends so to be flexible rules should account for that.

williballenthin: Good discussion here!

> For much better handling moving forward we need a different approach (like a new scope within a certain range, https://github.com/mandiant/capa-rules/discussions/951#discussioncomment-11112001).

Agree.

> I expect that for some rules, changing the scope from call to threat will result in many false positives. I think that as a result we will have to remove these rules, which is a pitty.

In spirit, I agree with you. Though, I'm curious about this in practice. It *seems* like there might be FPs, but could we test this a bit? Maybe the behavior isn't quite as bad as we worry about.

I only suggest this (which doesn't *feel* very good) because we think we have a plan for a better implementation (#951) so many a little inaccuracy in the meantime is acceptable.

williballenthin: > That raises another issue with capa's existing rule set registry value. The dynamic scope for this rule should also be thread instead of call otherwise the [optional match for create or open registry key](https://github.com/mandiant/capa-rules/blob/ed816a8e53446cd21a9d634b7d0531df664ba1c1/host-interaction/registry/create/set-registry-value.yml#L22) can never match (it's optional so it's not show stopper). Thoughts @mr-tz ?

@mike-hunhoff https://github.com/mandiant/capa/issues/2124 raises the issue that capa doesn't validate scope dependencies correctly, so there are probably more of these hidden mistakes lurking around.

In the meantime, I agree this isn't a showstopper, since it's only an optional term that can't match. We have a bit of work to do to get all these scopes straightened out...

mr-tz (approved):
Willi and I discussed this offline:
* even if there are some minor issues, we can address those when we have better thread/region/call scope
* if we run into many FPs (or any FNs) we can adjust on a case by case basis
* let's see how common these are in practice

PR #956, opened 2024-11-05
Add 4 miscellaneous persistence techniques

jorik-utwente:
Add 4 miscellaneous persistence techniques.

The BITS Jobs and WMI subscription rules may result in false positives because of the `ole32.CoCreateInstance` matching.
We might want to remove that part, but that would result in false negatives. Thoughts on this are welcome! :)

This PR requires https://github.com/mandiant/capa-rules/pull/952 to be merged.


nursery/persist-via-bits-job.yml
```diff
@@ -0,0 +1,29 @@
...
+      dynamic: call
+    att&ck:
+      - Persistence::BITS Jobs [T1197]
+    references:
+      - https://cloud.google.com/blog/topics/threat-intelligence/attacker-use-of-windows-background-intelligent-transfer-service/
+  features:
+    - or:
+      - and:
+        - api: ole32.CoCreateInstance
+        - string: "4991d34b-80a1-4291-83b6-3328366b9097" # BITS
```
mr-tz, line 18: `"BackgroundCopyManager": ["4991D34B-80A1-4291-83B6-3328366B9097"]` per our com classes, consider using https://github.com/mandiant/capa-rules/blob/master/doc/format.md#com

nursery/persist-via-application-shimming.yml
```diff
@@ -0,0 +1,25 @@
...
+    authors:
+      - j.j.vannielen@utwente.nl
+    scopes:
+      static: function
+      dynamic: call
+    att&ck:
+      - Persistence::Event Triggered Execution::Application Shimming [T1546.011]
+    references:
+      - https://cloud.google.com/blog/topics/threat-intelligence/fin7-shim-databases-persistence/
+  features:
```
mr-tz, line 14: should we add `sdbinst.exe` usage here?

nursery/persist-via-application-shimming.yml
```diff
@@ -0,0 +1,25 @@
...
+      - and:
+        - match: set registry value
+        - string: /Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\InstalledSDB\\/i
+        - string: /DatabasePath/i
+      - and:
+        - or:
+          - match: copy file
+          - match: move file
+          - match: host-interaction/file-system/write
+        - string: /.sdb/i
```
mr-tz, line 25: I'm not sure about these features, do you have an example sample handy?

nursery/persist-via-application-shimming.yml
```diff
@@ -0,0 +1,25 @@
...
+      dynamic: call
+    att&ck:
+      - Persistence::Event Triggered Execution::Application Shimming [T1546.011]
+    references:
+      - https://cloud.google.com/blog/topics/threat-intelligence/fin7-shim-databases-persistence/
+  features:
+    - or:
+      - and:
+        - match: set registry value
+        - string: /Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\InstalledSDB\\/i
```
mr-tz, line 18: what about the `Custom` entry?

nursery/persist-via-application-shimming.yml
```diff
@@ -0,0 +1,25 @@
+rule:
+  meta:
+    name: persist via application shimming
+    namespace: persistence/file-system-and-registry
```
mr-tz, line 4:
```suggestion
    namespace: persistence
```

mr-tz: IMHO fine here

nursery/persist-via-bits-job.yml
```diff
@@ -0,0 +1,29 @@
+rule:
+  meta:
+    name: persist via BITS job
+    namespace: persistence/custom-db
```
mr-tz, line 4:
```suggestion
    namespace: persistence
```

nursery/persist-via-wmi-event-subscription.yml
```diff
@@ -0,0 +1,24 @@
...
+    att&ck:
+      - Persistence::Event Triggered Execution::Windows Management Instrumentation Event Subscription [T1546.003]
+    references:
+      - https://www.blackhat.com/docs/us-15/materials/us-15-Graeber-Abusing-Windows-Management-Instrumentation-WMI-To-Build-A-Persistent%20Asynchronous-And-Fileless-Backdoor-wp.pdf
+      - https://cloud.google.com/blog/topics/threat-intelligence/dissecting-one-ofap/
+  features:
+    - or:
+      - and:
+        - api: ole32.CoCreateInstance
+        - string: "4590F811-1D3A-11D0-891F-00AA004B2E24" # IWbemLocator
```
mr-tz, line 19: see com comment above

mr-tz (commented):
very cool, see my initial feedback attached

nursery/persist-via-application-shimming.yml
jorik-utwente: The idea behind this part is that malware can look for existing shim files and overwrite them.

nursery/persist-via-application-shimming.yml
jorik-utwente: We could add that as an optional field. The malware can gain persistence by only overwriting the InstalledSDB of an existing entry

nursery/persist-via-application-shimming.yml
mr-tz: I see, would you please add a comment to this part directly to the rule?

nursery/persist-via-application-shimming.yml
mr-tz: ok, if `Custom` is not a indicator by itself, let's leave as is

nursery/persist-via-print-processors-registry-key.yml
```diff
@@ -0,0 +1,25 @@
...
+      - and:
+        - match: set registry value
+        - string: /SYSTEM\\(CurrentControlSet|ControlSet001)\\Control\\Print\\Environments\\.*\\Print Processors\\/i
+        - string: /^Driver$/i
+      - and:
+        - or:
+          - match: copy file
+          - match: move file
+          - match: host-interaction/file-system/write
+        - string: /\\spool\\PRTPROCS\\/i
```
mr-tz, line 25: with the `dynamic: call` scope this would only work for static processing, is that intended?

jorik-utwente: It is intended to work for dynamic as well.
When https://github.com/mandiant/capa-rules/pull/952 is merged, this should work for dynamic as well (as it changes the scope of `copy file`, `move file`, and `write file to the `call` scope).
Or is there another reason why it wouldn't work for dynamic analysis?

mr-tz: nope, you're right, thanks!

nursery/persist-via-bits-job.yml
```diff
@@ -0,0 +1,29 @@
...
+      dynamic: call
+    att&ck:
+      - Persistence::BITS Jobs [T1197]
+    references:
+      - https://cloud.google.com/blog/topics/threat-intelligence/attacker-use-of-windows-background-intelligent-transfer-service/
+  features:
+    - or:
+      - and:
+        - api: ole32.CoCreateInstance
+        - com/class: BackgroundCopyManager # 4991d34b-80a1-4291-83b6-3328366b9097
```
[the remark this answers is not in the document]
jorik-utwente: There could be false positives here indeed, but from my experience malware mostly uses BITS for persistence. So in this case I would argue the COM usage is enough

nursery/persist-via-wmi-event-subscription.yml
```diff
@@ -0,0 +1,24 @@
...
+    att&ck:
+      - Persistence::Event Triggered Execution::Windows Management Instrumentation Event Subscription [T1546.003]
+    references:
+      - https://www.blackhat.com/docs/us-15/materials/us-15-Graeber-Abusing-Windows-Management-Instrumentation-WMI-To-Build-A-Persistent%20Asynchronous-And-Fileless-Backdoor-wp.pdf
+      - https://cloud.google.com/blog/topics/threat-intelligence/dissecting-one-ofap/
+  features:
+    - or:
+      - and:
+        - api: ole32.CoCreateInstance
+        - com/class: WbemLocator # 4590F811-1D3A-11D0-891F-00AA004B2E24
```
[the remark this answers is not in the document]
jorik-utwente: Malware also uses WMI for information gathering etc.. To reduce false positives we should remove it indeed. I don't see any way in Cape that we can use to reduce the false positives, as we can't match the communication with WMI.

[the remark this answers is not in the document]
mr-tz: agreed, for static analysis we could add more details, for dynamic I'd have to do more research myself to see what features are available

nursery/persist-via-bits-job.yml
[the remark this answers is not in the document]
mr-tz: per the reference, BITS could also be used to download a payload, for example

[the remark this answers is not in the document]
mr-tz: we could change the rule to detect BITS usage in general here

nursery/persist-via-wmi-event-subscription.yml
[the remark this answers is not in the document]
mr-tz: like above we could generalize the rule to detect WMI usage

nursery/persist-via-bits-job.yml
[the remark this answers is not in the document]
jorik-utwente: I removed the COM part in the latest commit. Now the rule looks for the cmd/powershell commands. This removes the false positives. As you said, we could add a rule in the future to detect BITS usage in general.

nursery/persist-via-wmi-event-subscription.yml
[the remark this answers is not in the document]
jorik-utwente: I moved this to the rule for WMI usage detection. Let me know what you think.

PR #961, opened 2024-11-13
Removed clipboard elements of capture-screenshot rule

mwilliams31:
Unclear why these were included.


williballenthin:
i *suspect* this technique uses a special-purpose key code (for those extended keyboards with lots of media buttons) that places the full screen screenshot into the clipboard. if that's the case, i think these match items should be retained. thoughts? input @re-fox ?

mwilliams31:
You're right, disregard. I did this in a hurry and missed that.

PR #969, opened 2024-12-02
extend rule features and rename

### LINKED ISSUE #966: Modify ProcessDynamicCodePolicy Value
[anti-analysis/anti-av/block-operations-on-executable-memory-pages-using-arbitrary-code-guard.yml](https://github.com/mandiant/capa-rules/blob/e50d824553ae8d7168ac2d1f7b5540a93f944e28/anti-analysis/anti-av/block-operations-on-executable-memory-pages-using-arbitrary-code-guard.yml)

The rule currently uses 8 for `ProcessDynamicCodePolicy`, but the correct value should be 2 according to documentation and observations. 
I kindly request an update to the rule. 

Thank you!

```cpp
typedef enum _PROCESS_MITIGATION_POLICY {
    ProcessDEPPolicy,
    ProcessASLRPolicy,
    ProcessDynamicCodePolicy, // 2
    ProcessStrictHandleCheckPolicy,
    ProcessSystemCallDisablePolicy,
    ProcessMitigationOptionsMask,
    ProcessExtensionPointDisablePolicy,
    ProcessControlFlowGuardPolicy,
    ProcessSignaturePolicy,
    ProcessFontDisablePolicy,
    ProcessImageLoadPolicy,
    ProcessSystemCallFilterPolicy,
    ProcessPayloadRestrictionPolicy,
    ProcessChildProcessPolicy,
    ProcessSideChannelIsolationPolicy,
    ProcessUserShadowStackPolicy,
    ProcessRedirectionTrustPolicy,
    ProcessUserPointerAuthPolicy,
    ProcessSEHOPPolicy,
    MaxProcessMitigationPolicy
} PROCESS_MITIGATION_POLICY, *PPROCESS_MITIGATION_POLICY;
```

mr-tz:
closes #966


mr-tz:
@ezbeat @jtothej please take a look

Ezbeat:
Thank you for taking my feedback into consideration.

1. Regarding the `ProcessSignaturePolicy` configuration, wouldn't it be clearer and more consistent to use `policy.MicrosoftSignedOnly` instead of `policy.flags`? This would enhance readability and align with the explicit naming convention used in other mitigation policies.

2. For the detection of `blockdlls`, rather than relying on the `InitializeProcThreadAttributeList` function, how about using the `UpdateProcThreadAttribute` function? This approach could directly focus on when the `Attribute` is actually set, potentially improving accuracy and reducing false positives.

mr-tz:
nice, great feedback here, thanks! does the rule now look good to you?

Ezbeat:
Looks good. Thank you for considering the suggestions!

mr-tz:
Oh :facepalm: @jtothej already did all the work... Thanks, for noticing. I'll update accordingly.

mr-tz:
Thanks, @Ezbeat (and @jtothej)!!

PR #978, opened 2025-01-06
update create-process-suspended to include DEBUG_ONLY_THIS_PROCESS

mike-hunhoff:
> With the Debug Injection method, the ransomware first uses the function “CreateProcessW” with the “DEBUG_ONLY_THIS_PROCESS” creation flag. Creating the “Explorer.exe” process in this state allows the ransomware to suspend the process and modify its memory. This is when then ransomware calls “NtMapViewOfSection” to copy itself to the process and fix its relocations. Afterward, it calls “NtGetContextThread” to receive the thread context of the main thread for “Explorer.exe”. It then sets the instruction pointer (EIP) to the mapped ransomware’s selected entry point. “ContinueDebugEvent” is then called to continue the execution of “Explorer.exe” and the following call is to “DebugActiveProcessStop” which finally detaches the ransomware’s debugger from the process.

_source: https://www.trustwave.com/en-us/resources/blogs/spiderlabs-blog/an-in-depth-look-at-mailto-ransomware-part-two-of-three/_

```
DEBUG_ONLY_THIS_PROCESS (0x00000002)
The calling thread starts and debugs the new process. It can receive all related debug events using the WaitForDebugEvent function.
```

### DIFF
--- a/host-interaction/process/create/create-process-suspended.yml
+++ b/host-interaction/process/create/create-process-suspended.yml
@@ -4,6 +4,7 @@ rule:
     namespace: host-interaction/process/create
     authors:
       - william.ballenthin@mandiant.com
+      - mehunhoff@google.com
     scopes:
       static: basic block
       dynamic: call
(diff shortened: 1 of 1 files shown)


williballenthin:
would you please add the documentation/reference links from this issue into the rule?

williballenthin (approved):

Issue #962, opened 2024-11-15
RSA & bigint


Ana06:
I have recently analysed a ransomware sample that uses the RSA encryption implementation from [github.com/ezhangle/krypton](https://github.com/ezhangle/krypton/blob/147d69429bfb03cce7113dca6dba36e77f8a9325/src/rsa.c#L232) (which is also similar to the implementation in [github.com/bnoordhuis/mongrel2](https://github.com/bnoordhuis/mongrel2/blob/3e9b57d82aeb627be0aebfb346199bfdfd67e530/src/crypto/rsa.c#L233)). I found tricky to identify it, until I found the `0x7FFF55AA = PERMANENT` constant that pointed me in the right direction.

## Issue to match RSA
So, I was wondering if we can create a capa rule for this type of RSA encryption. Characteristic code in the `RSA_encrypt` function:

- `- 3` -> substruct 3 to calculate pads needed
  - assembly code: `sub     eax, 3`
- `out_data[0] = 0` -> to ensure encryption block is < modulus
  - assembly code:  `mov     byte ptr [ecx], 0`
- `out_data[1] = 2`-> I think this is the encryption flag (signing would be 1)
  - assembly code:  `mov     byte ptr [edx+1], 2`
- `out_data[2+num_pads_needed] = 0;` -> zero terminated
  - assembly code: `mov     byte ptr [edx+2], 0`
- `get_random_nonzero(&out_data[2], num_pads_needed)` -> to randomize the encryption padding with non-zero bytes 

The problem is that with the currently supported capa features we can only match offsets and numbers and in this case they are common numbers (0,1,2,3). Although matching then together inside a `instruction` feature could be enough to prevent false positives. **@mandiant/capa-dev what do you think? Could this be enough?** 

The `0x7FFF55AA = PERMANENT` constant is in a subfunction (likely several levels), so extending capa to add subfunction support would likely allows us to match this implementation in a more robust way. **Is there already an issue in capa to support subfunction support? Could this be implemented and be useful for other cases as well?**

## Bigint

We could create capa rules for bigint functions, as identifying this functions may help reverse engineers identifying crypto. We can do this by creating a rule for every of the following three functions or a single generic rule for bigint functions. Note that the offset `8` and the `0x7FFF55AA = PERMANENT` constant in the same instruction is shared by all functions.

**@mandiant/capa-dev what do you think? Should I send a PR with tree rules (one per function) or a single rule that matches the three of them?**

### bi_copy

[...]

PR #982, opened 2025-01-20
New rules: RSA & bigint

Ana06:
- Add new rule `data-manipulation/encryption/rsa/rsa-encrypt.yml` that identifies the RSA encryption implementation from:
  - https://github.com/ezhangle/krypton/blob/147d69429bfb03cce7113dca6dba36e77f8a9325/src/rsa.c#L232
  - https://github.com/bnoordhuis/mongrel2/blob/3e9b57d82aeb627be0aebfb346199bfdfd67e530/src/crypto/rsa.c#L233
- Add new `data-manipulation/encryption/use-bigint-function` rule to identify bigint functions as these functions may help reverse engineers identifying crypto. Example of bigint functions identified by this rule: [`bi_copi`](https://github.com/ezhangle/krypton/blob/147d69429bfb03cce7113dca6dba36e77f8a9325/src/bigint.c#L149), [`bi_permanent`](https://github.com/ezhangle/krypton/blob/147d69429bfb03cce7113dca6dba36e77f8a9325/src/bigint.c#L161), [`bi_depermant`](https://github.com/ezhangle/krypton/blob/147d69429bfb03cce7113dca6dba36e77f8a9325/src/bigint.c#L177), [`bi_free`](https://github.com/ezhangle/krypton/blob/147d69429bfb03cce7113dca6dba36e77f8a9325/src/bigint.c#L196).

We need to add `009c2377b67997b0da1579f4bbc822c1` to the test files.

Closes https://github.com/mandiant/capa-rules/issues/962 :bowtie:


data-manipulation/encryption/rsa/rsa-encrypt.yml
```diff
@@ -0,0 +1,45 @@
+rule:
+  meta:
+    name: rsa encrypt
+    namespace: data-manipulation/encryption/rsa
+    authors:
+      - "Ana06"
+    scopes:
+      static: function
+      dynamic: unsupported
```
mike-hunhoff, line 9:
```suggestion
      dynamic: unsupported # requires mnemonic, offset features
```

mike-hunhoff: This sort of comment helps us with future rule maintenance

data-manipulation/encryption/rsa/rsa-encrypt.yml
```diff
@@ -0,0 +1,45 @@
...
+        - mnemonic: mov
+        - offset: 1
+        - number: 2
+      # `mov  byte ptr [edx+2], 0` Terminate with zero
+      - instruction:
+        - mnemonic: mov
+        - offset: 2
+        - number: 0
+      # call `get_random_nonzero`, `memcpy`, `bi_import`, `RSA_public`, `bi_export`, and `bi_clear_cache`
+      # if the signing code is included, also call `memcpy` and `RSA_private`
```
mike-hunhoff, line 42: All of these comments are awesome, thank you!

data-manipulation/encryption/rsa/rsa-encrypt.yml
```diff
@@ -0,0 +1,45 @@
+rule:
+  meta:
+    name: rsa encrypt
```
mike-hunhoff, line 3: Let's update this rule name to _encrypt data using RSA via YY_. If we don't have a good substitute for _YY_ then we should combine _encrypt data using RSA via WinAPI_ (below) with this rule under the name _encrypt data using RSA_.

mike-hunhoff (changes requested):
Great work @Ana06 , thank you! I've left a few minor comments for your review. I've also merged https://github.com/mandiant/capa-testfiles/pull/273.

data-manipulation/encryption/rsa/rsa-encrypt.yml
Ana06: I think it is better to keep them separated to make the rules easy to understand. I am not completely sure what a good _YY_ would be, what about _encrypt data using RSA via embedded library`? @mandiant/capa-dev any other ideas?

data-manipulation/encryption/use-bigint-function.yml
```diff
@@ -0,0 +1,30 @@
+rule:
+  meta:
+    name: use bigint function # such as bi_copy and bi_permanent. Useful to identify crypto.
```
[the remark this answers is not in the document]
Ana06: good point! :+1: I have forgotten the description field. I can also move most of the comments about the rsa encryption instructions to descriptions :thinking:

data-manipulation/encryption/rsa/rsa-encrypt.yml
mike-hunhoff: My first thought is _...via x86 assembly_ but with the calls included maybe _...via static library_ is a better fit?

mike-hunhoff (approved):
🚀

PR #984, opened 2025-01-26
remove duplicate features from some rules

v1bh475u:
removing duplicate features found in following rules using new lint (https://github.com/mandiant/capa/pull/257):
  - act as Office COM add-in
  - create TCP socket via raw AFD driver
  - reference anti-VM strings targeting VMWare
  - reference anti-VM strings targeting VirtualBox
  - reference anti-VM strings targeting Xen
  - send data to Internet

### DIFF
--- a/communication/socket/tcp/create-tcp-socket-via-raw-afd-driver.yml
+++ b/communication/socket/tcp/create-tcp-socket-via-raw-afd-driver.yml
@@ -41,6 +41,7 @@ rule:
             - number: 0x64 = d
             - number: 0x4F = O
             - number: 0x70 = p
+            # - number: 0x65 = e
             - number: 0x6E = n
             - number: 0x50 = P
             - number: 0x61 = a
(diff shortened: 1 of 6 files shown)


communication/socket/tcp/create-tcp-socket-via-raw-afd-driver.yml
```diff
@@ -41,7 +41,6 @@ rule:
             - number: 0x64 = d
             - number: 0x4F = O
             - number: 0x70 = p
```
williballenthin, line 43: let's comment these two out, rather than removing them, since the sequence is intended to show an inline structure ("AfdOpenPacketX")

williballenthin (changes requested):
This is really cool!

One easy request, and then ready to merge.

communication/socket/tcp/create-tcp-socket-via-raw-afd-driver.yml
v1bh475u: Ok. but the current lint will then fail for this rule. will that be fine?

v1bh475u: Done :heavy_check_mark:

williballenthin (approved):
thank you @v1bh475u!

PR #986, opened 2025-01-29
fix scope issues identified by new lint pass

williballenthin:
ref https://github.com/mandiant/capa/pull/2579

Most of these are obvious. But I did make one set of debatable changes: when a rule *can* have a really small scope, but we had an optional block to provide more context and therefore has a loose scope, i tightened the scope and removed the optional block. for example:

image

Otherwise, it gets kinda weird when a rule really just describes a single API call but has to have span scope, which then requires the rules that reference it to also have that looser scope.


mike-hunhoff (approved):
This looks great, thank you! I think removing the _optional_ statements is worth the trade offs. LGTM 🚀

PR #987, opened 2025-01-30
remove redundant matches for dynamic scope

mike-hunhoff:
This PR restricts _link function at runtime on Windows_ submatches to static scope to remove redundant matches for dynamic scope. Basically, some capa rules had logic to match both direct API calls and resolution of those APIs via dynamic linking. This results in redundant matches when matching dynamic scope because the logic matches both the dynamic link and the invocation. These usually match within a few traced calls of each other.

### DIFF
--- a/lib/change-memory-protection.yml
+++ b/lib/change-memory-protection.yml
@@ -5,7 +5,7 @@ rule:
       - "@mr-tz"
     lib: true
     scopes:
-      static: basic block
+      static: function
       dynamic: call
     mbc:
       - Memory::Change Memory Protection [C0008]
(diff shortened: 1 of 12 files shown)


lib/change-memory-protection.yml
```diff
@@ -5,7 +5,7 @@ rule:
       - "@mr-tz"
     lib: true
     scopes:
-      static: basic block
+      static: function
```
williballenthin, line 8: can you explain while the scopes here are relaxed? is it so there can be a basic block sub scope?

williballenthin (approved):

lib/change-memory-protection.yml
williballenthin: if so, i think maybe we can have sub scope blocks that are the same as the outer scope (i *think*). we should try.

mike-hunhoff: Yes, unfortunately this forces us to loosen the scopes of some of the rules and the rules that subsequently match them.

mike-hunhoff: > if so, i think maybe we can have sub scope blocks that are the same as the outer scope (i _think_). we should try.

I tried this at first, but capa outputs the following error: _ERROR    capa: invalid rule: `basic block` subscope supported only for `function` scope_

williballenthin: i think we might be able to relax this. i'll try it tomorrow.

williballenthin: see https://github.com/mandiant/capa/pull/2584. once we merge that, let's re-tighten the scopes in this PR.

williballenthin (approved):

PR #990, opened 2025-02-04
internal-dotnet-file-limitation.yml: make it valid for static scope too

williballenthin:
ref https://github.com/mandiant/capa/issues/2591

### DIFF
--- a/internal/limitation/dynamic/internal-dotnet-file-limitation.yml
+++ b/internal/limitation/dynamic/internal-dotnet-file-limitation.yml
@@ -10,7 +10,7 @@ rule:
       capa rules are not yet tuned for the .NET runtime, 
       so its analysis may be incomplete or misleading.
     scopes:
-      static: unsupported
+      static: file
       dynamic: file
     examples:
       - 2f8a79b12a7a989ac7e5f6ec65050036588a92e65aeb6841e08dc228ff0e21b4_min_archive.zip


mike-hunhoff (commented):
With this change, capa will print the description and force users to run capa with >= _-v_ option for every .NET file that it analyzes statically:
```
      This dynamic analysis trace describes a .NET file.

      capa rules are not yet tuned for the .NET runtime, 
      so its analysis may be incomplete or misleading.
```
This is confusing on two fronts:
- _This dynamic analysis trace describes a .NET file._ could confuse users attempting to analyze a .NET file statically
- _capa rules are not yet tuned for the .NET runtime_ isn't true for static analysis as capa has a robust set of .NET rules and we don't want to turn users away from using capa to process .NET files statically.

@williballenthin I agree with your comments in https://github.com/mandiant/capa/issues/2591#issuecomment-2633673829 that a proper fix likely requires many changes. Unfortunately, I think our best option at this time is to exclude this rule until we have time to implement these changes.

williballenthin:
> With this change, capa will print the description and force users to run capa with >= -v option for every .NET file that it analyzes statically:

Thanks for calling this out. That wasn't my intention for the logic of capa and the dynamic limitations. These dynamic limitations should only be used to warn users during dynamic analysis.

@v1bh475u would you reproduce this behavior on your side and open a new PR to ensure these rules aren't used to warn for samples in the wrong analysis mode.

@mike-hunhoff agree that in the short term we can remove the dynamic .NET rule

v1bh475u:
Sure. Will start working on it!

v1bh475u:
The issue is there are rigid matches for static file limitations, but for dynamic dotnet limitations, I couldn't find any firm constraints. Do you have any suggestions?

williballenthin:
@mike-hunhoff 

> However, the rule is matched and displayed in my local instance, which isn't great but not show stopping

This is fixed in the pending #2592, where we now hide any internal rule from the default output (but will show it during vverbose, which isn't great but probably fine enough). 

@v1bh475u we cleared up the confusion and i don't think is any action needed by you right now, sorry!

Issue #964, opened 2024-11-20
improve TCP/UDP socket creation matching


mike-hunhoff:
Add additional check for `parameter` argument of `0`.

https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket
> If a value of 0 is specified, the caller does not wish to specify a protocol and the service provider will choose the protocol to use.

https://man7.org/linux/man-pages/man2/socket.2.html
> The protocol specifies a particular protocol to be used with the
       socket.  Normally only a single protocol exists to support a
       particular socket type within a given protocol family, in which
       case protocol can be specified as 0.

https://github.com/mandiant/capa-rules/blob/e50d824553ae8d7168ac2d1f7b5540a93f944e28/communication/socket/udp/send/create-udp-socket.yml
https://github.com/mandiant/capa-rules/blob/e50d824553ae8d7168ac2d1f7b5540a93f944e28/communication/socket/tcp/create-tcp-socket.yml

PR #993, opened 2025-02-04
improve TCP/UDP socket creation matching #964

dhruvak001:
Fixes https://github.com/mandiant/capa-rules/issues/964

I have also added number: 17 = IPPROTO_UDP , can make the rule more precise by explicitly matching cases where the UDP protocol is specified. (but not needed specifically)

[x] No CHANGELOG update needed
[x] No new tests needed
[x] No documentation update needed

### DIFF
--- a/communication/socket/tcp/create-tcp-socket.yml
+++ b/communication/socket/tcp/create-tcp-socket.yml
@@ -17,7 +17,9 @@ rule:
   features:
     - or:
       - and:
-        - number: 6 = IPPROTO_TCP
+        - or:
+          - number: 0 = protocol (default)
+          - number: 6 = IPPROTO_TCP
         - number: 1 = SOCK_STREAM
         - number: 2 = AF_INET
         - or:
--- a/communication/socket/tcp/create-tcp-socket.yml
+++ b/communication/socket/tcp/create-tcp-socket.yml
@@ -12,6 +12,9 @@ rule:
       dynamic: call
     mbc:
       - Communication::Socket Communication::Create TCP Socket [C0001.011]
+    references:
+      - https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket
+      - https://man7.org/linux/man-pages/man2/socket.2.html
     examples:
       - Practical Malware Analysis Lab 01-01.dll_:0x10001010
   features:
--- a/nursery/create-udp-socket.yml
+++ b/nursery/create-udp-socket.yml
@@ -12,7 +12,7 @@ rule:
     mbc:
       - Communication::Socket Communication::Create UDP Socket [C0001.010]
     examples:
-      - 203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082:0x401240
+      # Have to find a suitable sample for this rule
   features:
     - or:
       - and:
--- a/nursery/create-udp-socket.yml
+++ b/nursery/create-udp-socket.yml
@@ -19,7 +19,7 @@ rule:
       - and:
         - number: 2 = AF_INET
         - number: 2 = SOCK_DGRAM
-        - or: 
+        - or:
           - number: 0 = protocol (default)
           - number: 17 = IPPROTO_UDP
         - or:
--- a/nursery/create-udp-socket.yml
+++ b/nursery/create-udp-socket.yml
@@ -11,14 +11,15 @@ rule:
       dynamic: call
     mbc:
       - Communication::Socket Communication::Create UDP Socket [C0001.010]
-  # examples: # Have to find a suitable sample for this rule
-      #- 203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082:0x401240
+    references:
+      - https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket
+      - https://man7.org/linux/man-pages/man2/socket.2.html
   features:
     - or:
       - and:
         - number: 2 = AF_INET
         - number: 2 = SOCK_DGRAM
-        - or: # Protocol (0 or IPPROTO_UDP=17)
+        - or: 
           - number: 0 = protocol (default)
           - number: 17 = IPPROTO_UDP
         - or:
--- a/nursery/create-udp-socket.yml
+++ b/nursery/create-udp-socket.yml
@@ -11,8 +11,8 @@ rule:
       dynamic: call
     mbc:
       - Communication::Socket Communication::Create UDP Socket [C0001.010]
-    examples:
-      # Have to find a suitable sample for this rule
+  # examples: # Have to find a suitable sample for this rule
+      #- 203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082:0x401240
   features:
     - or:
       - and:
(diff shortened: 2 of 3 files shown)


dhruvak001:
@williballenthin can you please review it, the example given initially in the code is said to be not existing.

mike-hunhoff:
@dhruvak001 your updates to the rule _create udp socket_ are causing the rule to no longer match for the test sample `203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082:0x401240`. You'll need to review the referenced function `0x401240` in a disassembly tool (Ghidra/IDA/etc.) to determine why the changes are not matching as expected. The referenced sample `203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082` can be found at https://github.com/mandiant/capa-testfiles.

dhruvak001:
@mike-hunhoff i have solved it by loosening overly strict conditions. But I think there maybe some other potential issue due to this?

williballenthin:
@dhruvak001 what are the issues? are you having trouble solving them?

dhruvak001:
@williballenthin As I decompiled it in Ghidra, 
The `create UDP socket` rule failed because it strictly required two AF_INET/SOCK_DGRAM (2) constants and the socket API call to appear in the same basic block, which the sample didn’t satisfy due to code structure/ordinal usage.
So, I:
1. Relaxed the rule to check for two separate 2 constants (for AF_INET and SOCK_DGRAM).
2. Changed the scope from basic block to entire function, allowing features to match across blocks.

So, the rule now accommodates split parameters and ordinal-based API calls.

mike-hunhoff:
@dhruvak001 the following change causes your rule to no longer match for the referenced sample:
```yaml
        - or:
          - number: 0 = protocol (default)
          - number: 6 = IPPROTO_TCP
```

Here is the assembly code that matched before:
```
        004012a4 66 89 45 ea     MOV        word ptr [EBP + local_1a],AX
        004012a8 33 c0           XOR        EAX,EAX
        004012aa 50              PUSH       EAX
        004012ab 6a 02           PUSH       0x2
        004012ad 6a 02           PUSH       0x2
        004012af 89 45 f0        MOV        dword ptr [EBP + local_14],EAX
        004012b2 89 45 f4        MOV        dword ptr [EBP + local_10],EAX
        004012b5 ff 15 08        CALL       dword ptr [->WS2_32.DLL::Ordinal_23]             = 80000017
                 61 40 00
```

You'll notice that there is no `0` constant for capa to extract as a `number` feature. In this scenario, the `0` constant comes from the `XOR EAX, EAX -> PUSH EAX` instructions, which capa does not presently handle. Increasing the scope to `function` only works because there are `0` constants elsewhere in the function for capa to match, but they are not related to this `socket` API call.

Your next step here should be to change the scope back to `basic block` and find a sample where the newly updated rule matches. If you can't find a new sample, we can move the rule to `nursery` until we do.

mike-hunhoff:
@dhruvak001 I've opened a [separate issue ](https://github.com/mandiant/capa/issues/2622) to discuss adding support for extracting `number(0)` features for cases like `xor eax, eax`. This is not something we should try and solve here. Please see my previous response for a recommended next step.

dhruvak001:
@mike-hunhoff I have changed it back to Basic Block, But could not find a way around the failed checks, seems like we have to solve the number(0) issue first. Or is there something else we could do about this?

mike-hunhoff:
You have two options:
- find a new example file that matches rule, list it under _examples_, and remove the entry that no longer matches
- remove _examples_ for now and move the rule to the _nursery_ directory

dhruvak001:
@mike-hunhoff for now i have removed the example and moved the rule `create-udp-socket.yml` to nursery. I will try to find a suitable example for this till then. Also can we build our own example in capa-testfiles and if so how can we do so(some steps to keep in mind)? or we have to search example from the existing testfiles?

nursery/create-udp-socket.yml
```diff
@@ -11,12 +11,16 @@ rule:
       dynamic: call
     mbc:
       - Communication::Socket Communication::Create UDP Socket [C0001.010]
-    examples:
-      - 203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082:0x401240
+  # examples: # Have to find a suitable sample for this rule
+      #- 203BD48BCC18434314AD60F4C8BC21E3D3422EB0624B22B827410F9BC63B4082:0x401240
```
mike-hunhoff, line 15: Please remove these comments

nursery/create-udp-socket.yml
```diff
@@ -11,12 +11,16 @@ rule:
...
   features:
     - or:
       - and:
-        - count(number(2 = AF_INET/SOCK_DGRAM)): 2 or more
+        - number: 2 = AF_INET
+        - number: 2 = SOCK_DGRAM
+        - or: # Protocol (0 or IPPROTO_UDP=17)
```
mike-hunhoff, line 21: Please remove this comment as the descriptions below are good enough.

communication/socket/tcp/create-tcp-socket.yml
```diff
@@ -12,12 +12,17 @@ rule:
       dynamic: call
     mbc:
       - Communication::Socket Communication::Create TCP Socket [C0001.011]
+    references:
+      - https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket
+      - https://man7.org/linux/man-pages/man2/socket.2.html
```
mike-hunhoff, line 17: Should these be added to _create UDP socket_?

mike-hunhoff (changes requested):
Thanks @dhruvak001 , almost there, I left minor comments for you to address.

mike-hunhoff:
> @mike-hunhoff for now i have removed the example and moved the rule `create-udp-socket.yml` to nursery. I will try to find a suitable example for this till then. Also can we build our own example in capa-testfiles and if so how can we do so(some steps to keep in mind)? or we have to search example from the existing testfiles?

We'd prefer to use an existing file from _capa-testfiles_, if possible, to save space. If there isn't an existing file then you're welcome to write your own. I don't think this is necessary at this time and is totally up to you.

dhruvak001:
@mike-hunhoff I have resolved the comments.

> 
> We'd prefer to use an existing file from _capa-testfiles_, if possible, to save space. If there isn't an existing file then you're welcome to write your own. I don't think this is necessary at this time and is totally up to you.

How should I find the correct testfile?, as there are many testfiles and it is not possible to decompile every file and read every line in the decompiler to find a suitable one.

mike-hunhoff:
> @mike-hunhoff I have resolved the comments.
> 
> > We'd prefer to use an existing file from _capa-testfiles_, if possible, to save space. If there isn't an existing file then you're welcome to write your own. I don't think this is necessary at this time and is totally up to you.
> 
> How should I find the correct testfile?, as there are many testfiles and it is not possible to decompile every file and read every line in the decompiler to find a suitable one.

The simplest way would be to run capa against `capa-testfiles` and see if any of the existing files match the rule you've written here. However, this is more or less like trying to find a needle in a haystack which is why the _nursery_ folder exists. Rules that are placed in this folder do not require an example file, which accounts for cases where an example file is not available, like here, or the example file is not public, etc..

mike-hunhoff (approved):
LGTM 🚀

PR #997, opened 2025-02-15
Add rules for detecting donut-related features

### LINKED ISSUE #994: detect donut loader
## Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your rule idea isn't already filed: [search](https://github.com/mandiant/capa-rules/issues?q=is%3Aissue+is%3Aopen+)

## Summary

Add detection rule(s) for [donut loader](https://github.com/TheWover/donut.git). At the moment, capa only detects the presence of RC4 in donut-related shellcode.

Image

## Examples

744ba02a89e2fe82b048e79520b6e6ac0e2ad21ed2b39fa06d0a0a22df91a45b

## Features

- maru calculation used for API hash calculation (https://github.com/TheWover/donut/blob/47758d787209dd1744f58c140102ac91b649df16/hash.c#L65)
- possibly parameters passed to various calls (e.g., HttpSendRequest, VirtualAlloc, RtlDecompressBuffer, etc.)
- PEB walking (at the moment existing PEB walk rules do not match the ones found in this loader)
- .NET loading via COM (IIDs/CLSIDs don't appear to be stored as-is in the compiled shellcode, may need to match by COM calls parameters?)
- ...probably more I haven't had the time to look into that might be good for rules

Still34:
## Summary

- This PR implements a few new rules for detecting a few features found in donut loader, a popular PE/assembly/script loader commonly used by malware.
- Adds
    - data-manipulation/encryption/chaskey/encrypt-data-using-chaskey
        - used for module encryption/decryption
    - data-manipulation/encryption/speck/encrypt-data-using-speck
        - used for hash calculations
    - load-code/dotnet/load-assembly-via-iassembly
        - used to load assembly in memory (this is currently a very loose rule; it shouldn't match false positive, but I haven't tested it on wide variety of samples yet - ideally this rule should be modified to look for known offsets related to CLR COM, but I could not get offset operand to work)
    - malware-family/donut-loader/load-shellcode-via-donut
        - uses all of the features above and matches partial functions that aren't unique enough to be created as their own rule
- Changes
    - data-manipulation/compression/decompress-data-using-aplib
        - adds number 31999 as some compilers may turn less than equal to less than, causing the condition to fail
- Closes #994

## Notes

- Rules regarding PEB access currently still do not match this or other donut shellcode samples, even though it should. I could not figure out how to edit the existing rules accordingly.
- Additional rules may be able to be written that looks for known offsets to various COM calls, but I also could not get offset-based rules to work.

## Sample

https://bazaar.abuse.ch/sample/d890c1c67d83f1131c065b5eb5f263cbf54559dbcdb4562c3bde3dc30d1a3205/


malware-family/donut-loader/load-shellcode-via-donut.yml
```diff
@@ -0,0 +1,62 @@
...
+      - Execution::Inter-Process Communication::Component Object Model [T1559.001]
+    examples:
+      - d890c1c67d83f1131c065b5eb5f263cbf54559dbcdb4562c3bde3dc30d1a3205
+  features:
+    - 3 or more:
+      - match: encrypt data using chaskey
+      - match: encrypt data using speck
+      - match: load assembly via IAssembly
+      - match: decompress data using aPLib
+      - function:
```
mike-hunhoff, line 25: Can this (or a subset of features) be tightened with _basic block_ to avoid potential FPs?

malware-family/donut-loader/load-shellcode-via-donut.yml
```diff
@@ -0,0 +1,62 @@
...
+                - number: 0x800000 = INTERNET_FLAG_SECURE
+                - number: 0x3000 = INTERNET_FLAG_IGNORE_CERT_CN_INVALID | INTERNET_FLAG_IGNORE_CERT_DATE_INVALID
+                - number: 0x803000 = INTERNET_FLAG_SECURE | INTERNET_FLAG_IGNORE_CERT_CN_INVALID | INTERNET_FLAG_IGNORE_CERT_DATE_INVALID
+          - instruction:
+            - mnemonic: cmp
+            - number: 200 = HTTP_STATUS_OK
+          - instruction:
+            - mnemonic: cmp
+            - number: 3 = DONUT_ENTROPY_DEFAULT
+      - function:
```
mike-hunhoff, line 45: Can this (or a subset of features) be tightened with _basic block_ to avoid potential FPs?

data-manipulation/encryption/speck/encrypt-data-using-speck.yml
```diff
@@ -0,0 +1,37 @@
+rule:
+  meta:
+    name: encrypt data using speck
+    namespace: data-manipulation/encryption/speck
+    authors:
+      - still@teamt5.org
+    scopes:
+      static: function
```
mike-hunhoff, line 8: Can this (or a subset of features) be tightened with _basic block_ to avoid potential FPs?

data-manipulation/encryption/chaskey/encrypt-data-using-chaskey.yml
```diff
@@ -0,0 +1,42 @@
+rule:
+  meta:
+    name: encrypt data using chaskey
+    namespace: data-manipulation/encryption/chaskey
+    authors:
+      - still@teamt5.org
+    scopes:
+      static: function
```
mike-hunhoff, line 8: Can this (or a subset of features) be tightened with _basic block_ to avoid potential FPs?

Still34:
I've been busy with other stuff this week and possibly for the next few days as well. I'll come back and address the comments before the month ends. Thanks!

malware-family/donut-loader/load-shellcode-via-donut.yml
```diff
@@ -0,0 +1,62 @@
+rule:
+  meta:
+    name: load shellcode via donut
```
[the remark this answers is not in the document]
Still34: Donut outside of debug mode does not feature any strings that can be matched. All of the strings that it does use are decrypted at runtime using chaskey.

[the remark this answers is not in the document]
mike-hunhoff: That makes sense, we should remove the _donut_ classification then. I don't think this rule as-is can confidently determine that the input file is _donut_.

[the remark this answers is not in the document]
Still34: I don't believe there are any other loaders that would match 3 or more of these rules. Perhaps tightening or adding additional function scope signatures would help? Ideally this rule should match donut or donut-adjacent features as that is the intent of writing such rule in the first place.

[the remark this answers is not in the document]
Still34: Alternatively, what naming choice would there be? As this seems too specific to engineer into just a generic shellcode loading routine. I'm not opposed to renaming it to a more generic one, but the current rule seem too specific to be named anything else - at least nothing that comes to my my mind.

load-code/dotnet/load-assembly-via-iassembly.yml
```diff
@@ -0,0 +1,22 @@
...
+      - d890c1c67d83f1131c065b5eb5f263cbf54559dbcdb4562c3bde3dc30d1a3205:0x1641B
+  features:
+    - and:
+      - instruction:
+        - mnemonic: mov
+        - number: 0x2008 = VT_ARRAY | VT_BSTR (for passing args)
+      - instruction:
+        - mnemonic: mov
+        - number: 0x118 = BindingFlags_InvokeMethod | BindingFlags_Static | BindingFlags_Public (InvokeMember_3)
+      - match: contain loop
```
williballenthin, line 22: maybe also `characteristic: indirect call`?

data-manipulation/encryption/speck/encrypt-data-using-speck.yml
Still34: I've experimented a bit, and I'm not entirely sure how I'd go about matching this at block level.

data-manipulation/encryption/chaskey/encrypt-data-using-chaskey.yml
Still34: Ditto with speck comment.

Still34 (commented):
I've tightened down the donut rule, and as for the basic block ones, I'm not entirely sure how I'd go about rewriting the rules to tighten the scope.

load-code/dotnet/load-assembly-via-iassembly.yml
Still34: Added.

malware-family/donut-loader/load-shellcode-via-donut.yml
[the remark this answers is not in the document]
mike-hunhoff: I think the _function_ scope matching helps here, thanks!

mike-hunhoff:
> The IAssembly rule - ideally we'd match additional conditions like mov from certain offset to match the structure, but last I tried something like operand offset it would not match. As it stands, that rule is extremely weak, and I think it may match false positives. Some suggestions or help on that rule would be appreciated as well. image

What about `bytes` features, e.g. https://github.com/mandiant/capa-rules/blob/5fb8cee820ab87a11b7cc9a4c08a07ce982152a5/communication/http/initialize-iwebbrowser2.yml?

Still34:
I'm more so hoping to match the structure offset. For example, this call to `SafeArrayCreateVector`
image

  I was hoping to use the `offset` feature but I could not get it to match with something like `offset: 0x70`

Still34:
I finally found out why the offset rule isn't working... The vivisect workspace loaded within IDA for some reason is not matching the `offset` rules whilst capa standalone does...?
image

mike-hunhoff:
@Still34 the IDA plugin and standalone binary use different implementations for feature extraction. It's possible that there is a bug in the IDA plugin implementation that is preventing it from extracting the expected _offset_ features. Can you please open a [new issue](https://github.com/mandiant/capa/issues) to describe what you're seeing here and provide the hash of the source sample so we can investigate further?

load-code/dotnet/load-assembly-via-iassembly.yml
```diff
@@ -0,0 +1,40 @@
...
+          - or:
+            - operand[1].offset: 0xE4
+            - operand[1].offset: 0x88
+        - instruction:
+          - mnemonic: mov
+          - number: 0x2008 = VT_ARRAY | VT_BSTR (for passing args)
+        - instruction:
+          - mnemonic: mov
+          - number: 0x118 = BindingFlags_InvokeMethod | BindingFlags_Static | BindingFlags_Public (InvokeMember_3)
+      - match: contain loop
```
mike-hunhoff, line 39: Do we need a _contain loop_ match? I worry that this may results in FNs

load-code/dotnet/load-assembly-via-iassembly.yml
```diff
@@ -0,0 +1,40 @@
...
+      - characteristic: indirect call
```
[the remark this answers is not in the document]
Still34: Moving it inside any of the instruction scope doesn't seem to work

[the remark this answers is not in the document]
mike-hunhoff: 🤦🏻 I missed that the _instruction_ blocks match _call_ or _mov_ mnemonics, where _indrect call_ won't match for _mov_ instructions. Thank you for checking.

mike-hunhoff (approved):
LGTM 🚀

PR #999, opened 2025-02-20
use "span of calls" scope for registry operations that span multiple calls, e.g. open registry key and set value

mike-hunhoff:
see original concerned expressed [here](https://github.com/mandiant/capa-rules/pull/954#pullrequestreview-2482035858).


mr-tz (approved):

williballenthin (approved):

williballenthin:
@jorik-utwente fyi

jorik-utwente:
Very nice work!

PR #1003, opened 2025-02-21
add "change registry key timestamp"

### LINKED ISSUE #1000: registry timestomping
https://www.inversecos.com/2022/04/malicious-registry-timestamp.html

> By calling a set of native API keys an attacker can modify the Last Write time of a registry key in an extremely trivial manner. The most important API is the “NtSetInformationKey” API which takes in a few parameters – the most important being the KEY_SET_INFORMATION_CLASS parameter which is used to modify the ‘Last Write’ time by passing in “KEY_WRITE_TIME_INFORMATION”.

> In the PoC built by [Joakim Schicht](https://github.com/jschicht), the tool uses the following API keys in conjunction with the offending “NtSetInformationKey”. I have broken down the use cases below:
> - NtCreateKey >> Creates or opens an existing registry key
> - NtOpenKey >> Opens an existing registry key
> - NtSetInformationKey >> Used to modify the time
> - NtEnumerateKey >> Returns information about a registry key
> - NtQueryKey >> Returns information from a registry key 
> - NtFlushKey >> Write the new timestamp immediately to disk

williballenthin:
closes #1000
https://github.com/mandiant/capa-testfiles/pull/275

### DIFF
--- a/host-interaction/registry/change-registry-key-timestamp.yml
+++ b/host-interaction/registry/change-registry-key-timestamp.yml
@@ -1,0 +1,23 @@
+rule:
+  meta:
+    name: change registry key timestamp
+    namespace: host-interaction/registry
+    authors:
+      - wballenthin@google.com
+    scopes:
+      static: basic block
+      dynamic: call
+    att&ck:
+      - Defense Evasion::Modify Registry [T1112]
+    references:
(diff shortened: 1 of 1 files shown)


mike-hunhoff (approved):
🚀

PR #1004, opened 2025-02-21
tighten Windows mutex related rules

mike-hunhoff:


williballenthin:
this broke capa tests: https://github.com/mandiant/capa/actions/runs/13476008913/job/37655288984

PR #1005, opened 2025-02-22
Added related wevtapi functions

JakePeralta7:
Added detection of clearing event logs with functions from wevtapi.dll, these functions can also be used remotely after a session is established with RPC using the EvtOpenSession function.

Reference Code:
[getel-arch/ClearLogsRemotely](https://github.com/getel-arch/ClearLogsRemotely)


JakePeralta7:
I've signed the CLA, how do I make the workflow rerun?

PR #1006, opened 2025-02-22
Clearing Event Log with wevtapi functions

JakePeralta7:
Added detection of clearing event logs with functions from wevtapi.dll, these functions can also be used remotely after a session is established with RPC using the EvtOpenSession function.

Reference Code:
[getel-arch/ClearLogsRemotely](https://github.com/getel-arch/ClearLogsRemotely)


williballenthin (approved):
thank you!

Initially I was unsure if the name "... remotely" was appropriate, but as I look at some examples and the API documentation, I see that NULL would be provided instead of the session, so requiring `EvtOpenSession` is great here.

williballenthin:
for example here is `WSReset.exe`, which clears a local log:
image

it imports *only* EvtClearLog, not any of the other routines.

so here's a VT search for local clearing: https://www.virustotal.com/gui/search/imports%253AEvtClearLog%2520and%2520not%2520imports%253AEvtOpenSession?type=files

and likely remote clearing: https://www.virustotal.com/gui/search/imports%253AEvtClearLog%2520and%2520imports%253AEvtOpenSession?type=files

anti-analysis/anti-forensic/clear-logs/clear-windows-event-logs-remotely.yml
```diff
@@ -0,0 +1,19 @@
+rule:
+  meta:
+    name: clear Windows event logs remotely
+    namespace: anti-analysis/anti-forensic/clear-logs
+    authors:
+      - 99.elad.levi@gmail.com
+    scopes:
+      static: function
```
williballenthin, line 8: please add `dynamic: span of calls`

JakePeralta7: Got it

JakePeralta7:
Sorry about all of the mess in the commit, it's my first time writing a CAPA rule and I'm not used to your workflows

williballenthin:
no worries! I appreciate the patience. and I'll squash merge so the history ends up clean.

really neat that you took the time to submit a rule after developing the POC. i hope we can continue to work together :-)

PR #1017, opened 2025-03-08
Addresses #965: Create separate rules for socket connects (TCP/UDP)

### LINKED ISSUE #965: create separate rule to match socket connects
capa presently only matches connecting TCP sockets, but it should match all connects as this is usually of interest. At a minimum, we should add a new rule to detect connecting UDP sockets.

https://github.com/mandiant/capa-rules/blob/e50d824553ae8d7168ac2d1f7b5540a93f944e28/communication/socket/tcp/connect-tcp-socket.yml#L19-L24

hafizfarhad:
- Add a generic `connect socket` rule capturing common connection APIs.
- Refactor the TCP connect rule to require a match on `create TCP socket` and the  `connect socket` rule
- Add a new UDP connect rule that requires a match on `create UDP socket` and the  `connect socket` rule.

Closes #965.

### DIFF
--- a/communication/socket/connect-socket.yml
+++ b/communication/socket/connect-socket.yml
@@ -1,0 +1,71 @@
+rule:
+  meta:
+    name: connect socket
+    namespace: communication/socket
+    authors:
+      - moritz.raabe@mandiant.com
+      - joakim@intezer.com
+      - mrhafizfarhad@gmail.com
+    description: >
+      Detects socket connection attempts using common APIs or ConnectEx setup.
+    scopes:
+      static: basic block
+      dynamic: span of calls
+    examples:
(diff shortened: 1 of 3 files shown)


mike-hunhoff:
Hi @hafizfarhad , it appears lints are failing because there is no example file. Please open a PR at https://github.com/mandiant/capa-testfiles with your example file. If you don't have an example file then you can place the corresponding rule(s) in the `nursery` directory,

hafizfarhad:
Hi @mike-hunhoff, I have updated `connect udp socket` rule with an example. And also opened a PR for test file [here](https://github.com/mandiant/capa-testfiles/pull/280). Please let me know if any further changes are required. Thank you :)

mike-hunhoff (approved):

PR #1019, opened 2025-03-11
Fix 'clear Windows event logs remotely'

JakePeralta7:
EvtOpenLog has been removed from the logic since it is unnecessary for EvtClearLog. The only handle EvtClearLog can accept is a session handle (acquired by EvtOpenSession), which is optional and only needed for remote operations.
image


mike-hunhoff (approved):
Great, thank you @JakePeralta7 LGTM 🚀

Issue #1023, opened 2025-03-14
dynamic: "create thread bypassing process freeze" FPs


mike-hunhoff:
I'm seeing many FPs for _create thread bypassing process freeze_ for capa dynamic anlaysis of CAPE sandbox, e.g.:
```
  NtCreateThreadEx(
    ThreadHandle=0x3f8,
    ProcessHandle=0xffffffff,
    StartAddress=0x100100,
    Parameter=0x40,
    CreateFlags=0x1,
    ThreadId=0x1df8,
    ProcessId=0x1a70,
    Module="example.exe",
  ) -> 0x0
```

capa is detecting `0x40` constant passed via _Parameter_ but we only want to match if it is passed via _CreateFlags_.

dhruvak001:
Even though with that I have properly nested the code, it still reads them like:
```
  features:
    - and:
      - api: ntdll.NtCreateThreadEx
      properties:
        arguments:
          CreateFlags: 0x40
``` 

Instead of: 
```
  features:
    - and:
      - api: ntdll.NtCreateThreadEx
        properties:
          arguments:
            CreateFlags: 0x40
```

PR #1025, opened 2025-03-15
dynamic: "create thread bypassing process freeze" FPs #1023

dhruvak001:
Fixes https://github.com/mandiant/capa-rules/issues/1023

[x] No CHANGELOG update needed
[x] No new tests needed
[x] No documentation update needed

### DIFF
--- a/nursery/create-thread-bypassing-process-freeze.yml
+++ b/nursery/create-thread-bypassing-process-freeze.yml
@@ -19,5 +19,6 @@ rule:
     - and:
       - or:
         - api: ntdll.NtCreateThreadEx
-        - string: "NtCreateThreadEx"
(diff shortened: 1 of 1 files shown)


dhruvak001:
Even though with that I have properly nested the code, it still reads them like:
```
  features:
    - and:
      - api: ntdll.NtCreateThreadEx
      properties:
        arguments:
          CreateFlags: 0x40
``` 

Instead of: 
```
  features:
    - and:
      - api: ntdll.NtCreateThreadEx
        properties:
          arguments:
            CreateFlags: 0x40
```

mr-tz:
This isn't valid rule syntax and is not something we can express unfortunately.

PR #1026, opened 2025-03-16
Update manipulate-boot-configuration.yml

JakePeralta7:
Escaping dots in regex is important

image


williballenthin (approved):

williballenthin:
neat counter example! thank you

PR #1028, opened 2025-03-18
fix: escape "." in regex features that are not expected to be dot operators

mike-hunhoff:
see https://github.com/mandiant/capa-rules/pull/1026 for an example of how unescaped `.` can results in FPs.


williballenthin (approved):
wow, great!

in addition to being more specific, I think this will make the patterns slightly faster, too!

PR #1034, opened 2025-03-21
add new rules to detect disabling system features via registry on Windows

mike-hunhoff:

### DIFF
--- a/nursery/disable-device-guard-features-via-registry-on-windows.yml
+++ b/nursery/disable-device-guard-features-via-registry-on-windows.yml
@@ -1,0 +1,27 @@
+rule:
+  meta:
+    name: disable Device Guard features via registry on Windows
+    namespace: impact/features
+    authors:
+      - mehunhoff@google.com
+    scopes:
+      static: function
+      dynamic: span of calls
+    att&ck:
+      - Defense Evasion::Impair Defenses::Disable or Modify Tools [T1562.001]
+    mbc:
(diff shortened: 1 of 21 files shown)


nursery/disable-device-guard-features-via-registry-on-windows.yml
```diff
@@ -0,0 +1,27 @@
...
+    att&ck:
+      - Defense Evasion::Impair Defenses::Disable or Modify Tools [T1562.001]
+    mbc:
+      - Defense Evasion::Disable or Evade Security Tools [F0004]
+  features:
+    - and:
+      - match: set registry value
+      - or:
+        - and:
+          - string: /SYSTEM\\ControlSet001\\Control\\DeviceGuard/i
```
williballenthin, line 19: should these be strictly ControlSet001 or might they be 002 or CurrentControlSet?

williballenthin (approved):

nursery/disable-device-guard-features-via-registry-on-windows.yml
mike-hunhoff: Good catch - I'm guessing that the file is referencing the symlink _CurrentControlSet_ while CAPE is recording the specific control set that was modified. I've updated these rules and all other existing rules that referenced control registry keys to use a more robust regex that captures indirect reference via _CurrentControlSet_ and direct access via _ControlSetXXX_.

williballenthin: thanks!

Issue #998, opened 2025-02-15
Linux kernel rootkit techniques


mike-hunhoff:
https://inferi.club/post/the-art-of-linux-kernel-rootkits

dhruvak001:
Some of the rule ideas I thought of is:

> Detect rootkits that intercept readdir to hide processes by filtering /proc entries.
> 
> Identify rootkits that manipulate VFS functions (e.g., vfs_readdir, filldir) to hide files.
> 
> Detect rootkits that use Netfilter hooks (e.g., nf_register_net_hook) to intercept/modify packets.
> 
> Detect modification of the system call table (sys_call_table) to redirect syscalls.
> 
> Identify rootkits that hide their presence by unlinking from the kernel module list.
> 
> Detect rootkits that escalate privileges by modifying struct cred (e.g., UID/GID overwrite).
> 
> Detect rootkits that capture keystrokes by intercepting input handlers (e.g., input_register_handler).

mike-hunhoff:
@dhruvak001 @hafizfarhad we've had the most success with capa rules that leverage `api`, `number`, and `string` features. I'd recommend picking technique(s) that could be matched by one or more of these features.

hafizfarhad:
Got It! Thank you so much @mike-hunhoff

dhruvak001:
@mike-hunhoff According to me 5. Unlinking a module from the kernel’s module list to avoid detection and 6. Modifying struct cred to escalate privileges (e.g., root UID), would be good to start with for capa rules?

mike-hunhoff:
> [@mike-hunhoff](https://github.com/mike-hunhoff) According to me 5. Unlinking a module from the kernel’s module list to avoid detection and 6. Modifying struct cred to escalate privileges (e.g., root UID), would be good to start with for capa rules?

Those sound reasonable to me, although, based on your description here these rules will require `offset` features so you'll need a good example file to use as a guide for your rule development.

aryanyk:
Hi @mike-hunhoff 

, I’ve opened a PR that adds two new Linux kernel rootkit detection rules for capa:

* **Privilege escalation via `prepare_kernel_cred` → `commit_creds`**
* **Netfilter hook registration via `nf_register_net_hook` / `nf_register_hook`**

Both rules are currently placed in the **nursery** directory since example samples are not yet available.

PR: #1136 

Feedback or suggestions are welcome.

PR #1039, opened 2025-04-12
Add rule: linux-rootkit-netfilter-hooks

hafizfarhad:
Addresses #998 
Hi @mike-hunhoff 

I am working on linux rootkit detection rules.

I started with Netfilter hooks detection since it doesn’t require offsets. I’m also trying to learn more about rootkits, so I thought this would be a straightforward approach.

However, I have a few questions:

1. I initially tried using API-based feature detection instead of string/substring matching, but the rule didn’t trigger (tested against this [rootkit code](https://gist.github.com/loneicewolf/226e3e20e6041d12a63a5e833ebb0503) that uses Netfilter hooks). I switched to substring detection as a workaround. Can we proceed with this approach, or might it introduce false positives?
2. Also, Is it acceptable to temporarily add these rules to the `nursery` directory until we develop 7–8 robust rules? If not, should they go under `anti-analysis/linux/rootkits/`, or do you recommend another location?

### DIFF
--- a/nursery/linux-rootkit-netfilter-hooks.yml
+++ b/nursery/linux-rootkit-netfilter-hooks.yml
@@ -1,0 +1,15 @@
+rule:
+  meta:
+    name: linux-rootkit-netfilter-hooks
+    namespace:
+    authors:
+      - mrhafizfarhad@gmail.com
+    scopes:
+      static: file
+      dynamic: file
+    references:
+      - https://gist.github.com/loneicewolf/226e3e20e6041d12a63a5e833ebb0503
+  features:
+    - or:
+      - substring: nf_register_net_hook
+      - substring: nf_register_net_hooks


mike-hunhoff:
> Addresses #998 Hi @mike-hunhoff
> 
> I am working on linux rootkit detection rules.
> 
> I started with Netfilter hooks detection since it doesn’t require offsets. I’m also trying to learn more about rootkits, so I thought this would be a straightforward approach.
> 
> However, I have a few questions:
> 
> 1. I initially tried using API-based feature detection instead of string/substring matching, but the rule didn’t trigger (tested against this [rootkit code](https://gist.github.com/loneicewolf/226e3e20e6041d12a63a5e833ebb0503) that uses Netfilter hooks). I switched to substring detection as a workaround. Can we proceed with this approach, or might it introduce false positives?
> 2. Also, Is it acceptable to temporarily add these rules to the `nursery` directory until we develop 7–8 robust rules? If not, should they go under `anti-analysis/linux/rootkits/`, or do you recommend another location?

Hi @hafizfarhad ,

1. Do you know why the API detection is not working? e.g. do the `api` features show when running https://github.com/mandiant/capa/blob/master/scripts/show-features.py against the target sample? This could be a bug in capa if we expect the corresponding `api` features to be present...so it's worth investigating further.
2. We only add rules to to the `nursery` directory when we don't have an example file for the rule. In this case, do you have an example file? If yes, you should open a PR to https://github.com/mandiant/capa-testfiles with your example file and add your rules to the capa directory that matches the namespace for each rule. If no, you can add the rules to the `nursery` directory until we find a good example file.

PR #1040, opened 2025-04-17
Deleting file content via truncation

JakePeralta7:
I'm not sure if `host-interaction/file-system/delete/delete-file.yml` is the correct place for this detection.

The technique is very straight forward and uses `SetEndOfFile` to clear the content of a file.

Here is an example of clearing PowerShell history using `SetEndOfFile`, https://github.com/getel-arch/Clear-ConsoleHistory/blob/main/src/main.c

### DIFF
--- a/host-interaction/file-system/write/clear-file-content.yml
+++ b/host-interaction/file-system/write/clear-file-content.yml
@@ -9,6 +9,8 @@ rule:
       dynamic: span of calls
     mbc:
       - File System::Writes File [C0052]
+    examples:
+      - e3a6fbbc9b315141da37e5abbae05bf20aa9f48d5f569c6353360f59a0315245:0x140001450
   features:
     - and:
       - api: kernel32.SetEndOfFile


JakePeralta7:
Also, MDE doesn't log anything about this interaction with the file

mike-hunhoff:
@JakePeralta7 this is neat, thank you! Is the file deleted? If yes, then this is great as-is. If no, then we should create a new rule file named something like `host-interaction/file-system/write/modify-file.yml` or something.

JakePeralta7:
> @JakePeralta7 this is neat, thank you! Is the file deleted? If yes, then this is great as-is. If no, then we should create a new rule file named something like `host-interaction/file-system/write/modify-file.yml` or something.

SetEndOfFile works with SetFilePointer (the ponter is zero by default - meaning the start of the file), so by calling it without setting the file pointer you basically erase the file content (not the file itself).

This technique is so cool, because the operating system will zero out all of the bytes of the file to prevent slack space abuse (Microsoft added a callback to SetEndOfFile that does that in Vista). (Source: Black Hat 2009, Anti-Forensics The Rootkit Connection slide 15)

JakePeralta7:
Maybe I should add exclusion to the usage of SetFilePointer in the function

JakePeralta7:
> > Maybe I should add exclusion to the usage of SetFilePointer in the function
> 
> I'm not sure what you mean here?

Take a look at the proposed change now

mike-hunhoff (changes requested):
This looks good! The linter is failing because you haven't provided an example file. If you don't have an example file, no problem, simply move the rule under the `nursery` directory for now and we'll move it out when an example file is found.

Lint error:
```
 clear file content
  FAIL: missing examples: Add meta.examples so that the rule can be tested and 
verified
```

JakePeralta7:
> This looks good! The linter is failing because you haven't provided an example file. If you don't have an example file, no problem, simply move the rule under the `nursery` directory for now and we'll move it out when an example file is found.
> 
> Lint error:
> 
> ```
>  clear file content
>   FAIL: missing examples: Add meta.examples so that the rule can be tested and 
> verified
> ```

I've added to the capa-testfiles repo my own example if that's alright

mike-hunhoff (approved):

PR #1044, opened 2025-05-02
added EtwpEventWriteFull

SolitudePy:
EtwpEventWriteFull is the core component for ETW. 

image


SolitudePy:
This function is not exported actually so it may not be easy to find it.

PR #1047, opened 2025-05-18
Unload minifilter drivers

JakePeralta7:
Sysmon driver can be unloaded using `FilterUnload` from `fltlib.dll`.

Proof of concept:
https://github.com/getel-arch/Unload-Sysmon/blob/main/src/main.c

### DIFF
--- a/host-interaction/driver/unload-driver.yml
+++ b/host-interaction/driver/unload-driver.yml
@@ -17,4 +17,3 @@ rule:
     - or:
       - api: NtUnloadDriver
       - api: ZwUnloadDriver
-      - api: fltlib.FilterUnload
--- a/host-interaction/driver/unload-driver.yml
+++ b/host-interaction/driver/unload-driver.yml
@@ -4,7 +4,6 @@ rule:
     namespace: host-interaction/driver
     authors:
       - moritz.raabe@mandiant.com
-      - JakePeralta7
     scopes:
       static: basic block
       dynamic: call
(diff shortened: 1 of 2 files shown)


mike-hunhoff (changes requested):
Thanks @JakePeralta7 . I'm unsure if this API should be added to `unload-driver.yml` or if a new rule should be created under https://github.com/mandiant/capa-rules/tree/d64c2c91ea4be309fb42aea13bf185bf76013ea2/host-interaction/filter called e.g. `unload-minifilter-driver.yml`. Thoughts?

JakePeralta7:
> Thanks @JakePeralta7 . I'm unsure if this API should be added to `unload-driver.yml` or if a new rule should be created under https://github.com/mandiant/capa-rules/tree/d64c2c91ea4be309fb42aea13bf185bf76013ea2/host-interaction/filter called e.g. `unload-minifilter-driver.yml`. Thoughts?

Missed that, certainly more fitting

host-interaction/driver/unload-driver.yml
```diff
@@ -4,13 +4,15 @@ rule:
     namespace: host-interaction/driver
     authors:
       - moritz.raabe@mandiant.com
+      - JakePeralta7
```
mike-hunhoff, line 7: Please revert all of the changes to this file

mike-hunhoff (changes requested):

PR #1064, opened 2025-08-13
Update patch-antimalware-scan-interface-function.yml

jtothej:
Update patch-antimalware-scan-interface-function.yml - include additional functions that can be patched


mike-hunhoff (approved):
Great, LGTM 🚀

PR #1072, opened 2025-09-01
Move `impersonate user` from nursery

JakePeralta7:
Added example and moved from `nursery` to `host-interaction/user`.

https://github.com/mandiant/capa-testfiles/pull/293

P.S. There is no need for the *Ex or clause, right?

### DIFF
--- a/host-interaction/user/impersonate-user.yml
+++ b/host-interaction/user/impersonate-user.yml
@@ -17,10 +17,14 @@ rule:
       - api: advapi32.LogonUser
       - and:
         - api: userenv.LoadUserProfile
-        - api: advapi32.GetUserName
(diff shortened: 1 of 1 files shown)


mike-hunhoff:
Also @JakePeralta7 , lints are failing now:
```
INFO     lint: rule "impersonate user" does not match for sample     lint.py:409
         bf9a888e226d8396e84e042732daf32b8c1df8fd675eb128e8da3f0d776            
         8bea5                                                                  

 impersonate user
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a 
different example
```

mike-hunhoff (approved):
Nice, LGTM 🚀

PR #1079, opened 2025-10-13
feat: log keystrokes via direct input

zeze-zeze:
test file: https://github.com/mandiant/capa-testfiles/pull/297
reference: https://www.gironsec.com/blog/2017/07/keylogger-using-directx/


mr-tz:
We don't need samples to cover all features. Would you agree that the other APIs would achieve the same?

PR #1097, opened 2026-01-06
add encrypt-data-using-hc-256

williballenthin:
https://github.com/mandiant/capa-testfiles/pull/300

inspired by here: https://blu3eye.gitbook.io/malware-insight/moonstone-sleet-trojaned-putty

image


williballenthin:
we already have HC-128, which is a similar but different algorithm. in particular:

> HC-128 is similar in function, and reduces each of key length, nonce, number of words in the tables P and Q, and number of table updating steps by half.[[2]](https://en.wikipedia.org/wiki/HC-256#cite_note-hc-128-2)
https://en.wikipedia.org/wiki/HC-256

that rule does not match the new example.

PR #1098, opened 2026-01-19
feat(targeting): add cloud and container collection/interaction rules

Maijin:
This adds rules for:
- enumerating AWS resources (CloudFormation, CloudTrail, DirectConnect, EC2, IAM, S3, Support)
- stealing credentials for AWS, GCP, Cloudflare
- stealing credentials for Docker and Kubernetes
- 
Rules are categorized into host-interaction and collection namespaces.

### DIFF
--- a/nursery/access-aws-credentials.yml
+++ b/nursery/access-aws-credentials.yml
@@ -1,0 +1,25 @@
+rule:
+  meta:
+    name: access AWS credentials
+    namespace: collection/cloud/aws
+    authors:
+      - maximemorin@google.com
+    scopes:
+      static: function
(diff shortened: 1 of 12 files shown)


collection/cloud/aws/steal-aws-credentials.yml
```diff
@@ -0,0 +1,27 @@
+rule:
+  meta:
+    name: steal AWS credentials
```
williballenthin: i'd loosen this a bit to "access..." since there are legit programs that read these files, too

Maijin:
Will do once I'm back on laptop!

PR #1110, opened 2026-01-29
Flexible version of send HTTP request

saniyafatima07:
The previous version of issue - #1099 is rigid.
So, made the regex slightly flexible which covers the issue - test_fix262 #2843 (capa)


saniyafatima07:
@mike-hunhoff Made the regex a bit flexible. 
Ran the test case which was failing, it passes now.
Could you please review if this fix works?

Thank you for you time!

Issue #1100, opened 2026-01-22
persist via Windows service


mike-hunhoff:
sample:

https://www.virustotal.com/gui/file/41d097e47778252aec073666502f9ab7eda2e47c6800f529064128dff186f39d

verdict:

0x404427,persist via Windows service,FP,"Function modifies NetBT parameters (NetbiosOptions), does not create or modify a service binary path for persistence.","Verify registry value is ImagePath or StartType."

decompilation:
```c
undefined4 __cdecl FUN_00404427(LPCSTR param_1,char *param_2)

{
  char cVar1;
  LSTATUS LVar2;
  undefined4 uVar3;
  HKEY local_410;
  CHAR local_40c [1024];
  BYTE local_c [8];
  
  memset(local_40c,0,0x400);
  lstrcpyA(local_40c,"SYSTEM\\CurrentControlSet\\Services\\NetBT\\Parameters\\Interfaces\\Tcpip_");
  lstrcatA(local_40c,param_1);
  LVar2 = RegOpenKeyExA((HKEY)0x80000002,local_40c,0,0x20006,&local_410);
  if (LVar2 == 0) {
    local_c[4] = '\x04';
    local_c[5] = '\0';
    local_c[6] = '\0';
    local_c[7] = '\0';
    cVar1 = *param_2;
    if (cVar1 == '0') {
      local_c[0] = '\0';
    }
    else if (cVar1 == '1') {
      local_c[0] = '\x01';
    }
    else if (cVar1 == '2') {
      local_c[0] = '\x02';
    }
    else {
      local_c[0] = '\0';
    }
    local_c[1] = 0;
    local_c[2] = 0;
    local_c[3] = 0;
    RegSetValueExA(local_410,"NetbiosOptions",0,4,local_c,4);
    RegCloseKey(local_410);
    uVar3 = 1;
  }
  else {
    uVar3 = 0;
  }
  return uVar3;
}
```

williballenthin:
i am very curious about the data you're quoting above 😁

PR #1113, opened 2026-01-30
Fix/persistence service registry fp

AdityaSkhorne:
Fixes issue #1100 — previous rule matched any registry writes under
SYSTEM\CurrentControlSet\Services\* and produced false positives
(e.g. NetBT\Parameters\Interfaces\Tcpip_* -> NetbiosOptions).

This patch tightens registry-based detection so we only flag persistence
when service APIs are used (CreateService/ChangeServiceConfig) or when
registry writes target canonical persistence values:

- ImagePath
- Start
- Type

This preserves detection of true service persistence while preventing
fp hits on network configuration edits.  Added sample hash to examples
for reviewer context.

Notes: recommend adding CI checks to require examples and negative tests
for registry-based persistence rules.

### DIFF
--- a/persistence/service/persist-via-windows-service.yml
+++ b/persistence/service/persist-via-windows-service.yml
@@ -1,33 +1,51 @@
+rule:
+  meta:
+    name: persist via Windows service
+    namespace: persistence/service
+    authors:
+      - moritz.raabe@mandiant.com
+    scopes:
+      static: function
+      dynamic: span of calls
+    att&ck:
+      - Persistence::Create or Modify System Process::Windows Service [T1543.003]
+      - Execution::System Services::Service Execution [T1569.002]
+    examples:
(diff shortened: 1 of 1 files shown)


persistence/service/persist-via-windows-service.yml
```diff
@@ -13,30 +13,20 @@ rule:
...
-            - api: advapi32.OpenService
-            - api: advapi32.StartService
-      - and:
-        - match: host-interaction/process/create
-        - or:
-          - and:
-            - string: /^sc(|\.exe)$/i
-            - string: /create /i
-          - string: /^sc(|\.exe) create/i
-          - string: /New-Service /i
```
mike-hunhoff: Please keep this logic and extend with any missing features that are included in your updates.

PR #1127, opened 2026-02-26
improve Heaven's Gate detection for computed selector variants

### LINKED ISSUE #1096: [64-bit execution via heavens gate] current rule misses obfuscated variants
## Summary

While analyzing a sample, I observed a false negative when the code segment (CS) selector value is computed at runtime
using arithmetic operations. The current static rule relies on detecting the constant `0x33` directly (for example via a `push` or `mov`), which means obfuscated variants that compute the same value dynamically are not detected.

```yaml
  - instruction:
            - mnemonic: push
            - number: 0x33
          - instruction:
            - mnemonic: mov
            - number: 0x33
```
## Examples

The sample I am analysing does the following:
```asm
push 3
pop eax
shl eax, 4
add al, 3
push eax
push 0
retf
```
In this binary, the Heaven’s Gate transition itself is implemented as runtime-executed shellcode rather than a statically visible code sequence.

Both approaches eventually place `0x33` into the CS register and perform a Heaven’s Gate  transition. The second version is simply obfuscated, which causes the current capa rule to miss it. Enumerating all possible arithmetic ways to compute the same value does not seem practical.

The sample was one I developed for a CTF challenge, please have a look:
Sample: <https://drive.proton.me/urls/ZD2V8HKZCM#fSqFaLtxBubo>
Password: infected

Link to the CTF challenge writeup: https://github.com/Cryptonite-MIT/niteCTF-2025/tree/main/rev/lament_of_the_seraph/solution

## Possible improvements

I understand that determining the value of the CS register at runtime is out of scope for static capa analysis. However, the presence of Heaven’s Gate may still be inferred indirectly. One possible way to make the rule more robust could be to detect the use of 64-bit–only instructions inside a 32-bit binary, which strongly suggests a transition into 64-bit mode.

```yaml
    - arch: i386
        - or:
          - mnemonic: syscall
          - mnemonic: sysret
          - mnemonic: swapgs
          - mnemonic: retfq
          - mnemonic: lretq
          - mnemonic: cdqe
          - mnemonic: movsxd
```

## Additional context

This appears to be a limitation of static pattern-based detection when values are constructed dynamically at runtime. I wanted to raise this for discussion and to understand whether this is an accepted limitation or if similar heuristics have been considered before.

akshat4703:
This PR improves detection for `64-bit execution via heavens gate` to cover obfuscated variants where the `0x33` selector is computed at runtime instead of used as an immediate constant.

Fixes #1096.

### Rule logic updates

1. Kept the existing strong pattern (`call $+5` + `add [esp], 5` + `retf`) and corrected the selector setup clause to properly allow either:
   - `push 0x33`, or
   - `mov 0x33`

2. Added a conservative secondary branch for the obfuscated selector construction pattern:
   - `push 0x3`
   - `shl ... , 0x4`
   - `add ... , 0x3`
   - `push 0x0`
   - `retf`

3. Added a reference to issue #1096 in `meta.references`.

## Validation

- `capafmt` passed for the updated rule.
- `lint --thorough -t "64-bit execution via heavens gate"` passed.
- Rule still matches referenced example:
  - `tests/data/79abd17391adc6251ecdc58d13d76baf.dll_`
- Negative sanity check:
  - `tests/data/mimikatz.exe_` -> no match for this rule.

### DIFF
--- a/anti-analysis/anti-disasm/64-bit-execution-via-heavens-gate.yml
+++ b/anti-analysis/anti-disasm/64-bit-execution-via-heavens-gate.yml
@@ -13,21 +13,40 @@ rule:
     references:
       - https://download.bitdefender.com/resources/files/News/CaseStudies/study/318/Bitdefender-TRR-Whitepaper-Maze-creat4351-en-EN-GenericUse.pdf
       - https://www.malwaretech.com/2014/02/the-0x33-segment-selector-heavens-gate.html
+      - https://github.com/mandiant/capa-rules/issues/1096
     examples:
       - 79abd17391adc6251ecdc58d13d76baf:0x10002385
   features:
     - and:
+      - arch: i386
       - or:
-        - description: set up retf to push 0x33 to CS indicating 64-bit mode
-        - instruction:
-          - mnemonic: push
-          - number: 0x33
-        - instruction:
-          - mnemonic: mov
-          - number: 0x33
-      - characteristic: call $+5
-        description: call $+5 pushes the current EIP onto the stack, +5 to jump past call insn bytes
-      - instruction:
-        - mnemonic: add = 'add dword ptr[esp], 5' updates the return address to point after retf
-        - number: 0x5 = length of add + retf insn bytes
-      - mnemonic: retf = set EIP = [ESP] and CS = [ESP+4]
+        - and:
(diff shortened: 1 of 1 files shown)


mr-tz:
fixing too narrow and stale fix, thanks for the input all!

PR #1130, opened 2026-03-08
Split .NET features from windows file & process rules

### LINKED ISSUE #710: do not use OS-specifics in .NET rules e.g. "... on Windows"
.NET is cross-platform so we should not us OS-specific references unless we are absolutely sure e.g. #706 where OS-specific types may be used. Instead, we should use "... in .NET" e.g. "read file in .NET" and, in the case of reading a OS-specific file, opt to add a match for "read file in .NET" alongside OS-specific features.

Overall, I think this change results in less user confusion e.g. #706 we see matches for "read file on Windows" when, in fact, we are looking at a file intended to be run on Android via Mono.

Open to more discussion!

CosmoWorker:
Addresses remaining 3 files in #710 
* Extracts .NET features from read, write & process windows rules.
* Migrates test sample (692f7fd198e804...) to new rule to pass lints & maintain test coverage.
* Passes all lint checks with no failure.
fin-capa-image
<br></br>
Closes #710


mike-hunhoff (changes requested):
Thanks @CosmoWorker , I've left comments for your review.

mike-hunhoff (changes requested):

mike-hunhoff (approved):

PR #1142, opened 2026-03-15
Add rule for zlib fast inflate

### LINKED ISSUE #494: rule idea: zlib fast inflate
see source: https://github.com/madler/zlib/blob/master/contrib/masmx86/inffas32.asm

I've seen this code used in shellcode; we can hit on the hard-coded strings or the assembly itself.

strings: 
- `Fast decoding Code from Chris Anderson`
- `invalid literal/length code`
- `invalid distance code`
- `invalid distance too far back`

priyank766:
Closes #494.

Adds a function-scope rule, `decompress data using ZLIB fast inflate`, for Chris Anderson's x86 assembly implementation of zlib `inflate_fast`, based on the assembly implementation referenced in the issue.

Validation:
- ran `scripts/capafmt.py` on the new rule
- local example-based lint is currently blocked because the referenced `capa-testfiles` sample is not available as a readable file in my environment

Reference:
- https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/contrib/masmx86/inffas32.asm

### DIFF
--- a/data-manipulation/compression/decompress-data-using-zlib-fast-inflate.yml
+++ b/data-manipulation/compression/decompress-data-using-zlib-fast-inflate.yml
@@ -1,0 +1,28 @@
+rule:
+  meta:
+    name: decompress data using ZLIB fast inflate
+    namespace: data-manipulation/compression
+    authors:
+      - priyank766
+    description: detects Chris Anderson's x86 assembly implementation of zlib inflate_fast
+    scopes:
+      static: function
+      dynamic: unsupported  # requires bytes features
+    mbc:
+      - Data::Decompress Data [C0025]
+    references:
+      - https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/contrib/masmx86/inffas32.asm
+      - https://github.com/mandiant/capa-rules/issues/494
+    examples:
+      - c2ba065654f13612ae63bca7f972ea91c6fe97291caeaaa3a28a180fb1912b3a.dll_:0x3FE40
+  features:
+    - and:
+      - description: Chris Anderson x86 assembly implementation of zlib inflate_fast
+      - string: Fast decoding Code from Chris Anderson
+      - string: invalid literal/length code
+      - string: invalid distance code
+      - string: invalid distance too far back
+      - bytes: 0F A2 81 FB 47 65 6E 75 75 38 81 F9 6E 74 65 6C 75 30 81 FA 69 6E 65 49 75 28
+        = cpuid GenuineIntel checks before enabling MMX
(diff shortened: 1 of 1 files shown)


priyank766:
@mike-hunhoff  & @mr-tz Can you please review it

data-manipulation/compression/decompress-data-using-zlib-fast-inflate.yml
```diff
@@ -0,0 +1,25 @@
...
+      - priyank766
+    description: detects Chris Anderson's x86 assembly implementation of zlib inflate_fast
+    scopes:
+      static: function
+      dynamic: unsupported  # requires bytes features
+    mbc:
+      - Data::Decompress Data [C0025]
+    references:
+      - https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/contrib/masmx86/inffas32.asm
+      - https://github.com/mandiant/capa-rules/issues/494
```
mike-hunhoff:
```suggestion
```

mike-hunhoff (changes requested):

priyank766:
@mike-hunhoff  i removed the unnecesarry reference line 
i wanted to know about this failed check why it is failed ?? i think `capa-testfiles` i have to create it in the test-files repo can i do that ??

mike-hunhoff:
@priyank766 rule_linter error:

```
INFO     lint: rule "decompress data using ZLIB fast inflate" does   lint.py:409
         not match for sample                                                   
         c2ba065654f13612ae63bca7f972ea91c6fe97291caeaaa3a28a180fb19            
         12b3a.dll_                                                             

 decompress data using ZLIB fast inflate
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a 
different example

rules with FAIL:
  - decompress data using ZLIB fast inflate
```

Please ensure all unit tests pass locally before requesting another review.

priyank766:
@mike-hunhoff  So sorry about jumping the gun on the review. 

I was actually trying to run the linter locally, but Windows Defender kept silently deleting the test sample from `capa-testfiles` every time I downloaded it . Because of that, my local tests were just failing with a "missing example" error rather than showing me the actual logic failure, so I mistakenly thought the GitHub CI was just acting up.

After digging into the assembly source, I finally saw my mistake: the "Fast decoding" string is just an unreferenced watermark in the binary, and the mask bytes are in an external data table. Since the rule is scoped to the function body, capa couldn't actually match them!

I just pushed a new commit removing those two unreachable features, so the rule should perfectly match the function scope now. I'll make sure everything is completely green before asking for another review next time!

PR #1152, opened 2026-04-13
Add rule for LZMA decompression

edeca:
Added a rule to detect the `LzmaProps_Decode` function from LZMA. 

Compared to other LZMA routines this code contains useful constants and is relatively static - e.g. it does not depend on the compression level and does not use small power-of-two constants.

A sample executable is referenced. This is not the one which I used to develop the rule, but contains almost identical code which can be used as a test case.


mr-tz:
@corkamig do you have thoughts on this one?

data-manipulation/compression/decompress-data-using-lzma.yml
```diff
@@ -0,0 +1,41 @@
...
+          - instruction:
+            - description: unoptimised divide by 9 or modulo 9
+            - number: 9
+        - and:
+          - instruction:
+            - description: optimized divide by 9
+            - number: 0x38E38E39
+          - instruction:
+            - description: optimized divide by 5
+            - number: 0x66666667
```
mr-tz:
```suggestion
            - number: 0x66666667

```

PR #1156, opened 2026-05-13
Add rule for payload stored as IPv6 strings

corkamig:


nursery/decrypt-shellcode-using-ipv6.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: decrypt shellcode using IPV6
+    namespace: load-code/shellcode
+    authors:
+      - corkami@google.com
+    scopes:
+      static: file
+      dynamic: file
+    examples:
+      - 48bada9bc9759492da3944645fbe3bc6
+      - 664b85996b069b0698c6ae8c8c0d3ce4
```
mike-hunhoff: Please open a PR for at least one of these files at https://github.com/mandiant/capa-testfiles so we can  graduate this rule from the  `nursery`. If not available, please remove this meta block and leave the rule under `nursery`.

nursery/decrypt-shellcode-using-ipv6.yml
```diff
@@ -0,0 +1,18 @@
+rule:
+  meta:
+    name: decrypt shellcode using IPV6
```
mike-hunhoff: Can we be a little more descriptive with the name?

nursery/decrypt-shellcode-using-ipv6.yml
```diff
@@ -0,0 +1,18 @@
...
+    namespace: load-code/shellcode
+    authors:
+      - corkami@google.com
+    scopes:
+      static: file
+      dynamic: file
```
mike-hunhoff: I'm concerned that `file` scope is too broad, especially for `dynamic` where we could match the `RtlIpv6[...]` calls directly.

nursery/decrypt-shellcode-using-ipv6.yml
```diff
@@ -0,0 +1,18 @@
...
+    examples:
+      - 48bada9bc9759492da3944645fbe3bc6
+      - 664b85996b069b0698c6ae8c8c0d3ce4
+  features:
+    - and:
+      - string: "RtlIpv6StringToAddressA"
+      - or:
+        - string: "4D5A:9000:0300:0000:0400:0000:FFFF:0000"
+        - string: "5648:89E6:4883:E4F0:4883:EC20:E80F:0000"
```
mike-hunhoff, line 16: See my comment above about `file` scope being too broad.

mike-hunhoff (changes requested):
Thank you @corkamig ! I've left comments for your review, please let us know if you have any questions.

nursery/decrypt-shellcode-using-ipv6.yml
mr-tz: this hopefully works for basic block (static) and call (dynamic) scopes

corkamig: Reduced scopes as function (static) and call (dynamic) work.

nursery/decrypt-shellcode-using-ipv6.yml
corkamig: I removed the hashes for now.

nursery/decrypt-shellcode-using-ipv6.yml
corkamig: Done

nursery/decrypt-shellcode-using-ipv6.yml
corkamig: Done

nursery/decrypt-shellcode-using-ipv6.yml
```diff
@@ -0,0 +1,15 @@
...
+      - corkami@google.com
+    scopes:
+      static: function
+      dynamic: call
+  features:
+    - and:
+      - string: "RtlIpv6StringToAddressA"
+      - or:
+        - string: "4D5A:9000:0300:0000:0400:0000:FFFF:0000"
+        - string: "5648:89E6:4883:E4F0:4883:EC20:E80F:0000"
```
mr-tz, line 16: do we need descriptions of what these are?

for the name: does it have to be shellcode or could it be a PE?

I assume infinite strings could be added here, can we use a regex/substring to catch more variants?

mr-tz:
once the name is finalized, need to adjust to fix the linter warning:

```
WARN: filename doesn't match the rule name: Rename rule file to match the 
rule name, expected: "decrypt-shellcode-stored-with-ipv6-strings.yml", found: 
"decrypt-shellcode-using-ipv6.yml"
```

nursery/decrypt-payload-stored-as-IPV6-strings.yml
```diff
@@ -0,0 +1,15 @@
...
+      - corkami@google.com
+    scopes:
+      static: function
+      dynamic: call
+  features:
+    - and:
+      - string: "RtlIpv6StringToAddressA"
+      - or:
+        - string: "4D5A:9000:0300:0000:0400:0000:FFFF:0000" # PE header
+        - string: "5648:89E6:4883:E4F0:4883:EC20:E80F:0000" # Havoc shellcode
```
mr-tz: nit: use the `description:` syntax here

mr-tz: https://github.com/mandiant/capa-rules/blob/master/doc/format.md#descriptions

mr-tz:
see lint fail, then good to go

for future reference please consider a local lint run first

mr-tz:
Thanks!

PR #1158, opened 2026-05-18
Add rule for PE stored as UUID

corkamig:


nursery/decode-pe-stored-as-uuid.yml
```diff
@@ -0,0 +1,20 @@
...
+  features:
+    - and:
+      - api: "UuidFromStringA"
+      - substring: "90909090"
+        description: nops
+      - substring: "00000000"
+      - substring: "-0000-"
+      - substring: "ffffffff"
+      - substring: "-ffff-"
+      - string: /.*4d-?5a.*/
```
mr-tz: also use substring?

nursery/decode-pe-stored-as-uuid.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: decode PE stored as UUID
+    namespace: load-code/pe
+    authors:
+      - corkami@google.com
+    scopes:
+      static: function
+      dynamic: call
+  features:
+    - and:
+      - api: "UuidFromStringA"
```
mr-tz: what about UuidFromStringW? use `UuidFromString`?

mr-tz (commented):
I think a reference sample would be helpful for this rule.

nursery/decode-pe-stored-as-uuid.yml
corkamig: It's a regexp - sometimes there's a dash between the hex digits.

nursery/decode-pe-stored-as-uuid.yml
corkamig: Done

nursery/decode-pe-stored-as-uuid.yml
```diff
@@ -0,0 +1,26 @@
...
+      - corkami@google.com
+    scopes:
+      static: function
+      dynamic: call
+  features:
+    - and:
+      - or:
+        - api: "UuidFromStringA"
+        - api: "UuidFromStringW"
+        - api: "UuidFromString"
```
mr-tz:
```suggestion
        - api: UuidFromString
```

mr-tz: this covers all cases

nursery/decode-pe-stored-as-uuid.yml
```diff
@@ -0,0 +1,26 @@
...
+        - and:
+          - substring: "90909090"
+            description: nops
+          - substring: "00000000"
+          - substring: "-0000-"
+          - substring: "ffffffff"
+          - substring: "-ffff-"
+          - string: /.*4d-?5a.*/
+            description: PE signature
+        - string: "90909090-9090-9090-9090-909090909090"
```
mr-tz: is this enough by itself?

mr-tz: for the rule name this doesn't line up then for me - how does this indicate a PE?

corkamig: There's the PE signature, and the ranges of 00/FF's wouldn't happen on a compact payload like a shellcode.

mr-tz: isnt the second branch there just this?

```yml
      - api: "UuidFromString"
      - or:
...
        - string: "90909090-9090-9090-9090-909090909090"
```

corkamig: I removed it to focus on pure PE.

corkamig:
Renaming the branch accidentally closed the PR.

corkamig:
I prefer keeping the nops string in the first version of the rule:
they are frequent in PEs, and present in the sample.

`00` and `FF` are extremly frequent, and `4d5a` could happen randomly.
Let's test the waters and avoid unexpected FPs with a robust rule first, and adapt it and loosen it later accordingly.

PR #1159, opened 2026-05-20
Add rule for samples compiled by Crystal

corkamig:
Sample in PR https://github.com/mandiant/capa-testfiles/pull/307


mr-tz (approved):
LGTM, the testfile is slow to process - let's discuss there how to deal with that and wait here with merging

PR #1166, opened 2026-06-19
Fix namespace typo error

CaptWake:
This PR fixes a typo in the capa rules where the namespace was incorrectly specified as:
```
host-interation
```
It should be:
```
host-interaction
```

### DIFF
--- a/nursery/get-current-process-command-line.yml
+++ b/nursery/get-current-process-command-line.yml
@@ -1,7 +1,7 @@
 rule:
   meta:
     name: get current process command line
-    namespace: host-interation/process
+    namespace: host-interaction/process
     authors:
       - william.ballenthin@mandiant.com
     scopes:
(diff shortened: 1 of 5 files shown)


williballenthin (approved):
nice catch @CaptWake!

williballenthin:
we should add a linter pass that checks nursery rule namespaces against the existing directory structure.

CaptWake:
I’m not sure if this has already been discussed in other issues or PRs and I'm a n00b in that field, but maybe it could be interesting to add an ai assistant like gemini, as an additional reviewer to help identify inconsistencies between rules and assist in suggesting/reviewing MBC and MITRE ATT&CK tags when promoting new rules? In some repositories like [CAPEv2](https://github.com/kevoreilly/CAPEv2) seems to help

williballenthin:
nice idea @CaptWake. i do think this would work well, especially for rule style and categorization sort of things.

unfortunately Gemini is going away, but I wonder if Google will have an alternative that they support for their repos. any ideas @mr-tz?

mr-tz:
Not that I know of unfortunately as of now.

mr-tz (approved):

PR #1168, opened 2026-06-28
rules: detect ps2exe binaries compiled with current MainApp/MainModule naming

shreethaar:
This adds detection coverage for PS2EXE's current code-generation
template. PS2EXE renamed its internal PSHost implementation classes
at some point after the samples this rule was originally written
against — older builds embed `PS2EXEApp`/`PS2EXE_Host`, while the
current MScholtes/PS2EXE master (confirmed against v0.5.0.33,
2025-08-19) generates `ModuleNameSpace.MainApp` / `MainModule*` /
`ConsoleColorProxy` instead. Binaries compiled with a current PS2EXE
install were producing false negatives against this rule.

Verified with a SantaStealer sample in the wild (sha256:
9df11356c5ac61d2aa7b5425e6322fc016b0ed5790dacb201396500b3eee03f7)
that was previously going undetected by this rule.

Sample uploaded to capa-testfiles: mandiant/capa-testfiles#316

Ran:
- `python scripts/lint.py .` (clean)
- `python scripts/lint.py --thorough -t "compiled with ps2exe" -v --samples .. .`
  — note: surfaces an unrelated FAIL on the dependency rule
  `compiled to the .NET platform` since its own examples aren't in my
  local sample set; full corpus validation runs in CI against the
  complete capa-testfiles checkout.

### DIFF
--- a/compiler/ps2exe/compiled-with-ps2exe.yml
+++ b/compiler/ps2exe/compiled-with-ps2exe.yml
@@ -14,6 +14,7 @@ rule:
     examples:
       - 8775ed26068788279726e08ff9665aab
       - 805f79e57e4ce24b95a04541e4ea55dd2ce9ba314ab2c43491328204bc25d017
+      - 9df11356c5ac61d2aa7b5425e6322fc016b0ed5790dacb201396500b3eee03f7
   features:
     - and:
       - match: compiled to the .NET platform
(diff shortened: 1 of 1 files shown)


mike-hunhoff (changes requested):
Thank you @shreethaar . Please add the example from https://github.com/mandiant/capa-testfiles/pull/316 to the rule's examples field.

shreethaar:
Added — example now points at the merged testfiles sample from mandiant/capa-testfiles#316. Thanks!

shreethaar:
Also added myself to authors for the new detection branch — happy to drop if you'd prefer.

mike-hunhoff (approved):

PR #1172, opened 2026-07-20
Detect assembly loading via nLoadImage

### LINKED ISSUE #1054: dotnet: load assembly via RuntimeAssembly.nLoadImage
source: https://dr4k0nia.github.io/posts/NixImports-a-NET-loader-using-HInvoke/#the-runtime

The sample that I found references this method via:

```
Assembly.GetExecutingAssembly().GetType().GetMethod("nLoadImage", BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.InvokeMethod)
```

Still34:
## Summary 

This PR aims to close #1054, which refers to a technique that uses the internal static method `nLoadImage` to load an assembly as opposed to the usual `Assembly.LoadImage`. There is one concern that might warrant additional discussion before merging,

Ideally, this rule should be scoped to `function` and use *just* the `api` references. Unfortunately, based on existing ITW samples and likelihood of how this is being abused in the wild, it is more than likely that `nLoadImage` will not appear right next to things like `Invoke` or `GetMethod` etc,. 

This is being tackled by re-scoping to `file` and expanding from just the `api` calls to include `import`, which is handled by the `file` scope. The problem is it doesn't look pretty and prone to mishandle copy-paste shall this rule be edited in the future.

This should address both scenarios where it's either being called like how the original issue calls it, 

```cs
Assembly.GetExecutingAssembly().GetType().GetMethod("nLoadImage", BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.InvokeMethod)
```

or, like how this ITW sample does it, address both in separate calls,

```cs
		// Token: 0x0600000C RID: 12 RVA: 0x0001EABC File Offset: 0x0001CCBC
		public static Assembly Wyqijumylyshe(byte[] containerSet)
		{
			object[] array = new object[5];
			CommandDesigner.Wyqijumy();
			if (!CommandDesigner.Wyqijumylyshe())
			{
				array[0] = containerSet;
			}
			array[4] = false;
			return (Assembly)CommandDesigner.Wyqijumylyshefishaepyky(CommandDesigner.Wyqijumylyshefi().GetType().GetMethod(CommandDesigner.Wyqijumylyshefishaepyky(), CommandDesigner.Wyqijumylyshefi()), null, array);
		}

		// Token: 0x0600000D RID: 13 RVA: 0x0001E8AD File Offset: 0x0001CAAD
		private static string Wyqijumylyshefishaepyky()
		{
			return "nLoadImage";
		}
```

I'm bringing this up because this potentially might impact or warrant discussions surrounding other .NET-specific rules as well. I can PR the sample into testfiles before or after the discussion.


load-code/dotnet/load-dotnet-assembly-via-runtimeassemblynloadimage.yml
```diff
@@ -0,0 +1,42 @@
...
+          - import: System.Reflection.Module::GetMethod
+          - import: System.Reflection.Module::GetMethodImpl
+          - import: System.Reflection.Module::GetMethods
+          - import: System.Reflection.Module::ResolveMethod
+          - import: System.Type::GetMethod
+          - import: System.Type::GetMethodImpl
+          - import: System.Type::GetMethods
+          - import: System.Reflection.MethodBase::Invoke
+          - import: System.Reflection.MethodInfo::Invoke
+          - import: System.Delegate::CreateDelegate
```
mike-hunhoff: Can this be condensed into a match for https://github.com/mandiant/capa-rules/blob/aed45e2571ebf7d2330e3daddbb5c472cc54966e/nursery/invoke-dotnet-assembly-method.yml? The referenced rule can be extended to include any missing techniques that you've listed here.

mike-hunhoff: Assuming yes, I noticed that the dynamic scope https://github.com/mandiant/capa-rules/blob/aed45e2571ebf7d2330e3daddbb5c472cc54966e/nursery/invoke-dotnet-assembly-method.yml should be updated to `span of calls` scope. Would you be up for including this change in your pull request? With this change, you'll also need to update the `dynamic` scope to `span of calls` for your rule.

load-code/dotnet/load-dotnet-assembly-via-runtimeassemblynloadimage.yml
```diff
@@ -0,0 +1,42 @@
+rule:
+  meta:
+    name: load .NET assembly via RuntimeAssembly.nLoadImage
+    namespace: load-code/dotnet
+    authors:
+      - still@teamt5.org
+    scopes:
+      static: file
```
mike-hunhoff, line 8: I see where you're going with this, and it's the result of a pitfall in capa's analysis, as it doesn't match functionality that is split across functions, etc.. This is not unique to capa's .NET analysis.

Generally, we want rules to match the smallest scope possible to reduce false positives. The risk of false positives for this rule using `file` scope depends on the uniqueness of the string `nLoadImage` within the context of .NET assemblies. If it's unique, we can stick with `file` scope, but if it's not unique, we should reduce static scope to `function` to avoid false positives. Thoughts?

mike-hunhoff (changes requested):
Thank you @Still34 , I've left comments for your review.

load-code/dotnet/load-dotnet-assembly-via-runtimeassemblynloadimage.yml
Still34: Generally, `nLoadImage` is not something that will exist on its own in a standard .NET compiled application. `nLoadImage` belongs to the underlying BCL that that is called when Assembly.Load is called, and therefore should belong to the RuntimeAssembly class under `System.Private.CoreLib.dll`.

Therefore, the `string` search *should* be safe. This is obviously not guaranteed when this is being scanned in memory as the entire BCL will be present, but that isn't what capa is designed for anyways.

Still34: I'd like to also explain the choice of the `file` scope - normally, I'd agree with the rationale that `function` should be preferred. However, this means the vast majority of the obfuscated samples that make use of the string and the invoke .NET assembly paths will tend to be in different functions, much like the example sample that was included with this PR - i.e., `file` will catch the usage whilst `function` won't. 

The question I have is do we maximize the scope (i.e., `file`) for better chance of catching this behavior, or do we not want to include obfuscation as part of the discussion but solely focus on the standard usage alone - since obviously we're not gonna be able to catch *all* obfuscated scenarios.

load-code/dotnet/load-dotnet-assembly-via-runtimeassemblynloadimage.yml
Still34: Just tested - the condensed match rule works. I've included the reference to other rule into this one with `match`. I've also adjusted the other rule per your request.

load-code/dotnet/load-dotnet-assembly-via-runtimeassemblynloadimage.yml
mike-hunhoff: > I'd like to also explain the choice of the `file` scope - normally, I'd agree with the rationale that `function` should be preferred. However, this means the vast majority of the obfuscated samples that make use of the string and the invoke .NET assembly paths will tend to be in different functions, much like the example sample that was included with this PR - i.e., `file` will catch the usage whilst `function` won't.
> 
> The question I have is do we maximize the scope (i.e., `file`) for better chance of catching this behavior, or do we not want to include obfuscation as part of the discussion but solely focus on the standard usage alone - since obviously we're not gonna be able to catch _all_ obfuscated scenarios.

I appreciate the additional context that you've provided here. We prefer the latter to reduce false positives, understanding there is a risk for false negatives in obfuscated samples.

mike-hunhoff: That being said, as we can assume the referenced string is unique to its use we can leverage `file` scope to reduce false negatives in this instance.

mike-hunhoff:
Thank you for the updates @Still34 . Rule lints are failing:
```
 load .NET assembly via RuntimeAssembly.nLoadImage
  FAIL: referenced example doesn't exist: Add the referenced example to samples 
```

Please address the issue and run the linter locally before requesting another review.

Still34:
PR over at https://github.com/mandiant/capa-testfiles/pull/319

mike-hunhoff (approved):

PR #1184, opened 2026-08-14
get-process-heap-flags: use GetProcessHeap

### LINKED ISSUE #1183: FN: process heap flags
in 5e17360d32e9b272bb7e1b97c8e4dca34622ec9ce08fd240fe2758cc3f67dc4a:0x14000193A

Image

we only look for access via the PEB, not via `GetProcessHeap`:

Image

williballenthin:
does not match our test sample because the `GetProcessHeap` call is split across basic blocks from the access to the field member. fixing that means increasing the scope from bb to function.

image

closes #1183

### DIFF
--- a/host-interaction/process/get-process-heap-flags.yml
+++ b/host-interaction/process/get-process-heap-flags.yml
@@ -15,7 +15,9 @@ rule:
       - al-khaser_x86.exe_:0x425470
   features:
     - and:
-      - match: PEB access
+      - or:
+        - match: PEB access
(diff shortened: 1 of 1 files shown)


williballenthin:
i'm not sure that GetProcessHeap returns the same thing as in `PEB->ProcessHeap`

ah, but ReactOS thinks so:

image

mike-hunhoff (approved):
LGTM 🚀

